PR #17, opened 2017-12-28
Use interconnect names without spaces

awygle:
Trivial fix - interconnect names must not contain spaces (name="i to o"), else they are interpreted as multiple nets (a compound name containing "i", a net named "to", and a net named "o").

### DIFF
--- a/ice40/tiles/plb/pb_type.xml
+++ b/ice40/tiles/plb/pb_type.xml
@@ -70,7 +70,7 @@
   <output name="o"  num_pins="8" equivalent="true"/>
 
   <interconnect>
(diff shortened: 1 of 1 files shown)


mithro:
Thank you for sending this! Could you add a "Signed-off-by" line to your commit?
```
       -s, --signoff
           Add Signed-off-by line by the committer at the end of the commit log message. The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit
           this work under the same license and agrees to a Developer Certificate of Origin (see http://developercertificate.org/ for more information).
```

See https://probot.github.io/apps/dco/ for more info too.

PR #33, opened 2018-02-12
Work on generating an iCE40 rr_graph.xml

daveshah1:
This is not much progress, but hopefully moving things in the right direction. Feel free not to merge if this actually makes things worse, I'm still getting my head around VPR...

This adds a few things that were missing from the icebox -> rr_graph.xml script and the architecture, so the initial parsing of the rr_graph now succeeds.

Now processing the node_indicies is still failing (segfault, or an assertion failure if you modify the vector accesses at 689 and 696 in rr_graph_reader.cpp to use "at"). This is because the PTCs aren't strictly consecutive, and VPR only allocates space for PTCs 0..n-1 (where n is number the of nodes), and the Python script now needs to be modified to not have gaps in the PTCs.


utils/icebox-rr_graph-import.py
```diff
@@ -172,14 +172,24 @@ def __init__(self, *args, **kw):
...
 switch_buffer = ET.SubElement(
     switches, 'switch',
-    {'id': str(buffer_id), 'name': 'buffer', 'buffered': "1"},
+    {'id': str(buffer_id), 'name': 'buffer', 'buffered': "1", 'type': "mux"},
 )
 
-# Routing switch connects two nets together to form a longer wire.
+switch_buffer_sizing = ET.SubElement(
+    switch_buffer, 'sizing',
+    {'mux_trans_size': "2.32", 'buf_size': "23.54"},
```
mithro: Could you add a comment were values these come from?

utils/icebox-rr_graph-import.py
```diff
@@ -522,8 +532,8 @@ def add_pin(pos, localname, dir, idx):
         }
         node = add_node(gname, attribs)
         ET.SubElement(node, 'loc', {
```
mithro: I think we could use a general x/y coordinate translation function?

mithro:
FYI - You might want to look at these following resources;
 * https://github.com/mithro/symbiflow-arch-defs/tree/master-ice40-local-route
 * https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/286
 * https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/245
 * https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/242

utils/icebox-rr_graph-import.py
daveshah1: Admittedly I don't know, they were taken from the example XML in the comment immediately above those lines...

PR #34, opened 2018-02-14
New testarch plus other fixes to common utils.

mithro:
Adding a new `testarch` to allow prototyping a bunch of vpr proposals without having to deal with the full complexity of a real device. 

Changing to using the `eblif` format [now that vpr supports it](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/281).

Also started coming up with a naming convention which includes type information.


mithro:
This pull request adds a new simple testarch which will hopefully make it easy to research if https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/286 is a good idea or not.

PR #37, opened 2018-02-17
generate model.xml and pb_type.xml from Verilog

daveshah1:
Current status:
 - Start of a Python framework for working with Yosys
 - model.xml generation from Verilog working
 - Basic pb_type.xml generation from Verilog working
 - multiple modes supported by @rqou's suggested method - setting `(*ALTERNATIVE_TO="fancy_module" *)` on the "mode" modules, making them all modes of "fancy_module".
 - Support for clocked timing specification using attributes (setup, hold and clock->Q), and combinational delays (both single values and matrices for bus ports) also using attributes at present.

Still TODO, short term:
 - Modify the artix7 specification and Makefiles to use the scripts to generate pb_type.xml and model.xml
 - Write suitable Verilog models for all iCE40 (and possibly other families)

Still TODO, long term:
 - Look at using native Verilog timing specifications. This is hard because Yosys doesn't support them at all presently.
 - Support for other interconnect types (muxes in the pb_type without using mux_gen for example)


utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+
+parser = argparse.ArgumentParser(description='Convert a Verilog simulation model into a VPR model.xml file')
+parser.add_argument('infiles', metavar='input.v', type=str, nargs='+',
+                    help='one or more Verilog input files')
+parser.add_argument('--top', help='top level module, not needed if only one module across all files')
+parser.add_argument('-o', help='output filename, default model.xml')
+
+args = parser.parse_args()
+
+aig_json = yosys_exec.vlog_to_json(args.infiles, True, True)
```
mithro: I'd go with `yosys.verilog_to_json` personally.

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/python3
+import yosys_exec
+import lxml.etree as ET
+import argparse, re
+import os, tempfile
+from yosys_json import YosysJson
```
mithro:
```python
import yosys.exec
from yosys.json import YosysJSON
``` 
maybe?

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+
+"""
+Convert a Verilog simulation model to a VPR `model.xml`
+"""
```
mithro: Doc string should probably be above the imports.

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+
+parser = argparse.ArgumentParser(description='Convert a Verilog simulation model into a VPR model.xml file')
```
mithro: I like this trick :-)
```python
parser = argparse.ArgumentParser(description=__doc__.strip())
```

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+parser.add_argument('infiles', metavar='input.v', type=str, nargs='+',
+                    help='one or more Verilog input files')
```
mithro: I find the best way to wrap `parser.add_argument` function calls is as follows;
```python
parser.add_argument(
    'infiles',
    metavar='input.v', type=str, nargs='+',
    help="""\
A long help description. 
One or more Verilog input files.
""")
```

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+args = parser.parse_args()
+
+aig_json = yosys_exec.vlog_to_json(args.infiles, True, True)
+
+# TODO: All of these functions involve a fairly large number of calls to Yosys
+# Although performance here is unlikely to be a major priority any time soon,
+# it might be worth investigating better options?
+
+def do_select(infiles, module, expr):
+    outfile = tempfile.mktemp()
```
mithro: [`tempfile.NamedTemporaryFile` maybe?](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/python3
```
mithro: `#!/usr/bin/env python3`

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+
+args = parser.parse_args()
+
+aig_json = yosys_exec.vlog_to_json(args.infiles, True, True)
+
+# TODO: All of these functions involve a fairly large number of calls to Yosys
+# Although performance here is unlikely to be a major priority any time soon,
+# it might be worth investigating better options?
+
+def do_select(infiles, module, expr):
```
mithro: This function could really do with a comment :-)

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+    sel_cmd = "prep -top %s -flatten; cd %s; select -write %s %s" % (module, module, outfile, expr)
+    yosys_exec.run_yosys_commands(sel_cmd, infiles)
+    pins = []
+    with open(outfile, 'r') as f:
+        for net in f:
+            snet = net.strip()
+            if(len(snet) > 0):
+                m = re.match(r"([^/]+)/([^/]+)", snet)
+                if m and m.group(1) == module:
+                    pins.append(m.group(2))
```
mithro: This pin extraction function would probably be clearer in it's own function.
```python
def extract_pins(pstr, _regex=re.compile("([^/]+)/([^/]+)"):
    """Get the pin name and size."""
    m = _regex.match(pstr)
    if m:
        return m.group(1), int(m.group(2))
    else:
        return None, -1
```

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,92 @@
...
+    return do_select(infiles, module, "select -list %s %%x* i:* o:* %%u %%i %s %%d" % (clk, clk))
+
+if "top" in args:
+    yj = YosysJson(aig_json, args.top)
+else:
+    yj = YosysJson(aig_json)    
+
+top = yj.get_top()
+models = ET.Element("models")
+model = ET.SubElement(models, "model", dict(name=top))
```
mithro: This would be more normal;
`model = ET.SubElement(models, "model", {'name': top})`

I might suggest the following too;
`model_xml = ET.SubElement(models, "model", {'name': top})`

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python3
+import yosys_exec
+import lxml.etree as ET
+import argparse, re
+import os, tempfile
+from yosys_json import YosysJson
+
+"""
+Convert a Verilog simulation model to a VPR `pb_type.xml`
```
mithro: None of these attributes are in your tests?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,46 @@
...
+
+The following Verilog attributes are allowed on instances:
```
mithro: I don't think we want a `PB_PATH` attribute. That should be specified on the command line?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,46 @@
...
+    - (* PB_PATH="../path/to/pb_type.xml" *) : import the pb_type for a given instance
+        from an external XML file instead of including it inline.
+    
+    - (* GENERATE_PB *) : generate (and overwrite if applicable) the above file, as well as including it
```
mithro: I don't get what this attribute is used for?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,46 @@
...
+        
+The following are allowed on both instances (except where PB_PATH is set but GENERATE_PB isn't), and
+the top level module.
+    
+    - (* BLIF_MODEL=".latch" *) : specify the corresponding blif_model for a given instance
```
mithro: I think we should use `CLASS` in both these cases?

My suggestions would look like this I think;
 * `(* TYPE="bel" CLASS="lut" *)` --> `blif_model=".names` and `class="lut"`
 * `(* TYPE="bel" CLASS="routing" *)` --> No `blif_model` and maybe `class="routing"` when I added support for that to `vpr`.
 * `(* TYPE="bel" CLASS="flipflop" *)` --> `blif_model=".latch` and `class="flipflop"`
 * `(* TYPE="blackbox" CLASS="mem" *)` --> `blif_model=".subckt <name of model>` and `class="memory"`
 * `(* TYPE="blackbox" *)` --> `blif_model=".subckt <name of model>"` and no `class=xxx` property?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,46 @@
...
+    
+    - (* CLASS="flipflop" *) : specify the corresponding class for a given instance
+
+    - (* PB_NAME="name" *) : override the name of the pb_type (default: cell type)
+
+The following are allowed on nets within modules:
+
+    - (* SETUP="clk 10e-12" *) : specify setup time for a given clock
```
mithro: I'm not sure, but I think Verilog has a native way of specifying setup and hold times? 
http://verilog.renerta.com/mobile/source/vrg00052.htm
> Timing Check Tasks are for verification of timing properties of designs and for reporting timing violations.

http://athena.ecs.csus.edu/~changw/class_docs/VerilogManual/me95rh.html
>This section, the final part of the delay modeling chapter, discusses some of the various system tasks that exist for the purposes of timing checks. Verilog contains many timing-check system tasks, but only the three most common tasks are discussed here: $setup, $hold and $width. Timing checks are used to verify that timing constraints are upheld, and are especially important in the simulation of high-speed sequential circuits such as microprocessors. All timing checks must be contained within specify blocks as shown in the example below.

Also think delays are part of verilog too? http://verilog.renerta.com/mobile/source/vrg00011.htm ?
> Delays specify a time in which assigned values propagate through nets or from inputs to outputs of gates.

If these attributes are "just temporary" until we add proper support for these values, then can you add a "FIXME: Use proper verilog for these" or something?

utils/vlog/yosys_json.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/python3
```
mithro: I wonder if [pyverilog](https://pypi.python.org/pypi/pyverilog) or [hdlparse](https://kevinpt.github.io/hdlparse/) would be useful supplement to yosys here?

I definitely want our verilog models to compatible with [hdlparse](https://kevinpt.github.io/hdlparse/) so we can use [symbolator](https://github.com/kevinpt/symbolator) to generate pretty diagrams like this;
image

utils/vlog/tests/dff.v
```diff
@@ -0,0 +1,9 @@
+module dff(input clk, rst, d, output q);
```
mithro: FYI My plan is to use [Doxygen Verilog](https://github.com/avelure/doxygen-verilog) for comments in the Verilog files.

utils/vlog/vlog_to_model.py
daveshah1: Yes, that makes sense. Will put Yosys stuff into a new folder

utils/vlog/vlog_to_model.py
daveshah1: Didn't know that :-) Will clean things up

utils/vlog/vlog_to_model.py
daveshah1: Could do, although we actually want a filename to pass to Yosys not an open file here so mktemp is fine I think

utils/vlog/vlog_to_pbtype.py
daveshah1: The tests at the moment are just for the model.xml generation, still working on examples for the pb_type stuff

utils/vlog/vlog_to_pbtype.py
daveshah1: It does, but Yosys doesn't output them in the JSON output at the moment. I will discuss this with Clifford, maybe I will look into adding this to Yosys

utils/vlog/yosys_json.py
daveshah1: Not sure, they might be useful for some stuff but I think both are too limited - particularly (like no attribute support), so it seemed easiest to use Yosys for everything. I thing hdlparse at least will ignore the stuff that it doesn't understand though, so we should still get diagrams.

utils/vlog/tests/sim.v
```diff
@@ -0,0 +1,12 @@
+`include "adder/sim.v"
```
mithro: Can you include the `pb_type.xml` and `model.xml` files for these verilog files?

utils/vlog/tests/pb_type.xml
```diff
@@ -0,0 +1,34 @@
+<pb_type blif_model=".subckt test_pb" name="test_pb" num_pb="1">
```
mithro: I'm pretty sure `blif_model` is not valid on objects which have inner `<pb_type>` objects.

utils/vlog/tests/pb_type.xml
```diff
@@ -0,0 +1,34 @@
...
+  <input equivalent="false" name="rst" num_pins="1"/>
+  <input equivalent="false" name="a" num_pins="1"/>
+  <input equivalent="false" name="b" num_pins="1"/>
+  <input equivalent="false" name="ci" num_pins="1"/>
+  <output equivalent="false" name="q" num_pins="1"/>
+  <output equivalent="false" name="co" num_pins="1"/>
+  <mode name="ADD">
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="adder/pb_type.xml"/>
+    <interconn>
+      <direct input="test_pb.a" name="test_pb_a_to_adder_a" output="adder.a"/>
```
mithro: As you can only have a single driver on any net, you can just name these the same as the output value.

utils/vlog/tests/sim.v
```diff
@@ -0,0 +1,14 @@
...
+`include "dff/sim.v"
+`include "lut2/sim.v"
+
+module test_pb(input clk, rst, a, b, ci, output q, co);
+
+wire d;
+
+(* MODE="ADD" *) adder adder_i(.a(a), .b(b), .ci(ci), .y(d), .co(co));
+(* MODE="LUT" *) lut2 lut2_i(.a(a), .b(b), .y(d), .co(co));
```
mithro: So, this doesn't really make sense -- is this suppose to be a `test_pb` object which has two operational modes -- operating like an `adder` **or** a `lut2`?

I think you want something like this;
```verilog
`include "../vpr/bel/lut2.v"
`include "../vpr/bel/lut4.v"

module multi_lut(input a1, a2, a3, output o1, o2);
  parameter MODE;
  generate
    if (MODE == LUT3) begin
      LUT4 lut4 (.O(o1), .I0(a1), .I1(a2), .I2(1'b0));
    end else if (WIDTH == LUT2x2) begin
      LUT2 lut2_1 (.O(o1), .I0(a1), .I1(a2));
      LUT2 lut2_2 (.O(o2), .I0(a2), .I1(a3));
    end else begin
      $error("Invalid MODE set for muli_lut");
    end
  endgenerate
endmodule
```

Which should be translated too;
```xml
<pb_type name="multi_lut">
   ....
   <mode name="LUT3">
      <xi:include href="../vpr/bel/lut4/pb_type.xml"/>
      <interconnect>
         <direct input="multi_lut.a1" name="lut4.I0" output="lut4.I0"/>
         <direct input="multi_lut.a2" name="lut4.I1" output="lut4.I1"/>
         <direct input="multi_lut.a3" name="lut4.I2" output="lut4.I2"/>
         <!-- FIXME: Add once VPR supports constants  <direct input="VCC" name="lut4.I0" output="lut4.I2"/> -->
         <direct input="lut4.O" name="multi_lut.o1" output="multi_lut.o1"/>
         <direct input="dff.q" name="dff_q_to_test_pb_q" output="test_pb.q"/>
      </interconnect>
   </mode>
   <mode name="LUT2x2">
      <xi:include href="../vpr/bel/lut2/pb_type.xml" XXXX />
      <xi:include href="../vpr/bel/lut2/pb_type.xml" XXXX />
      <interconnect>
         <direct input="multi_lut.a1" name="lut2_1.I0" output="lut2_1.I0"/>
         <direct input="multi_lut.a2" name="lut2_1.I1" output="lut2_1.I1"/>
         <direct input="multi_lut.a2" name="lut2_2.I0" output="lut2_2.I0"/>
         <direct input="multi_lut.a3" name="lut2_2.I1" output="lut2_2.I1"/>
         <direct input="lut2_1.O" name="multi_lut.o1" output="multi_lut.o1"/>
         <direct input="lut2_2.O" name="multi_lut.o2" output="multi_lut.o2"/>
      </interconnect>
   </mode>
</pb_type>
```

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,210 @@
...
+
+The following are allowed on a top level module:
+
+    - (* TYPE="bel|blackbox" *) : specify the type of the module
+    (either a Basic ELement or a blackbox named after the pb_type
+
+    - (* CLASS="lut|routing|flipflop|mem" *) : specify the class of an given
+    instance. Must be specified for BELs
+
+    - (* PB_NAME="name" *) : override the name of the pb_type
```
mithro: Your adding this for when `.subckt` != the module name?

I think we can probably drop support for that?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,210 @@
...
+        cname_to_pb_name[cname] = mod.get_cell_attr(cname, "PB_NAME", i_of)
+
+    for cname, i_of in mod.get_cells(False):
+        pb_name = mod.get_cell_attr(cname, "PB_NAME", i_of)
+        module_path = os.path.dirname(yj.get_module_file(i_of))
+        pb_type_path = "%s/pb_type.xml" % module_path
+        pb_mode = mod.get_cell_attr(cname, "MODE", "")
+        # If MODE is used, need to wrap cell pb_type and interconnect in "mode"
+        is_modal = (pb_mode != "")
+        if is_modal:
```
mithro: I find the following quite useful...

```python
xi_url = "http://www.w3.org/2001/XInclude"
ET.register_namespace('xi', xi_url)
xi_include = "{%s}include" % xi_url

pb_type_xml = ET.Element(
    'pb_type', {
        'name': tile_name,
        'num_pb': str(1),
    },
    nsmap = {'xi': xi_url})

def xi_include(parent, href):
  return ET.SubElement(parent, xi_include, {'href': href})

```

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,210 @@
...
+            mode_xml = ET.SubElement(pb_type_xml, "mode", {'name': pb_mode})
+            ET.SubElement(mode_xml, "{http://www.w3.org/2001/XInclude}include", {'href': pb_type_path})
+            mode_ic = ET.SubElement(mode_xml, "interconn")
+        else:
+            ET.SubElement(pb_type_xml, "{http://www.w3.org/2001/XInclude}include", {'href': pb_type_path})
```
mithro: Nit: Add newline here...

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,210 @@
...
+        # In order to avoid overspecifying interconnect, there are two directions we currently
+        # consider. All interconnect going INTO a cell, and interconnect going out of a cell
+        # into a top level output - or all outputs if "mode" is used.
+        inp_cons = mod.get_cell_conns(cname, "input")
+        for pin, net in inp_cons:
+            drvs = mod.get_net_drivers(net)
+            if len(drvs) == 0:
+                print("WARNING: pin %s.%s has no driver, interconnect will be missing" % (pb_name, pin))
+            elif len(drvs) > 1:
+                print("WARNING: pin %s.%s has multiple drivers, interconnect will be overspecified" % (pb_name, pin))
```
mithro: I feel like this should be a fatal error?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,210 @@
...
+        else:
+            ET.SubElement(pb_type_xml, "{http://www.w3.org/2001/XInclude}include", {'href': pb_type_path})
+        # In order to avoid overspecifying interconnect, there are two directions we currently
+        # consider. All interconnect going INTO a cell, and interconnect going out of a cell
+        # into a top level output - or all outputs if "mode" is used.
+        inp_cons = mod.get_cell_conns(cname, "input")
+        for pin, net in inp_cons:
+            drvs = mod.get_net_drivers(net)
+            if len(drvs) == 0:
+                print("WARNING: pin %s.%s has no driver, interconnect will be missing" % (pb_name, pin))
```
mithro: This is pretty normal with multiple modes. Maybe this warning only shows up when not generating a mode thingy?

utils/vlog/yosys/json.py
```diff
@@ -0,0 +1,161 @@
...
+
+import os, sys
+import json
+
+class YosysModule:
+    def __init__(self, name, module_data):
+        self.name = name
+        self.data = module_data
+
+    # Return the name of the module
```
mithro: We don't use the getter / setter paradigm in Python. If you want values to be "private" prefix them with `_`. You can also use the [`@property decorator`](https://www.python-course.eu/python3_properties.php) to make "generated properties", see example below;
```python

class YosysModule:

    @property
    def ports(self):
        plist = []
        for port, pdata in self.data["ports"].items():
            plist.append((port, len(pdata["bits"]), pdata["direction"]))
        return plist

y = YosysModule()
print(y.ports)

```

utils/vlog/yosys/json.py
```diff
@@ -0,0 +1,161 @@
...
+    def get_name(self):
+        return self.name
+
+    # Return a list of ports of a module (default: top), as a 3-tuple (name, width, dir)
```
mithro: This should be a docstring in a format supported by [sphinx](http://www.sphinx-doc.org/en/stable/ext/napoleon.html).

I like the NumPy formatting;
```python

def get_ports(self):
    """List of ports on a module.

    Returns
    -------
    name : str or None
    width : int
        The width in bits
    dir : str
        The direction, should be either `in` or `out`
    """
    pass
```

utils/vlog/yosys/json.py
```diff
@@ -0,0 +1,161 @@
...
+        io_drivers = [(self.name, _) for _ in self.get_conn_io(net, "input")]
+        cell_drivers = self.get_conn_ports(net, "output")
+        return io_drivers + cell_drivers
+
+    # Return a list of sinks of a net, as a 2-tuple as above
+    def get_net_sinks(self, net):
+        # top level *outputs* and cell inputs are both drivers
+        io_drivers = [(self.name, _) for _ in self.get_conn_io(net, "output")]
+        cell_drivers = self.get_conn_ports(net, "input")
+        return io_drivers + cell_drivers
```
mithro: Nit: Newline here...

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,270 @@
+#!/usr/bin/env python3
+
+"""\
+Convert a Verilog simulation model to a VPR `pb_type.xml`
+
+The following are allowed on a top level module:
+
+    - `(* TYPE="bel|blackbox" *)` : specify the type of the module
```
mithro: Actually, now I think about it -- we are confusing things here; there are two types of objects;
 * `BEL`s - Things which can't be decomposed.
 * `BLOCK`s - An object which contains multiple things (IE more blocks or other `BEL`s).

A `bel` is either;
 * a blackbox (`BEL_BB`)
 * one of the primitive types (`BEL_LT` / `BEL_RX` / `BEL_MX` / `BEL_MM` / `BEL_FF`).

See this doc -> https://github.com/mithro/symbiflow-arch-defs/tree/working-on-fasm#names

We should be able to use the `(* blackbox *)` property you used above and any module which is a `(* backbox *)` and doesn't have any `(* ALTERNATIVE_TO=xxxx *)` modes is thus a "BEL_BB" module.

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,270 @@
...
+    must be visible at the time of pb_type generation.
+
+The following are allowed on nets within modules (TODO: use proper Verilog timing):
+    - `(* SETUP="clk 10e-12" *)` : specify setup time for a given clock
+
+    - `(* HOLD="clk 10e-12" *)` : specify hold time for a given clock
+
+    - `(* CLK_TO_Q="clk 10e-12" *)` : specify clock-to-output time for a given clock
+
+    - `(* DELAY_CONST_{input}="30e-12" *)` : specify a constant max delay from an input (applied to the output)
```
mithro: Why not `(* DELAY_CONST="{input} 30e-12" *)` like the clock examples?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -0,0 +1,270 @@
...
+    - `(* DELAY_MATRIX_{input}="30e-12 35e-12; 20e-12 25e-12; ..." *)` : specify a VPR
+        delay matrix (semicolons indicate rows). In this format columns specify
+        inputs bits and rows specify output bits. This should be applied to the output.
+
+The following are allowed on ports:
+    - `(* CLOCK *)` : force a given port to be a clock
+
+    - `(* ASSOC_CLOCK="RDCLK" *)` : force a port's associated clock to a given value
+
+    - `(* PORT_CLASS="clock" *)` : specify the VPR "port_class"
```
mithro: FYI - `port_class` can only be specified on non-blackbox BELs.

utils/vlog/yosys/json.py
```diff
@@ -0,0 +1,236 @@
...
+            if cname == cell:
+                return cdata["type"]
+        return None
+
+    @property
+    def module_attrs(self):
+        """All attributes of a module as a dictionary"""
+        return self.data["attributes"]
+
+    def attr(self, attr, defval = None):
```
mithro: Maybe use `__getattr__` ? https://blog.rmotr.com/python-magic-methods-and-getattr-75cf896b3f88

utils/vlog/yosys/run.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
+import os, subprocess, sys, json
+import yosys.utils, tempfile, re
+
+def get_yosys():
+    return os.getenv('YOSYS', "yosys")
+
+def yosys_get_output(params):
```
mithro: These commands probably don't need `yosys_` in the name. You can just do a 
```python

import yosys
yosys.get_output(params)
```

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
+"""
+Convert a Verilog simulation model to a VPR `model.xml`
+
+The following Verilog attributes are considered on ports:
+    - `(* CLOCK *)` : force a given port to be a clock
```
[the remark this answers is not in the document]
daveshah1: At the moment I look at connections to the clock inputs of $dff cells, which is not ideal, for auto-detection. Just using a pin named `clk` doesn't work because of BRAM cells or IO cells (in the iCE40) with multiple clocks.

utils/vlog/vlog_to_model.py
```diff
@@ -0,0 +1,88 @@
...
+
+    - `(* ASSOC_CLOCK="RDCLK" *)` : force a port's associated clock to a given value
```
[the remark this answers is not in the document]
daveshah1: It can apply to inputs too, hence the name. Maybe two attributes, `INPUT_CLOCK` and `OUTPUT_CLOCK` that do the same thing would actually be cleaner?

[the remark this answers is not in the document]
mithro: Maybe `(* USING_CLOCK="RDCLK" *)`

Issue #12, opened 2017-12-26
Refactor the Makefiles


mithro:
They are a pretty huge mess of stuff

mithro:
Did some refactoring of the `Makefile.mux` files here -> https://github.com/mithro/symbiflow-arch-defs/blob/working-on-fasm/common/make/mux.mk

mithro:
This is in progress here -> #56

elmsfu:
Do you have more in find to refactor the makefiles?

mithro:
I guess this is done!

PR #58, opened 2018-03-01
Adding DSP48E1 high level

elmsfu:
No attempt at registers or pattern detector


mithro:
@daveshah1 - It would be awesome to get some review of this and how easy this will be to auto-convert using your Verilog to XML tool.

artix7/primitives/dsp48e1/ALU/sim.v
```diff
@@ -0,0 +1,25 @@
+module ALU
```
mithro: Directory names should be all lowercase.

artix7/primitives/dsp48e1/ALU/sim.v
```diff
@@ -0,0 +1,25 @@
...
+   input wire [3:0]  ALUMODE;
+   input wire 	     CARRYIN;
+   input wire 	     MULTSIGNIN;
+
+   output wire [47:0] OUT;
+   output wire [3:0]  CARRYOUT;
+   output wire 	      MULTSIGNOUT;
+
+endmodule // ALU
+
```
mithro: Extra whitespace at the end of the file.

artix7/primitives/dsp48e1/pb_type.xml
```diff
@@ -0,0 +1,65 @@
+<!-- A diagram for the DSP48E1 is shown in;
```
mithro: We should add this comment to the verilog too.

artix7/primitives/dsp48e1/sim.v
```diff
@@ -0,0 +1,109 @@
...
+   NMUX4 #(.NBITS(48)) ymux (.I0(48'h000000000000), .I1({5'b00000, MULT_OUT[42:0]}), .I2(48'hFFFFFFFFFFFF), .I3(C), .S(OPMODE[3:2]), .O(Y));
+
+   // TODO(elmsfu): z mux atually has 7 outputs from table 2-9, but 2 and 4 are both P
+   NMUX7 #(.NBITS(48)) zmux (.I0(48'h000000000000), .I1(PCIN), .I2(P), .I3(C), .I4(P), .I5({ {17{PCIN[47]}}, PCIN[47:17]}), .I6({ {17{P[47]}}, P[47:17]}), .S(OPMODE[6:4]), .O(Z));
+
+   ALU alu (.X(X), .Y(Y), .Z(Z), .ALUMODE(ALUMODE), .CARRYIN(CARRYIN), .MULTSIGNIN(MULTSIGNIN), .OUT(P), .CARRYOUT(CARRYOUT), .MULTSIGNOUT(MULTSIGNOUT));
+
+   assign PCOUT = P;
+
+endmodule // DSP48E1
```
mithro: Extra whitespace again :-)

daveshah1:
Looks good to me from a Verilog to XML point of view. From VPR's point of view the DSP48E1 is a black box, so the inner content doesn't matter.

All you need to do is to wrap the contents of the DSP48E1 sim.v in `ifndef PB_TYPE` so it is ignored for that purpose, and both model.xml and pb_type.xml generation should work well.

daveshah1:
Of course at some point timing information will also need to be added, but we haven't got a good format for that yet so that can wait. The path information in model.xml (combinational sinks and associated clocks) should be picked up automatically but can be overridden by attributes (see the Verilog generation Python scripts) if needed.

mithro:
Hi @elmsfu,

I just landed #12 pull request which will require you to do a rebase of these changes. 

The names of files have also changed in the following way;
 * `pb_type.xml` to `XXX.pb_type.xml`
 * `model.xml` to `XXX.model.xml`
 * `sim.v` to `XXX.sim.v`

mithro:
@daveshah1 I think we want to follow the same approach as the carry4, IE have both a "complete" and a "split" versions.

daveshah1:
IMO that's a nice to have but not a major concern, as I don't think the Xilinx tools let you fracture a DSP48E1 
- they will infer it for certain Verilog patterns, but that's a synthesis rather than P&R issue, and would need much more work on the Yosys side to be useful.

mithro:
@daveshah1 We also need the split models for when converting back from bitstream into Verilog

daveshah1:
So long as we have a complete simulation model available, I don't see a problem just including a DSP48E1 instance in the generated Verilog - indeed that's the behaviour I would expect.

mithro:
@elmsfu What is the status of this pull request?

elmsfu:
Failing at clean. Will look at soon.
``` 
Running make clean
-------------------------------------------------------------------
make: execvp: /bin/bash: Argument list too long
```

@daveshah1 I feel like I wrapped more code than necessary in PB_TYPE and I've added Makefile.v2x files are not quite needed. Any suggestions? or is my gut wrong?

mithro:
This all looks good to me. I'm going to merge it.

PR #76, opened 2018-03-09
Using bash as shell in Makefile

### LINKED ISSUE #66: Figure out why echo doesn't support '-e' on travis
On Travis we are getting `-e <string>` when we use `echo -e <string>`. On my local computer we just get `<string>` but echo is correctly interpreting the escape characters.

From the man page;
> `-e     enable interpretation of backslash escapes`

I'm guessing it has something to do with shell builtins verse real commands or something?

mithro:
Hopefully fixes #66.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
 addons:
   apt:
     packages:
+     - bash
      - colordiff
      - coreutils
      - graphviz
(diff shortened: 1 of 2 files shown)


mithro:
@elms - Will see if this works.

mithro:
https://travis-ci.org/SymbiFlow/symbiflow-arch-defs/builds/351175350

mithro:
Looks like it worked! Merging.

PR #79, opened 2018-03-09
Adding running of python tests

mithro:
* Uses pytest as the runner.
 * Runs the tests on travis.
 * Fixes the currently broken tests.


mithro:
Seems to be working;
image

PR #80, opened 2018-03-09
Use nodejs from conda

mithro:
Means people have one less dependency to install themselves.


mithro:
Looks like it's working -> https://travis-ci.org/mithro/symbiflow-arch-defs/builds/351538303

Issue #68, opened 2018-03-07
Make the dependency generated messages colorful


mithro:
Currently when a dependency file is created it outputs the following message with no colors;
 `Generated dependency info /home/travis/build/mithro/symbiflow-arch-defs/.deps//vpr/ff/vpr_ff.model.xml.dmk`.

It would be good if these messages were colorful like the other build messages. See image below;
![image](https://user-images.githubusercontent.com/21212/37111467-9970eb3c-21f4-11e8-8a82-5ef59358b10e.png)

mithro:
@bushig has a pull request for working on this here -> https://github.com/SymbiFlow/symbiflow-arch-defs/pull/83

litghost:
Obsolete

PR #83, opened 2018-03-12
Make the dependency generated messages colorful #68

bushig:


JohnDMcMaster:
Thank you for your interest! A few comments

1) All patches need DCO to be accepted, please see this for details: https://jjasghar.github.io/blog/2016/10/04/signing-commits-in-git/

2) Typically when curses are added there is also a check for isatty() as piping this to a file usually does not have the desired effect. If we add this, IMHO it should also be wrapped in a color function to improve readability. I am not a developer though for this part of the project, so I'd try to get ffeedback from one of them

mithro:
Hi @bushig,

Thank you for your pull request, it is awesome to have new people contributing!

I assume you working on fixing this github issue -> fix https://github.com/SymbiFlow/symbiflow-arch-defs/issues/68 ?

The idea behind the color is to highlight important information. For example if you look at how the existing color is used on by looking at the build on travis here -> https://travis-ci.org/SymbiFlow/symbiflow-arch-defs you can see it highlights the input and output values. We should aim for the same here.

Regarding @JohnDMcMaster 's comments, I recommend we use something like [termcolor](https://pypi.python.org/pypi/termcolor) and [colorama](https://pypi.python.org/pypi/colorama) for Python colorization.

Hope that helps! Looking forward to your contributions.

Tim 'mithro' Ansell

mithro:
Hi @bushig,

I'm assuming you probably got busy and haven't had time to look at this. I'm going to close the pull request for now. Please reopen it when/if you have time to finish of this work!

Tim 'mithro' Ansell

PR #85, opened 2018-03-29
Fix MUX generation and remove generated SLICEM merged.xml files

elmsfu:
A few one-off cleanup items I ran into while testing other things

### DIFF
--- a/utils/mux_gen.py
+++ b/utils/mux_gen.py
@@ -250,13 +250,13 @@ Generated with %s
             if args.split_inputs:
                 port_names.extend((mux_lib.MuxPinType.INPUT, args.name_inputs[j], 1, '[%i]' % j) for j in range(args.width))
             else:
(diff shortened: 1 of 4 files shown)


utils/mux_gen.py
```diff
@@ -250,13 +250,13 @@ def normpath(p, to=None):
             if args.split_inputs:
                 port_names.extend((mux_lib.MuxPinType.INPUT, args.name_inputs[j], 1, '[%i]' % j) for j in range(args.width))
             else:
-                port_names.append((mux_lib.MuxPinType.INPUT, args.name_input, args.width, '[%i:0]' % args.width))
+                port_names.append((mux_lib.MuxPinType.INPUT, args.name_input, args.width, '[%i:0]' % (args.width-1)))
```
mithro: Can you add a comment here? I think it should be something like "Verilog indexes are inclusive"?

PR #89, opened 2018-04-18
Various fixes to the PLB

jhol:
This patch set includes various improvements to the XML description of the PLB


mithro:
The changes in this pull request should be in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/90

PR #94, opened 2018-04-24
WIP rr graph lib

mcmasterg:
This PR is towards completing a baseline python rr_graph library to make custom rr_graphs beyond what arch.xml can do.

Current test suite is something like:
python -m lib.rr_graph.channel: passes
python -m lib.rr_graph.graph: passes
python utils/testarch_graph.py tests/build/testarch/2x4/wire.rr_graph.xml: fails due to missing channels (being investigated)

After stabilizing above to create the expected data structure, a relatively small amount of work needs to be done to add to_xml() functions to assorted XML elements. For the final library there might be a little work related to open VPR PR on mapping pin names to pins in the rr_graph pins (which is ambiguous in current format).


utils/lib/rr_graph/channel.py
```diff
@@ -14,100 +19,120 @@
 class ChannelNotStraight(TypeError):
     pass
 
+# TODO: clean up type
+# real graphs have length 1 tracks, which confuses auto detection
+_Track = namedtuple("Track", ("start", "end", "idx", "type_hint"))
```
mithro: I don't understand why `type_hint` is needed? Is it for "zero length" tracks?

utils/lib/rr_graph/channel.py
```diff
@@ -243,318 +268,516 @@ def __str__(self):
 
 
 # Nice short alias..
-C = Channel
+T = Track
 
 
 class ChannelGrid(dict):
+    '''
+    Functionality:
```
mithro: FYI - Docstring comments should be [reStructuredText](http://docutils.sourceforge.net/docs/ref/rst/introduction.html). 

If you write mostly markdown style it'll work okay.

utils/lib/rr_graph/channel.py
```diff
@@ -243,318 +268,516 @@ def __str__(self):
...
+    -Manages single type of track (either chanx or chany)
+    -Channel width along grid
+    -Manages track allocation within channels
+    -A track allocator
+
+    dict is indexed by Pos() objects
+    This returns a list indicaitng all the tracks at that position
```
mithro: speling

utils/lib/rr_graph/channel.py
```diff
@@ -243,318 +268,516 @@ def __str__(self):
...
-            for y in range(0, self.y):
+        for x in range(0, self.width):
+            for y in range(0, self.height):
                 self[Pos(x,y)] = []
 
     @property
-    def x(self):
-        return self.size.x
+    def width(self):
+        """Grid width
```
mithro: Maybe something like "Number of channels (as opposed to the width a an individual channel?)"

utils/lib/rr_graph/channel.py
```diff
@@ -243,318 +268,516 @@ def __str__(self):
...
+        [T(B,0), T(T,1), T(E,2), T(G,3)]
         >>> g[(6,5)]
-        [C(B,0), C(C,1), None, C(G,3)]
+        [T(B,0), T(T,1), None, T(G,3)]
         >>> g[(7,5)]
         [None, None, None, None]
         """
-        assert ch.idx == None
+        assert t.idx == None
+        force_idx = idx
```
mithro: I would probably have a `set_track` and a `add_track` -- only the set_track allows an idx value to be set?

utils/lib/rr_graph/channel.py
```diff
@@ -243,318 +268,516 @@ def __str__(self):
...
+            p[t.idx] = t
+        return t
 
     def pretty_print(self):
         """
-        If type == Channel.Type.X
+        If type == Track.Type.X
 
-          A--AC-C
+          A--AC-T
```
mithro: Search and replace fail....
`A--AC-C`

utils/lib/rr_graph/channel.py
```diff
@@ -243,318 +268,516 @@ def __str__(self):
...
+        return s
 
 
 if __name__ == "__main__":
     import doctest
+    print('doctest: begin')
     doctest.testmod()
+    print('doctest: end')
+
+    if 1:
```
mithro: Maybe these are better as unit tests?

utils/lib/rr_graph/graph.py
```diff
@@ -1,4 +1,44 @@
...
+    xml: can import but doesn't keep track of a node
+class Block(MostlyReadOnly):
+    for <grid_loc>
+class BlockGrid:
+    Was: BlockGraph
+    xml: nothing, handled by intneral Block objects though
+class GraphIdsMap:
+    holds pins + edges
+    xml: updated as pins are added
+        inconsistent with the rest of the project
```
mithro: FYI -- The reason for the inconsistency is that the other parts are not designed to allow modification/changing the XML.

utils/lib/rr_graph/graph.py
```diff
@@ -783,59 +864,25 @@ def block_types_for(self, col=None, row=None):
             ss.append(self.block_grid[pos].block_type)
         return ss
 
-    def __getitem__(self, pos):
-        return self.block_grid[pos]
-
-
-def simple_test_graph():
```
mithro: Where did this go?

utils/lib/rr_graph/graph.py
```diff
@@ -1312,13 +1480,302 @@ def add_channel(self, globalname, start, end, segtype, _chantype=None):
...
+        self.ids.add_edge(int(xtrack_node.get("id")), int(ytrack_node.get("id")), int(switch.get("id")))
+        self.ids.add_edge(int(ytrack_node.get("id")), int(xtrack_node.get("id")), int(switch.get("id")))
+
+    def create_xy_track(self, start, end, idx=None, type=None):
+        '''Create track object and corresponding nodes'''
+        track = self.channels.create_xy_track(start, end, idx=idx, type=type)
+        track_node = self.ids.add_node_for_track(track)
+        return track, track_node
+
+    def index_node_objects(self):
```
mithro: I don't get the point of this function?

utils/lib/rr_graph/channel.py
mcmasterg: I'm not paying any attention to docstring stuff right now. I'll sort that out later

mithro:
The pull request at https://github.com/SymbiFlow/symbiflow-arch-defs/pull/115 has replaced this one.

PR #96, opened 2018-04-25
testarch: More cleanup

mithro:
* Restructuring to match ice40 slightly.
 * Change `lutff` back to not have a carry.
 * Add `clutff` for the `lutff` with a carry.


mcmasterg:
I haven't reviewed in detail, but have no objection to merging

PR #98, opened 2018-04-26
ice40: Converted LUT primitive to match naming specification

jhol:


mithro:
Can you add a DCO and then I can merge?

jhol:
@mithro: done!

PR #103, opened 2018-04-28
Adding autoformatting

mithro:
Uses [yapf](https://github.com/google/yapf) in the `pep8` mode.


utils/conftest.py
```diff
@@ -1,7 +1,7 @@
 import sys
 
 collect_ignore = [
-    "icebox-rr_graph-import.py", # ImportError: No Module named 'icebox'
+    "icebox-rr_graph-import.py",  # ImportError: No Module named 'icebox'
```
mcmasterg: I renamed this on my local branch to only use _ for similar reasons (- isn't legal in a module name)

mithro: Fixed.

PR #115, opened 2018-05-11
Creating an library for working with rr_graph files

mithro:
* Initial prototype and design by @mithro.
 * Large amount of work from @JohnDMcMaster (see #94) to actually make it work.
 * Rework, fixing and documentation by @mithro.
 * Allows generating rr_graph from ice40 using icebox and hand coded approaches.


utils/lib/rr_graph/channel.py
```diff
@@ -629,12 +590,14 @@ def create_track(self, t, idx=None):
 
         # Find start and end
         # TODO: BI should maybe sort
```
JohnDMcMaster: this TODO has been fixed per the min/max below

utils/lib/rr_graph/graph.py
```diff
@@ -578,7 +542,28 @@ def _add_pin(self, pin):
...
+        Name of the block type.
+
+    size : Size
+        
+
+    pin_classes : tuple of PinClass
+        
+
+    pin_index : int
+        ????
```
JohnDMcMaster: Why ??? Do you need clarification on something? Or just didn't finish writing?

utils/lib/rr_graph/graph.py
```diff
@@ -637,23 +622,23 @@ def from_xml(cls, g, block_type_node):
         ... '''
         >>> bt = BlockType.from_xml(None, ET.fromstring(xml_string))
         >>> bt # doctest: +ELLIPSIS
-        BlockType(g=None, id=1, name='BLK_BB-VPR_PAD', size=Size(w=2, h=3), pin_classes=[...], pin_index={...})
+        BlockType(graph=None, id=1, name='BLK_BB-VPR_PAD', size=Size(w=2, h=3), pin_classes=[...], pin_index={...})
```
JohnDMcMaster: Library is/was named graph, caused some namespace conflicts

utils/lib/rr_graph/graph.py
```diff
@@ -1508,168 +1814,246 @@ def add_nodes_for_pin_class(self, block, pin_class, switch, sides):
...
-        Long term we need to reshuffle these instead of deleting and re-adding
-        '''
+    """
+
+    def __init__(
+            self,
+            rr_graph_file=None,
+            verbose=False,
+            clear_fabric=False,
+            sides=sides_always_right):
```
JohnDMcMaster: Thoughts on pushing back on VPR devs about side placement? I feel this is something we shouldn't have to track

JohnDMcMaster:
There are still 3 or so instances of node_index=None, remove them since this is no longer in the code

mithro:
> I thought jhol did this? Is this going to result in a merge conflict somewhere?

jhol did this for the `tile-routing` version. This does the same for the `top-routing`.

utils/lib/rr_graph/channel.py
mithro: Done.

utils/lib/rr_graph/graph.py
mithro: Fixed.

utils/lib/rr_graph/graph.py
mithro: It shouldn't be a problem inside this file?

utils/lib/rr_graph/graph.py
mithro: I think we should log a bug about this.

PR #129, opened 2018-05-21
rr_graph: remove pin index

mcmasterg:
Per
https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/319

Signed-off-by: John McMaster <johndmcmaster@gmail.com>

### DIFF
--- a/utils/lib/rr_graph/graph.py
+++ b/utils/lib/rr_graph/graph.py
@@ -189,9 +189,6 @@ class Pin(MostlyReadOnly):
 
     Attributes
     ----------
-    pin_class_index : int
-        Starts at 0 within a pin_class and increments for each pin.
-        Equal to the index property in the rr_graph XML.
 
     port_name : string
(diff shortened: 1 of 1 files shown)


mcmasterg:
What for? I would argue most people would want to reference it by name

mithro:
We need a way to tell the difference between the same pins in a pin class?

mcmasterg:
On a given Pin there is a "port_name" field. Why isn't this sufficient? I wouldn't be opposed to changing pins to a dictionary to make looking these up easier, but this check is no less strict than was already existing (ie in the old code it would not reject two pins with the same name) and IMHO no less functional.

Can you provide a specific example of where this scheme is problematic? Note I'm also running the ice40 generation to test this and it didn't seem to have any issues (make clean wire.patched.disp ARCH=ice40 DEVICE_TYPE=top-routing-virt)

PR #132, opened 2018-05-26
Add methods to RoutingGraph to determine whether any sink nodes are i…

benreynwar:
Add methods to RoutingGraph to determine if there are any source nodes that cannot access all the sink nodes.

It doesn't have a test yet, so don't merge it.


utils/lib/rr_graph/graph.py
```diff
@@ -1922,6 +1922,54 @@ def edges_for_node(self, xml_node):
...
+            return cache[node_id]
+        # Initially put a cache entry in saying that they are all inaccessible.
+        # That way if there is a loop we have a cache entry, but it won't effect anything.
+        cache[node_id] = set(target_node_ids)
+        # Initially start assuming all sink nodes are inaccessible.
+        inaccessible = set(target_node_ids)
+        node = self.get_node_by_id(node_id)
+        if node_id in inaccessible:
+            inaccessible.remove(node_id)
+        edges = self.edges_for_node(node)
```
mithro: See the docs for `self.edges_for_node`;
```
        **WARNING:** This function is quite slow when a graph has a large
        number of edges. If doing lots of lookups, consider using
        `edges_for_allnodes` function to get a dictionary that you can lookup
        the edges in directly.
```

mithro:
@benreynwar Thanks!

This is probably bests as a separate utility, maybe something like `utils/check_graph.py`?

benreynwar:
Tests showed, unsurpisingly that it was broken, so I had to rewrite it.

There's now two steps.
One to convert the cyclic graph into an acyclic graph (coarse-graining connected cycles into nodes).
Second step just determines connectivity between sources and sinks.

Tests for script are run by not passing a filename.
cyclic_to_acyclic.py has doctests.

None of these tests are being run in CI.

benreynwar:
It'd be good to test this on a couple of real rr_graph that you know what the answer should be, otherwise I think it's ready.

utils/lib/rr_graph/cyclic_to_acyclic.py
```diff
@@ -0,0 +1,201 @@
+class CycleSeas:
```
mithro: CycleSeas?

Does this terminology come from somewhere?

utils/lib/rr_graph/cyclic_to_acyclic.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Contains the cycleseas that have been found in the graph so far.
+
+    A cyclesea is a the set of nodes contained in intersecting cycles so
+    that all nodes are accessible from one another.
+    """
+
+    def __init__(self):
+        # A dictionry that maps cyclesea_id to the set of nodes.
+        self.cycleseas = {}
```
mithro: Dictionary

utils/lib/rr_graph/cyclic_to_acyclic.py
```diff
@@ -0,0 +1,201 @@
...
+        # A dictionary that maps a node_id to the cyclesea index that it is present in.
```
mithro: FYI We use NumPy style docstrings. These would be good to be part of that.

http://www.sphinx-doc.org/en/stable/ext/example_numpy.html?highlight=interface
https://numpydoc.readthedocs.io/en/latest/

utils/lib/rr_graph/cyclic_to_acyclic.py
benreynwar: Nope.  I had a crack at googling what the appropriate term was in graph theory, but didn't find anything.  So I made one up.

benreynwar: Correct term is : Strongly Connected Subgraph

benreynwar:
Closing this pull request.  It is replaced by #133.
The solution here is too complicated and too slow.

PR #162, opened 2018-07-04
tests: Moving tests around

mithro:
* Moves architecture specific stuff under each directory.
 * Puts all the makefiles in one common place. 
 * Makes the common tests work for each architecture.


mithro:
@elmsfu PTAL

ice40/make/tests.mk
```diff
@@ -61,16 +61,19 @@ endif
 # ---------------------------------------------
 # ---------------------------------------------
 
-INPUT_IO_FILE=$(TEST_DIR)/$(BOARD).pcf
+INPUT_IO_FILE=$(wildcard $(TEST_DIR)/$(BOARD).pcf)
```
elmsfu: It's not clear when we'd need wildcard. Is there a case when a * would be in the BOARD?

ice40/tests/iceff/Makefile
```diff
@@ -0,0 +1,3 @@
+ARCH=ice40
+DEVICE=HX1K
```
elmsfu: DEVICE=hx1k
Needs to be lowercase now I believe

ice40/tests/iceff/Makefile
```diff
@@ -0,0 +1,3 @@
...
+include ../../../make/tests.mk
```
elmsfu: Is PACKAGE required?

ice40/make/tests.mk
mithro: `$(wildcard xxx)` will return empty when the file doesn't exist.

elmsfu: I think this is clearer
`INPUT_IO_FILE=$(if $BOARD,$(TEST_DIR)/$(BOARD).pcf)`

tests/Makefile
```diff
@@ -1,6 +1,7 @@
 
 SUBDIRS=$(sort $(wildcard [0-9]-*))
-ARCHS=$(sort $(basename $(notdir $(wildcard make/arch/*.mk))))
+ARCHS=$(sort $(basename $(notdir $(wildcard ../*/make/tests.mk))))
+ARCHS=$(sort $(patsubst ../%/make/tests.mk,%,$(wildcard ../*/make/tests.mk)))
```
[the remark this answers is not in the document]
mithro: Done.

ice40/make/tests.mk
```diff
@@ -61,16 +61,19 @@ endif
...
 endif
 ifeq ($(PACKAGE),)
-$(error No $$PACKAGE set.)
+#$(error No $$PACKAGE set.)
+PACKAGE=qn84
 endif
 ifeq ($(PROG_TOOL),)
-$(error No $$PROG_TOOL set.)
+#$(error No $$PROG_TOOL set.)
+PROG_TOOL=true
```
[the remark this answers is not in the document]
mithro: Reverted.

ice40/tests/iceff/Makefile
mithro: Done.

ice40/tests/iceff/Makefile
mithro: Fixed.

PR #184, opened 2018-07-12
make/test: fix PLACE_TOOL and comment update

elmsfu:
io.place files weren't being generated because PLACE_TOOL wasn't being set due to order of variables being set

### DIFF
--- a/make/tests.mk
+++ b/make/tests.mk
@@ -43,6 +43,13 @@ always-run:
 # Fully qualified device name
 FQDN = $(ARCH)-$(DEVICE_TYPE)-$(DEVICE)
 
+TEST_DIR = $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
+OUT_LOCAL = $(TEST_DIR)/build-$(FQDN)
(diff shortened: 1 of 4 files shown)


make/tests.mk
```diff
@@ -43,6 +43,13 @@ always-run:
 # Fully qualified device name
 FQDN = $(ARCH)-$(DEVICE_TYPE)-$(DEVICE)
 
+TEST_DIR = $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
```
mithro: There is a `TESTS_MK_DIR` up on line 1?

make/tests.mk
```diff
@@ -43,6 +43,13 @@ always-run:
...
+OUT_LOCAL = $(TEST_DIR)/build-$(FQDN)
```
mithro: Should this be defined in `$(TOP_DIR)/$(ARCH)/make/tests.mk` file then?

elmsfu: There is a interdependence now so I don't that would work. I think we could split the arch into a pre and post mk files and it would be cleaner.

PR #198, opened 2018-07-16
ecp5: Simulation model fixes

daveshah1:
- Change how LUTs are implemented so that "don't care" inputs at 'X' don't force the output to 'X' 
 - Fix missing "cemux" conditional in TRELLIS_FF
 - Other small tweaks


mithro:
LGTM!

PR #213, opened 2018-07-27
ice40: picosoc add extra make targets.

mithro:


ice40/tests/picosoc/Makefile
```diff
@@ -15,3 +15,18 @@ TESTBENCH_SOURCES = \
...
 
 include ../../../make/tests.mk
+
+firmware.hex: firmware.elf
+	riscv32-unknown-elf-objcopy -O verilog firmware.elf /dev/stdout | sed -e '1 s/@00000000/@00100000/; 2,65537 d;' > firmware.hex
+
+firmware.bin: firmware.elf
+	riscv32-unknown-elf-objcopy -O binary firmware.elf /dev/stdout | tail -c +1048577 > firmware.bin
+
+prog-firmware: firmware.hex
```
elmsfu: s/hex/bin/

mithro: You mean check in *both* the `.bin` and `.hex` rather than the `.elf`?

elmsfu: oh right the hex is needed for simulation... Then I guess I'd vote for adding all the rules to generate them from source.

litghost:
riscv32-binutils tools are now in conda, this pull request can be converted to CMake now.

Issue #177, opened 2018-07-10
Remove the 'virt' from the ice40 device name


mithro:
The ice40 is now using a real routing graph rather than a "virtual" routing graph.

Probably a good idea to no longer have `virt` in the name.

@JohnDMcMaster requested this.

PR #215, opened 2018-07-28
CMake build system for architecture definitions and FPGA targets

litghost:
Working version of CMake build system that can build ice40 targets and should be extensible for artix7 targets and beyond.

This does not replace "make env", conda setup is still done via Make.


CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.0)
+project(symbiflow-arch-defs)
+
+include(make/util.cmake)
```
mithro: nit: Should this directory be called `cmake` now?

artix7/primitives/dsp48e1/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+add_subdirectory(nreg)
```
mithro: Is there a way to auto-generate the subdirectory stuff?

artix7/primitives/dsp48e1/nreg/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+make_file_target(FILE reg.sim.v SCANNER_TYPE verilog)
+
+make_file_target(FILE nreg.sim.v SCANNER_TYPE verilog)
+v2x(NAME nreg SRCS nreg.sim.v)
+
+v2x(NAME reg SRCS reg.sim.v)
```
mithro: Should this be sorted something like;

```cmake
make_file_target(FILE reg.sim.v SCANNER_TYPE verilog)
v2x(NAME reg SRCS reg.sim.v)

make_file_target(FILE nreg.sim.v SCANNER_TYPE verilog)
v2x(NAME nreg SRCS nreg.sim.v)
```

or

```cmake
make_file_target(FILE reg.sim.v SCANNER_TYPE verilog)
make_file_target(FILE nreg.sim.v SCANNER_TYPE verilog)
v2x(NAME reg SRCS reg.sim.v)
v2x(NAME nreg SRCS nreg.sim.v)
```

artix7/primitives/slicel/N5ff/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+make_file_target(FILE ntemplate.N5ff.model.xml)
+
+make_file_target(FILE ntemplate.N5ff.sim.v SCANNER_TYPE verilog)
+
+make_file_target(FILE ntemplate.N5ff.pb_type.xml)
+n_template(NAME N5ff.model.xml PREFIXES a;b;c;d SRCS ntemplate.N5ff.model.xml)
+
+n_template(
+  NAME N5ff.pb_type.xml
+  PREFIXES a;b;c;d
```
mithro: Why does this use semicolons while `mux_gen` uses commas?

ice40/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+  BITSTREAM_EXTENSION asc
+  BIT_TO_BIN ${ICESTORM}/icepack/icepack
+  BIT_TO_BIN_CMD "\${BIT_TO_BIN} \${OUT_BITSTREAM} > \${OUT_BIN}"
+)
+
+add_subdirectory(primitives)
+add_subdirectory(cells)
+add_subdirectory(devices)
+
+define_board(BOARD icestick DEVICE hx1k PACKAGE tq144 PROG_TOOL ${ICEPROG_TOOL})
```
mithro: Could the boards be defined in a separate file which only has boards?

make/cmake_format.py
```diff
@@ -0,0 +1,137 @@
+# How wide to allow formatted cmake files
+line_width = 80
```
mithro: It seems like there are lines which are longer than 80 characters?

make/devices.cmake
```diff
@@ -0,0 +1,754 @@
+# This CMake include defines the following functions:
```
mithro: Do you know if these comments can be automatically extracted and put into sphinx documentations in some way?

 - https://eb2.co/blog/2012/03/sphinx-and-cmake-beautiful-documentation-for-c---projects/
 - https://github.com/Kitware/CMake/tree/master/Utilities/Sphinx
 - https://cmake.org/pipermail/cmake/2017-June/065557.html

FYI I start trying to add sphinx integration here -> https://github.com/SymbiFlow/symbiflow-arch-defs/pull/44

make/devices.cmake
```diff
@@ -0,0 +1,754 @@
...
+  #   DEVICE_TYPE <device_type>
+  #   PACKAGES <list of packages>
+  #   )
+  # ~~~
+  #
+  # Defines a device within a specified FPGA architecture.
+  #
+  # Creates dummy targets <arch>_<device>_<package>_rrxml_virt and
+  # <arch>_<device>_<package>_rrxml_virt  that generates the the virtual and
+  # real rr_graph for a specific device and package.
```
mithro: FYI - We want to get rid of the `-virt` in the device names sometime -> https://github.com/SymbiFlow/symbiflow-arch-defs/issues/177

make/devices.cmake
```diff
@@ -0,0 +1,754 @@
...
+  # ~~~
+  # ${CMAKE_CURRENT_BINARY_DIR}/${NAME}/${ARCH}-${DEVICE_TYPE}-${DEVICE}-${PACKAGE}
+  # ~~~
+  #
+  # Output files:
+  #
+  # * ${TOP}.eblif - Synthesized design
+  # * ${TOP}_io.place - IO placement.
+  # * ${TOP}.route - Place and routed design
+  # * ${TOP}.hlc - Place and routed design
```
mithro: HLC files are specific to iCE40. It is likely that `.fasm` files will be the generic output in the future.

Might be worth linking to the VPR documentation for each file type?

make/devices.cmake
```diff
@@ -0,0 +1,754 @@
...
+        ${INPUT_IO_FILE}
+        ${INPUT_IO_FILE_TARGET}
+        ${PINMAP}
+        ${PINMAP_TARGET}
+        ${VPR_DEPS}
+      COMMAND ${PLACE_TOOL_CMD_FOR_TARGET_LIST} --out ${OUT_IO}
+      WORKING_DIRECTORY ${OUT_LOCAL}
+    )
+
+    set(VPR_CMD ${VPR_CMD} --fix_pins ${OUT_IO})
```
mithro: You can't give `--fix_pins` to `--pack` IIRC?

make/devices.cmake
```diff
@@ -0,0 +1,754 @@
...
+    add_output_to_fpga_target(${NAME} IO_PLACE ${OUT_LOCAL_REL}/${TOP}_io.place)
+  endif()
+
+  # Generate packing.
+  # -------------------------------------------------------------------------
+  set(OUT_NET ${OUT_LOCAL}/${TOP}.net)
+  add_custom_command(
+    OUTPUT ${OUT_NET}
+    DEPENDS ${OUT_EBLIF} ${OUT_IO} ${VPR_DEPS}
+    COMMAND ${VPR_CMD} --pack --place
```
mithro: Ideally we wouldn't be giving `--place` to this step but I think we need to because of the `--fix_pins` argument?

make/file_targets.cmake
```diff
@@ -0,0 +1,265 @@
...
+        BASE_DIR
+        ${symbiflow-arch-defs_SOURCE_DIR}
+      )
+      list(APPEND ${var} ${ABS_PATH_TO_INCLUDE})
+    endif()
+  endforeach()
+  set(${var} "${${var}}" PARENT_SCOPE)
+endfunction()
+
+function(MAKE_FILE_TARGET)
```
mithro: Maybe this should be `CREATE_FILE_TARGET` -- I keep getting it confused is `makefile target` or something?

utils/deps_xml.py
```diff
@@ -16,29 +16,37 @@
 parser = argparse.ArgumentParser()
 parser.add_argument(
     "inputfile", type=argparse.FileType('r'), help="Input XML file")
+parser.add_argument(
+    "--file_per_line", action='store_true', help="Output dependencies file per line, rather than Make .d format.")
```
mithro: Maybe we just drop the `Make .d` format (or log an issue / add FIXME comment to remove it)?

mithro:
Is the pull request description still accurate? It looks like this had some conda stuff in it already?

utils/mux_gen.py
```diff
@@ -207,11 +204,6 @@ def normpath(p, to=None):
     # XML Files can't have "--" in them, so instead we use ~~
     xml_comment = generated_with.replace("--", "~~")
 
-    # ------------------------------------------------------------------------
-    # Create a makefile to regenerate files.
```
elmsfu: @mithro Do we still want a way to output the parameters that were used to a file or stderr?

make/devices.cmake
```diff
@@ -0,0 +1,754 @@
...
+  # implicitly passed to MAKE_FILE_TARGET.  If EXPLICIT_MAKE_FILE_TARGET is
+  # supplied, this behavior is supressed.
+  #
+  # The SOURCES file list will be used to synthesize the FPGA images.
+  # INPUT_IO_FILE is required to define an io map. TESTBENCH_SOURCES will be
+  # used to run test benches.
+  #
+  # Targets generated:
+  #
+  # * <name>_eblif - Generate eblif file.
```
elmsfu: Should these be <name>_<board>_ so we can build the same project for multiple boards? like we can with `make BOARD=icevision` for example

CMakeLists.txt
litghost: We'll leave the old Make system in place until everything is working, so I'm fine leaving it as make for now.  I don't think anyone is gonna be confused.

artix7/primitives/dsp48e1/CMakeLists.txt
litghost: Agreed, explicit over implicit.  Globing in CMake has unintended consequences, so lets avoid it if possible.

artix7/primitives/slicel/N5ff/CMakeLists.txt
litghost: Semicolons in CMake are special, they are the list divider.  So in this case PREFIXES is a list that CMake operates on (eg foreach(PREFIX ${PREFIXES}).  The commas are passed literally to mux_gen.  We could convert all lists to semicolons, and then convert the list to commas later, but that doesn't seem excessively valuable.

ice40/CMakeLists.txt
litghost: Done

make/cmake_format.py
litghost: cmake_format.py isn't perfect and CMake lacks good line continuation options in some cases.  I'm actually wary of further use of cmake_format.

make/devices.cmake
litghost: If I convert the comments to Sphinx comments it might work.  A task for a different pull I think.

make/devices.cmake
litghost: Ack

make/devices.cmake
litghost: I linked to eblif and .route format links in VPR

make/devices.cmake
litghost: This was from old make system, I'll update it.

make/devices.cmake
litghost: This was from the old make system, I'll see if I can --pack without --place if I only use the fix-pins argument on --place

artix7/primitives/slicel/muxes/f7amux/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+mux_gen(
+  NAME f7amux
+  COMMENT "F7AMUX is used to create a LUT7 from two LUT6s."
+  MUX_NAME F7AMUX
+  OUTPUT OUT
+  SPLIT_INPUTS
```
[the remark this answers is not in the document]
litghost: Opps, fixed

make/devices.cmake
litghost: I think the answer applies to the answer about add_subdirectory above, explicit of implicit.  If you support multiple boards, you would call add_fpga_target for each board.  I think only the synthesis and pack is reusable, the place and route will have to be regenerated.  If we think there is a lot of value in sharing the synthesis and pack output, I can refactor to support that.

litghost:
Description is still accurate.  Conda env is detected by this system, but it doesn't set it up like make env does.

artix7/primitives/slicel/N5ff/CMakeLists.txt
mithro: I feel like the difference is going to bite me in the future :-P

make/cmake_format.py
mithro: I'm strongly in support of having files get auto formatted.

I'd prefer *all* files be auto formatted if possible.

make/devices.cmake
mithro: Yeap, can you add a github issue for it?

make/devices.cmake
mithro: FYI, for each design board support is implied by the presence of a `.pcf` file for a board.

litghost: Explicit over implicit here.  For example, you might want to have multiple .pcf for a particular board.

utils/mux_gen.py
litghost: Not sure what you are suggesting here.  I could output the parameters used to define the mux into the verilog file.  Would that match your request?

artix7/primitives/slicel/N5ff/CMakeLists.txt
litghost: Agreed :( CMake's syntax is not pretty.  If this becomes and issue in the future, I can convert all lists in arguments to semi-colons and convert to commas when needed.

utils/deps_xml.py
litghost: File #218

artix7/primitives/slicel/N5ff/CMakeLists.txt
mithro: Lets log an issue and plan to do this? It can be a good intro task?

make/devices.cmake
mithro: By reusable do you mean the commands to run, or the output from the commands?

I really like that the current system forces you to name the PCF file exactly the same in each directory. It means we don't end up with people calling the same boards different names....

artix7/primitives/slicel/N5ff/CMakeLists.txt
litghost: #219

ice40/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+set(ICESTORM ${symbiflow-arch-defs_SOURCE_DIR}/third_party/icestorm)
+set(ICEBOX ${ICESTORM}/icebox/)
+set(ICEPROG_TOOL ${ICESTORM}/iceprog/iceprog)
+
+define_arch(
+  ARCH ice40
+  YOSYS_SCRIPT
+    "synth_ice40 -nocarry $<SEMICOLON>  ice40_opt -unlut $<SEMICOLON> abc -lut 4 $<SEMICOLON>  opt_clean"
```
mithro: I think @JohnDMcMaster showed that `synth_ice40 -vpr` was actually enough here now? Let's investigate after landing....

utils/mux_gen.py
mithro: @elmsfu Was suggesting that the tool have a way to generate the `CMakeLists.txt` file as the output.

PR #222, opened 2018-08-08
Add place-and-route and equivalence tests for ice40 targets

litghost:
Supports CTest (slow but nice output) and just using make (fast but merged output).  If there is a better way to handle CTest build test inputs first, do let me know.

CTest: make test
Make: make -j64 all_check_tests


ice40/CMakeLists.txt
```diff
@@ -19,12 +19,10 @@ define_arch(
   --blif \${OUT_EBLIF} \
   --pcf \${INPUT_IO_FILE}"
   CELLS_SIM ice40/cells_sim.v
-  EQUIV_CHECK_SCRIPT
-    "hierarchy; proc; clk2fflogic; miter -equiv -flatten -ignore_gold_x -make_outputs -make_outcmp gold gate miter; sat -verify-no-timeout -timeout 600 -seq 1000 -prove trigger 0 -prove-skip 1 -show-inputs -show-outputs miter"
   HLC_TO_BIT ${ICEBOX}/icebox_hlc2asc.py
   HLC_TO_BIT_CMD "\${HLC_TO_BIT} \${OUT_HLC} > \${OUT_BITSTREAM}"
   BIT_TO_V ${ICEBOX}/icebox_vlog.py
-  BIT_TO_V_CMD "\${BIT_TO_V} -D -c -n \${TOP} -p \${INPUT_IO_FILE} -d \${PACKAGE} \${OUT_BITSTREAM} > \${OUT_BIT_VERILOG} || rm \${OUT_BIT_VERILOG}"
+  BIT_TO_V_CMD "\${BIT_TO_V} -D -c -n \${TOP} -p \${INPUT_IO_FILE} -d \${PACKAGE} \${OUT_BITSTREAM} > \${OUT_BIT_VERILOG}"
```
mithro: FYI this tool writes a partial output file when it fails.

litghost: CMake will delete output on failed generate targets as of CMake 3.3 https://stackoverflow.com/questions/41597612/add-custom-command-does-not-re-run-on-failure

make/devices.cmake
```diff
@@ -698,9 +702,160 @@ function(ADD_FPGA_TARGET)
...
+endfunction()
+
+function(get_equiv_check_script var type)
+  # get_equiv_check_script sets var to a common EQUIV_CHECK_SCRIPT types.
+  #
+  # type can be one of the following:
+  #
+  #  * miter_and_tempinduct - Prove equivilence by creating a miter between
+  #    gold and gate, and then temporality inducting. This works on simple
+  #    cicruits.
```
mithro: `circuits`

make/devices.cmake
```diff
@@ -698,9 +702,160 @@ function(ADD_FPGA_TARGET)
...
+  #  * miter_and_tempinduct_skip_1 - Same as miter_and_tempinduct, but skips
+  #    first step.  This handles cases where the first clock is a reset step.
+  #    Outputs for the first step are undefined, but fine for the second step.
+
+  # See for more potential ideas on example EQUIV_CHECK_SCRIPT's:
+  # https://github.com/YosysHQ/yosys/issues/302#issuecomment-276019637
+  if("${type}" STREQUAL "miter_and_tempinduct")
+    set(${var}
+      "hierarchy $<SEMICOLON> proc $<SEMICOLON> clk2fflogic $<SEMICOLON> miter -equiv -flatten -ignore_gold_x -make_outputs -make_outcmp gold gate miter $<SEMICOLON> sat -prove trigger 0 -verify -tempinduct miter"
```
mithro: Are we better maybe storing these in Yosys script files?
http://www.clifford.at/yosys/cmd_script.html

tests/4-mlut/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# TODO: Add support for eblif based equivilence tests.
```
mithro: Hrm? I thought you had fixed that?

make/devices.cmake
litghost: Fixed

make/devices.cmake
litghost: Agreed

tests/4-mlut/CMakeLists.txt
litghost: It's a CMake TODO issue.  I removed the logic that was present in the Makefile system to ingest eblif as input.  Need to think about how to add it back.   I was thinking about refactoring the synthesis from the place and route.

tests/6-led/led.v
```diff
@@ -2,21 +2,19 @@ module top (
 	input  clk,
```
mithro: I think we should keep LED as something you could actually run on hardware? We already have a counter test at https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/tests/6-counter/counter.v ?

tests/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
+set(BOARDS "")
+list(APPEND BOARDS icestick)
```
mithro: This doesn't seem right? I think we want a fake / empty board?

yosys/equiv_simple_opt_full.yosys
```diff
@@ -0,0 +1,9 @@
+hierarchy
+proc
+equiv_make gold gate equiv
+prep -flatten -top equiv
+opt_clean -purge
+opt -full
+equiv_simple -seq 10
```
mithro: What does the `-seq 10` mean? A comment might be useful?

yosys/equiv_simple_clk2fflogic.yosys
```diff
@@ -0,0 +1,10 @@
+hierarchy
```
mithro: Does Yosys support comments in the script files?

If so could we get one at the top of each of these files?

mithro: BTW Yosys seems to use `xxx.ys` as the file extension for it's scripts...

yosys/miter_and_tempinduct_large_N.yosys
```diff
@@ -0,0 +1,5 @@
+hierarchy
+proc
+clk2fflogic
+miter -equiv -flatten -ignore_gold_x -make_outputs -make_outcmp gold gate miter
+sat -prove trigger 0 -verify -tempinduct -tempinduct-skip 1 -initsteps 32 -stepsize 16 miter
```
mithro: Do the `-initsteps` and `-stepsize` values matter at all?

utils/create_pinmap.py
```diff
@@ -0,0 +1,40 @@
...
+  pinidx = 0
+
+  with open(args.design_json) as f:
+    j = json.load(f)
+    for port in get_module_ports(j, args.module):
+      print('set_io %s %s' % (port, pins[pinidx]))
+      pinidx += 1
+
+if __name__ == '__main__':
+  main()
```
mithro: pinmap == pcf file? Should we just generate an empty `.ioplace` file instead?

BTW It might have been easier to use the EBLIF file like the [other tool does](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/ice40/utils/ice40_create_ioplace.py)?

I wonder if [ice40_create_ioplace.py](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/ice40/utils/ice40_create_ioplace.py) could just have a `--allow-no-pcf` argument?

Feel free to just log a bug to rework this at a later date?

make/devices.cmake
```diff
@@ -736,19 +855,94 @@ function(add_testbench)
...
+  #     SOURCES <list of sources to load>
+  #   )
+  # ~~~
+  #
+  # Generate pinmap blindly assigns each input and output from the module
+  # ${TOP} to valid pins for the specified board. In its current version,
+  # GENERATE_PINMAP may assign IO to global wire.
+  #
+  # TODO: Consider adding knowledge of global wires and be able to assign
+  # specific wires to global wires (e.g. clock or reset lines).
```
mithro: Maybe log an issue about this? Could probably be a good trivial "first issue" for someone -- adding support for [improving the layout csv output](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/ice40/utils/ice40_import_layout_from_icebox.py) would be the first step I think?

mithro:
Could you rebase on top of master?

Otherwise LGTM with some minor comments.

yosys/equiv_simple_clk2fflogic.yosys
litghost: Done

yosys/miter_and_tempinduct_large_N.yosys
litghost: Yes, I'll add a comment

make/devices.cmake
litghost: #224

utils/create_pinmap.py
litghost: Having a pinmap enables both check's on bit_to_v and testbinch's.

tests/CMakeLists.txt
mithro: SGTM

utils/create_pinmap.py
mithro: Yeap, I had missed that. Lets go ahead with this.

yosys/equiv_simple_clk2fflogic.yosys
litghost: Done

PR #227, opened 2018-08-09
Add support for designs that target multiple boards.

litghost:
TinyFPGA BX and TinyFPGA B2 support are still missing, will be added in a follow up CL.  They require adding pip install hooks through conda.

### DIFF
--- a/ice40/boards.cmake
+++ b/ice40/boards.cmake
@@ -5,7 +5,7 @@ define_board(
   DEVICE lp1k
   PACKAGE qn84
   PROG_TOOL ${ICEPROG_TOOL}
-  PROG_CMD ${ICEPROG_TOOL} -ew
+  PROG_CMD "${ICEPROG_TOOL} -ew"
 )
 
 define_board(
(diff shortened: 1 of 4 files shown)


ice40/boards.cmake
```diff
@@ -5,14 +5,21 @@ define_board(
...
-  PROG_CMD ${ICEPROG_TOOL} -S
+  PROG_CMD "${ICEPROG_TOOL} -S"
+)
+
+define_board(
+  BOARD icevision
+  DEVICE up5k
+  PACKAGE sg48
+  PROG_TOOL ${ICEPROG_TOOL}
 )
```
mithro: I think your still missing a couple of boards from https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/ice40/make/boards.mk?

Can you port across the comments too?

mithro: I think we should probably also have an empty / dummy board?

litghost: Added dummy board.

PR #229, opened 2018-08-09
Add VPR echo targets and VPR_EXTRA_ARGS cache variables.

litghost:


mithro:
As long as VPR_EXTRA_ARGS doesn't apply to the rr_graph.xml generation vpr runs, then LGTM!

PR #230, opened 2018-08-10
Add working testarch support to CMake build.

litghost:


make/devices.cmake
```diff
@@ -16,8 +16,10 @@ function(DEFINE_ARCH)
   #    BITSTREAM_EXTENSION <ext>
   #    RR_PATCH_TOOL <path to rr_patch tool>
   #    RR_PATCH_CMD <command to run RR_PATCH_TOOL>
+  #    [NO_PLACE]
```
mithro: No place? -- Should this be something like "no pins" or something? The pnr still goes through a placement phase right?

common/xml/fpga_architecture.xsd
```diff
@@ -293,28 +293,16 @@
     </xs:restriction>
   </xs:simpleType>
 
-  <xs:simpleType name="pb_io_equivalent">
```
mithro: I think this property is still valid, just not where we currently have it?
http://vtr-verilog-to-routing.readthedocs.io/en/latest/arch/reference.html#tag-pb-type-input

litghost: Updated

PR #233, opened 2018-08-10
Add documentation for new CMake build system.

litghost:


cmake.md
```diff
@@ -0,0 +1,30 @@
+## CMake build system
+### Terminology
+
+The CMake system in symbiflow-arch-defs uses the following hierarchy:
+
+* ARCH - The architecture of FPGA.  This is used to capture architecture level parameters (e.g. tech libraries).
```
mithro: Architecture is the high level grouping of a set of "similar" FPGA types. An example would be the "Lattice iCE40 Series" (`ice40`) or the "Xilinx Artix 7 Series" (`artix7`).

cmake.md
```diff
@@ -0,0 +1,30 @@
...
+* DEVICE_TYPE - A type of device.  This is defined as an ARCH and the arch.xml which defines the units with this device type.
```
mithro: This refers to the `tile-routing-virt` and `top-routing-virt` right?

I feel like we ended up choosing a bad name for this. It would be better characterized as something like "architecture representation". Ultimately the tile and top routing should produce identical results.

For example,
 * tile routing means that a bunch of the routing is contained inside the tile definition. This is things like the local wires, global2local and similar stuff which is highly associated with the tile. Only the multi-tile spanning wires such as span4/span12/global wires are represented at the "fabric level".
 * top routing means that all the "routing" is configured to be at the "fabric level".

I guess this kind of falls apart a bit with the testarch?

Layouts might need to be special in some way, for example the `tile-routing-virt` and `top-routing-vert` have (and probably should always have) the same set of layouts?

cmake.md
```diff
@@ -0,0 +1,30 @@
...
+* DEVICE - A specific instance of a DEVICE_TYPE.  A device definition is a DEVICE_TYPE, a PACKAGE list, and IO map definitions for each package if needed.
```
mithro: Does `PACKAGE` == the layouts inside the arch.xml file?

cmake.md
```diff
@@ -0,0 +1,30 @@
...
+* BOARD - A specific DEVICE and PACKAGE instance, along with a PROG_TOOL/PROG_CMD.  A board is what is targeted by ADD_FPGA_TARGET.
```
mithro: The last sentence is a bit strange to me? Maybe I'm just confused by the name `ADD_FPGA_TARGET`.

cmake.md
```diff
@@ -0,0 +1,30 @@
+## CMake build system
```
mithro: I would probably  wrap the all caps names in backticks. This prevents any weirdness around underscores getting parsed as underline and makes it monospace / stand out as something special.

cmake.md
```diff
@@ -0,0 +1,30 @@
...
+### Terminology
+
+The CMake system in symbiflow-arch-defs uses the following hierarchy:
+
+* ARCH - The architecture of FPGA.  This is used to capture architecture level parameters (e.g. tech libraries).
+* DEVICE_TYPE - A type of device.  This is defined as an ARCH and the arch.xml which defines the units with this device type.
+* DEVICE - A specific instance of a DEVICE_TYPE.  A device definition is a DEVICE_TYPE, a PACKAGE list, and IO map definitions for each package if needed.
+* BOARD - A specific DEVICE and PACKAGE instance, along with a PROG_TOOL/PROG_CMD.  A board is what is targeted by ADD_FPGA_TARGET.
+
+### Creating a new FPGA target
```
mithro: I'm not sure what "creating a new FPGA target" actually means? 

I think this means adding a new test (which will run pnr for an FPGA)?

cmake.md
```diff
@@ -0,0 +1,30 @@
...
+For non-generated verilog, ADD_FILE_TARGET should be invoked like:
+```
+add_file_target(FILE <current source dir relative path> SCANNER_TYPE verilog)
+```
+
+For generated verilog, ADD_FILE_TARGET should be invoked like:
+```
+add_file_target(FILE <current source dir relative path> GENERATED)
+```
+Note that the input path must be current source dir relative, not absolute, and not relative to the current binary dir.  Also note that CMake cannot take depedendencies during build time, so for generated files you must supply the dependencies for the target.  See (APPEND_FILE_DEPENDENCY)[https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/make/file_targets.cmake#L79] for adding depedencies to other file targets.  It is generally suggested that all source forms within symbiflow-arch-defs are file targets, but it is required for verilog files because of how verilog handles relative include paths.
```
mithro: > not relative to the current binary dir.

binary dir == directory you are currently building in?

mithro: I don't quite understand this last bit;
>  It is generally suggested that all source forms within symbiflow-arch-defs are file targets, but it is required for verilog files because of how verilog handles relative include paths.

cmake.md
mithro: Some examples are always good at clarifying english :-)

cmake.md
```diff
@@ -0,0 +1,30 @@
...
+* ARCH - The architecture of FPGA.  This is used to capture architecture level parameters (e.g. tech libraries).
+* DEVICE_TYPE - A type of device.  This is defined as an ARCH and the arch.xml which defines the units with this device type.
+* DEVICE - A specific instance of a DEVICE_TYPE.  A device definition is a DEVICE_TYPE, a PACKAGE list, and IO map definitions for each package if needed.
+* BOARD - A specific DEVICE and PACKAGE instance, along with a PROG_TOOL/PROG_CMD.  A board is what is targeted by ADD_FPGA_TARGET.
+
+### Creating a new FPGA target
+
+Assuming that the board(s) you are interested in targetting is already defined, ADD_FPGA_TARGET is what will create targets.  See [ADD_FPGA_TARGET documentation](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/make/devices.cmake#L559).
+
+If you want to target multiple boards, there is also ADD_FPGA_TARGET_BOARDS that will call ADD_FPGA_TARGET correctly for each board you request.  See [ADD_FPGA_TARGET_BOARDS](https://github.com/litghost/symbiflow-arch-defs/blob/re_enable_artix7_routing_tests/make/devices.cmake#L458).
```
mithro: Link to your own repository?

cmake.md
```diff
@@ -0,0 +1,30 @@
...
+The CMake system in symbiflow-arch-defs uses the following hierarchy:
+
+* ARCH - The architecture of FPGA.  This is used to capture architecture level parameters (e.g. tech libraries).
+* DEVICE_TYPE - A type of device.  This is defined as an ARCH and the arch.xml which defines the units with this device type.
+* DEVICE - A specific instance of a DEVICE_TYPE.  A device definition is a DEVICE_TYPE, a PACKAGE list, and IO map definitions for each package if needed.
+* BOARD - A specific DEVICE and PACKAGE instance, along with a PROG_TOOL/PROG_CMD.  A board is what is targeted by ADD_FPGA_TARGET.
+
+### Creating a new FPGA target
+
+Assuming that the board(s) you are interested in targetting is already defined, ADD_FPGA_TARGET is what will create targets.  See [ADD_FPGA_TARGET documentation](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/make/devices.cmake#L559).
```
mithro: FYI You should be able to use a relative link here. IE
`[ADD_FPGA_TARGET documentation](../make/devices.cmake#L559)`

cmake.md
litghost: Yes

litghost: Verilog defines relative paths as relative to the location of the file, not to the location of the compiler.  As a result, if you use relative includes to include both a generated and non-generated file, things get exciting in out of source builds.  add_file_target solves this by copying non-generated files to the binary directory.

cmake.md
litghost: When you say layout, what do you mean?

cmake.md
litghost: A package really pindefs, but the package is part of the --device during the rr_graph generation step.  The current CMake system assumes that each package needs its own rr_graph based on this argument.  So currently a package is a rr_graph and pindef map.

cmake.md
litghost: ADD_FPGA_TARGET is our version of CMake's ADD_EXECUTABLE.  An ADD_FPGA_TARGET targets a board (e.g. build these sources for this board).

cmake.md
litghost: Fixed

cmake.md
litghost: Updated.

cmake.md
litghost: Done

cmake.md
litghost: Fixed

cmake.md
litghost: Fixed

cmake.md
litghost: Updated doc.

cmake.md
```diff
@@ -0,0 +1,63 @@
...
+* `DEVICE_TYPE` - A type of device.  This is defined as an `ARCH` and the
+  arch.xml which defines the units with this device type.
+* `DEVICE` - A specific instance of a `DEVICE_TYPE`.  A device definition is
+   a `DEVICE_TYPE`, a `PACKAGE` list, and IO map definitions for each package
+   if needed.  Each `DEVICE` should coorispond with a layout definition within
+   the `DEVICE_TYPE`.
+* `BOARD` - A specific `DEVICE` and `PACKAGE` instance, along with a
+  `PROG_TOOL`/`PROG_CMD` (e.g. commands to program a specified board).
+
+### Creating a new FPGA target
```
mithro: `### Adding new tests`?

cmake.md
```diff
@@ -0,0 +1,63 @@
...
+
+Assuming that the board(s) you are interested in targetting is already
```
mithro: Maybe something like this?

```markdown
### Adding new tests

If you are familiar with cmake, `ADD_FPGA_TARGET` is the rough equivalent to the normal `ADD_EXECUTABLE` target you would use with a C project. 

`ADD_FPGA_TARGET` will create the targets needed to take a design from Verilog to bitstream through the synthesis, place and route, bitstream and potentially even programming. The required arguments are the `BOARD` and input files see
[`ADD_FPGA_TARGET` documentation](make/devices.cmake#L559) for further information on other target configuration options.

```

cmake.md
```diff
@@ -0,0 +1,63 @@
...
+[`ADD_FPGA_TARGET_BOARDS` documentation](make/devices.cmake#L458).
+
+#### Note on `ADD_FILE_TARGET` and `ADD_FPGA_BOARD`
+
+All source files in the symbiflow-arch-defs are required to have a file target
+associated with them.  This is done via
+[`ADD_FILE_TARGET`](make/file_targets.cmake#L193).  By default
+`ADD_FPGA_TARGET` and `ADD_FPGA_TARGET_BOARDS` will both implicitly invoke
+`ADD_FILE_TARGET` for you.  This handles the common case where input sources
+are not generated files.  However, if you want to generate files, then you
```
mithro: Maybe?

`However, if you are using generated files, then you`

cmake.md
```diff
@@ -0,0 +1,63 @@
...
+For non-generated verilog, `ADD_FILE_TARGET` should be invoked like:
+```
+add_file_target(FILE <current source dir relative path> SCANNER_TYPE verilog)
+```
+
+For generated verilog, `ADD_FILE_TARGET` should be invoked like:
+```
+add_file_target(FILE <current source dir relative path> GENERATED)
+```
+Note that the input path must be current source dir relative, not absolute,
```
mithro:
```
The input path must be relative to current source directory (the normal options of absolute and relative to cmake build directory will not work).

As CMake cannot take dependencies during build time, generated files thus must have their dependencies manually supplied via using the [`APPEND_FILE_DEPENDENCY`](make/file_targets.cmake#L79) call.

Because we have generated Verilog files and they use relative include paths, source files need to be copied into the build output directory. This mandates the usage of the `ADD_FILE_TARGET`.
```

BTW As mentioned on chat -- Yosys and iverilog both have ways to set the include paths... I'm unsure if that makes a difference here?

cmake.md
litghost: ADD_FPGA_TARGET isn't really just about tests, it also the way designs are compiled.  With a small amount of refactoring, this outline should work for a frozen repository of arch.merge.xml, rr_graph files and IO pinmaps.

cmake.md
litghost: Done

cmake.md
litghost: Done

PR #236, opened 2018-08-13
Add CMake testing on Travis CI.

litghost:


.travis/script.sh
```diff
@@ -67,3 +67,17 @@ start_section "symbiflow.redir.2" "Running ${GREEN}make${NC} inside ${PURPLE}vpr
...
+
+start_section "symbiflow.configure_cmake" "Configuring CMake"
+mkdir build
+cd build
+cmake ..
+end_section "symbiflow.configure_cmake"
+
+$SPACER
+
+start_section "symbiflow.run_cmake_tests" "Running tests with CMake"
```
mithro: I think these might be better as separate sections?

.travis/script.sh
```diff
@@ -67,3 +67,17 @@ start_section "symbiflow.redir.2" "Running ${GREEN}make${NC} inside ${PURPLE}vpr
 end_section "symbiflow.redir.2"
 
 $SPACER
+
```
mithro: Should we remove the other make ones?

.travis/script.sh
litghost: Done

.travis/script.sh
litghost: Removed

PR #245, opened 2018-08-21
Add cmake commands and use genhlc to generate HLC as separate step

elmsfu:
Signed-off-by: Jeffrey Elms <elmsfu@freshred.net>

### DIFF
--- a/make/devices.cmake
+++ b/make/devices.cmake
@@ -796,6 +796,8 @@ function(ADD_FPGA_TARGET)
   get_target_property_required(VPR env VPR)
   get_target_property_required(GENHLC env GENHLC)
   get_target_property(VPR_TARGET env VPR_TARGET)
+  get_target_property(GENHLC_TARGET env GENHLC_TARGET)
+
   separate_arguments(
     VPR_BASE_ARGS_LIST UNIX_COMMAND "${VPR_BASE_ARGS}"
     )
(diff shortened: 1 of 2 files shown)


make/devices.cmake
```diff
@@ -771,6 +771,7 @@ function(ADD_FPGA_TARGET)
   endforeach()
 
   get_target_property_required(VPR env VPR)
+  get_target_property_required(GENHLC env GENHLC)
```
litghost: Need to get_target_property GENHLC_TARGET and add ${GENHLC_TARGET} to the add_custom_command target list.

elmsfu: I think I did this correctly. Any recommended way to check it?

elmsfu:
Marking as WIP as it depends on SymbiFlow/vtr-verilog-to-routing#1

make/devices.cmake
litghost: Not really.  If we ever break the parts of verilog -> bit flow up, it would be testable.  But right now GENHLC_CMD depends on an output that depends on VPR_TARGET, and VPR_TARGET == GENHLC_TARGET.  However this is now correct.

mithro:
LGTM when we land the `genhlc` command into conda -- I think it should be maybe something like `vpr-hlc` or something though?

elmsfu:
I think we can switch over, but our VPR fork needs to rename current `master` to something else and then pull in https://github.com/elmsfu/vtr-verilog-to-routing/tree/hlc_min_rebase as the  `master`. Any objections @litghost @mithro ?

I don't have permissions to do the branch work on our vpr fork.

mithro:
@elmsfu This is all done now right?

PR #246, opened 2018-08-21
utils: simplify and clean up formatting of dependency generation

elmsfu:
Rebasing some simplification I had started a while ago.

### DIFF
--- a/utils/deps_xml.py
+++ b/utils/deps_xml.py
@@ -5,10 +5,10 @@ Generate a Makefile .d fragment for the XML includes.
 
 import argparse
 import os
-import re
 import sys
 
 from io import StringIO
+import xml.etree.ElementTree as ET
 
 from lib.deps import add_dependency
 from lib.deps import write_deps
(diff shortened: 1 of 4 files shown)


dev_setup.sh
```diff
@@ -0,0 +1,7 @@
+export SF_ROOT=$(realpath $(dirname $BASH_SOURCE))
+export ICEBOX=$SF_ROOT/third_party/icestorm/icebox
+export VPR=$SF_ROOT/../vtr-verilog-to-routing/vpr/vpr
+
+export PATH=$SF_ROOT/env/conda/bin:$PATH
+export PATH=$SF_ROOT/../arachne-pnr/bin:$PATH
+#add icestorm to conda for icepack or build from 3rd party?
```
litghost: conda probably.  File an issue on conda-packages?

elmsfu: Oops I didn't mean to add this file. But yeah I think that is something I should go back and check if it's still an issue.

utils/deps_xml.py
```diff
@@ -17,35 +17,33 @@
...
-      line = line.strip()
-      if 'xi:include' not in line:
-          continue
+def read_dependencies(inputfile):
+    inputpath = os.path.abspath(inputfile.name)
+    inputdir = os.path.dirname(inputpath)
+    tree = ET.parse(inputfile)
+    for el in tree.iter():
+        if str(el.tag).endswith('XInclude}include'):
+            yield os.path.abspath(os.path.join(inputdir, el.get('href')))
```
litghost: Why does this work?  xi:include in the tag visible in the XML?  If this works, can you explain what is going on?

elmsfu: The xi is expanded to the URI by the python xml package. so `xi:include` is expanded to `'{http://www.w3.org/2001/XInclude}include'`

dev_setup.sh
```diff
@@ -0,0 +1,7 @@
+export SF_ROOT=$(realpath $(dirname $BASH_SOURCE))
```
mithro: Is there a way to not need this shell script?

elmsfu: Yeah I added it by accident. Removed it now.

PR #249, opened 2018-08-29
Update README

elmsfu:
Simplify README and provide correct build instructions.


elmsfu:
I realize this conflicts some with #44, but I think this is a step in the right direction to simplify the README with correct build instructions now.

PR #250, opened 2018-08-29
WIP: make: Add XMLLINT targets

elmsfu:
add build targets to run xmllint on generate XML files


elmsfu:
@litghost Any feedback regarding cmake targets? maybe I should wrap them up to a target to lint all xml?

I'll try to the net and routing xmls passing. I'm not a huge fan of how I had to shortcircuit xmllint if there is an error with head.

litghost:
I think we should encapsulate the XML verification in a cmake function.  Signature would be something like:

xml_lint(
  NAME <target name>
  FILE <file>
  SCHEMA <schema file>
  [LINT_OUTPUT <lint output file>]
  [ERROR_IF_LINT_PRESENT]
 )

Something like that.  Default could be to lint without generating an error, and the option ERROR_IF_LINT_PRESENT would generate an error if linter found a problem.

elmsfu:
Agree it should be function and that it should error.

@litghost what is the best way to make them run as part of the full build?

litghost:
Probably best to add a dummy all target to run the lint targets.  See the other all targets (https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/CMakeLists.txt#L46)

ice40/tests/blink/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
+set(BOARDS iceblink40-lp1k icestick)
+
+if (NOT DEFINED ENV{CI} OR NOT $ENV{CI})
+set(BOARDS ${BOARDS} hx8k-b-evn icevision tinyfpga-b2 tinyfpga-bx)
+endif (NOT DEFINED ENV{CI} OR NOT $ENV{CI})
```
litghost: I think list(APPEND BOARDS is more idiomatic, but this works too

make/devices.cmake
```diff
@@ -268,6 +279,45 @@ function(DEFINE_DEVICE_TYPE)
 
 endfunction()
 
+function(XML_LINT)
```
litghost: This doesn't belong in device.cmake.  Maybe an xml.cmake?

ice40/tests/blink/CMakeLists.txt
elmsfu: Done

make/devices.cmake
elmsfu: Done

elmsfu:
@litghost, LGTY?

litghost:
Lgtm

PR #254, opened 2018-10-03
Implement more robust points to tracks algorithm.

litghost:
The previous point algorithm was more complicated than required, and didn't handle all cases.  The new algorithm is more robust, though it may be less optimum.

A follow on will add a function that converts x-lines and y-lines and points and outputs VPR tracks, along with connections.

In particular, the last doctest in the new function caused the previous algorithm to return a disjoint set of lines.

### DIFF
--- a/utils/lib/rr_graph/points.py
+++ b/utils/lib/rr_graph/points.py
@@ -12,10 +12,10 @@ import enum
 from collections import defaultdict
 from collections import namedtuple
 
-from lib.asserts import assert_eq
-from lib.asserts import assert_type
+from ..asserts import assert_eq
+from ..asserts import assert_type
 
-from lib.rr_graph import Position, P
+from ..rr_graph import Position, P
 
 
 _NamedPosition = namedtuple("NamedPosition", ["pos", "names"])
--- a/utils/lib/rr_graph/points.py
+++ b/utils/lib/rr_graph/points.py
@@ -12,10 +12,10 @@ import enum
 from collections import defaultdict
 from collections import namedtuple
 
-from ..asserts import assert_eq
-from ..asserts import assert_type
+from lib.asserts import assert_eq
+from lib.asserts import assert_type
 
-from . import Position, P
+from lib.rr_graph import Position, P
 
 
 _NamedPosition = namedtuple("NamedPosition", ["pos", "names"])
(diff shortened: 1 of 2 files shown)


utils/lib/rr_graph/points.py
```diff
@@ -744,6 +744,285 @@ def straight_closet(line1, line2):
...
+
+    def __repr__(self):
+        return 'Track(dim={},points={})'.format(repr(self.dim), repr(self.points))
+
+def decompose_points_into_tracks(points):
+    """ This function takes a bag of points and returns a set of x lines and
+        y lines that cover all points, and all lines touch each other.
+
+    This is the first step to forming VPR tracks from points.  VPR tracks have
+    limited length, whereas this function returns lines of infinite length.
```
mithro: Do you have something which converts the infinite length lines back to points?

utils/lib/rr_graph/points.py
```diff
@@ -12,10 +12,10 @@
 from collections import defaultdict
 from collections import namedtuple
 
-from ..asserts import assert_eq
-from ..asserts import assert_type
+from lib.asserts import assert_eq
```
mithro: Why did you change these imports?

utils/lib/rr_graph/points.py
litghost: I'll add something in another PR, but this function is only for generating the list of rows and columns required to cover the points.

utils/lib/rr_graph/points.py
litghost: Invoking doctest explodes without it.

mithro: Go into the symbiflow-arch-defs/utils directory and run `python -m lib.rr_graph.points`?

utils/lib/rr_graph/points.py
mithro: BTW Does this cover all the things in this diagram?

image

litghost: It covers the same test cases that the previous algorithm did.  However, it will generate different tracks.  For example, in the global wire case, it might put the span at the end rather than the middle.  Not sure it matters.

mithro: I think you do actually want to minimize total path length? (IE Put the spine in the middle) The previous version probably just got it right by accident though.....

litghost: Why exactly?  Does VPR have channel distance heuristics?

mithro: In some modes, yes.

mithro:
Otherwise, this change looks good to me! Feel free to merge.

utils/lib/rr_graph/points.py
litghost: Okay, if we can about minimizing distances, we need something more general like a minimum spanning tree algo.  I believe https://www.sciencedirect.com/science/article/pii/0166218X9400095U is a suitable algo.  I'll file an issue to see if https://www.sciencedirect.com/science/article/pii/0166218X9400095U is suitable and implement it.

PR #274, opened 2018-10-31
Enable building without Conda

nathanrossi:
These changes allow for the build to execute without the Conda environment (e.g. `USE_CONDA=FALSE`).

Specifically it fixes issues where targets are depending on binary executables which are on the `PATH` instead of at a specific path. Enables icestorm tools to be provided from the `PATH` instead of only via the submodule. As well as setting up paths to components via cmake `-D` variables.

The purpose of these changes is to enabling building in alternative build environments such as OpenEmbedded, a work-in-progress recipe for that can be found (https://github.com/nathanrossi/meta-hdl/blob/e80efeeef3c1e34d7bc0d62af27e24a25665b421/recipes-fpga/symbiflow/symbiflow-arch-defs.inc).

Note: these changes do not cover all the issues with building outside of a Conda environment (e.g. netlistsvg).


nathanrossi:
Just a note the CI build appears to have failed due to an unrelated issue. It looks like a regression in yosys? I tested locally and it appears yosys at an older rev `d36d11936f6725b9fc0702e9406bc55533d8e273` works fine but at the current HEAD of `82965d60f53c817fea303abf8d6887ced8c8420e` it fails with:

```
Warning: Replacing memory \mem with list of registers. See
  <path>/usr/share/yosys/ice40/cells_sim.v:948,
  <path>/usr/share/yosys/ice40/cells_sim.v:937
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc

```

make/devices.cmake
```diff
@@ -358,7 +358,7 @@ function(DEFINE_DEVICE)
         ${symbiflow-arch-defs_SOURCE_DIR}/common/wire.eblif
         ${DEVICE_MERGED_FILE} ${DEVICE_MERGED_FILE_TARGET}
         ${QUIET_CMD} ${QUIET_CMD_TARGET}
-        ${VPR} ${VPR_TARGET}
```
litghost: You do actually want the executable in the depends list.  For example, if you recompile yosys or vpr, this target needs to be rebuilt.  So this was correct originally.

make/devices.cmake
```diff
@@ -1047,13 +1047,13 @@ function(ADD_FPGA_TARGET)
...
       COMMAND ${BIT_TO_BIN_CMD_FOR_TARGET_LIST}
-      DEPENDS ${BIT_TO_BIN} ${OUT_BITSTREAM}
+      DEPENDS ${OUT_BITSTREAM}
       )
 
     add_custom_target(${NAME}_bin DEPENDS ${OUT_BIN})
     add_output_to_fpga_target(${NAME} BIN ${OUT_LOCAL_REL}/${TOP}.bin)
 
-    get_target_property_required(PROG_TOOL ${BOARD} PROG_TOOL)
+    get_target_property(PROG_TOOL ${BOARD} PROG_TOOL)
```
litghost: This should be required.

make/env.cmake
```diff
@@ -159,11 +161,15 @@ function(ADD_CONDA_PACKAGE)
...
-    set(${binary_upper} ${NAME})
-    replace_with_env_if_set(${binary_upper})
-    set_target_properties(env PROPERTIES ${binary_upper} ${${binary_upper}})
-    set_target_properties(env PROPERTIES ${binary_upper}_TARGET "")
+  else()
+    foreach(OUTPUT ${ADD_CONDA_PACKAGE_PROVIDES})
+      string(TOUPPER ${OUTPUT} binary_upper)
+      set(${binary_upper} ${OUTPUT})
+      replace_with_env_if_set(${binary_upper})
+      set_target_properties(env PROPERTIES
```
litghost: If you want this to support finding programs on the path, than if replace_with_env_if_set fails to find the executable, should use find_program?

make/devices.cmake
nathanrossi: That makes sense now, I had not understood exactly why the dependency on the executable was there (if the target was handling the build/rebuild). Should the `VPR` like variables simply point at the full path?

make/devices.cmake
nathanrossi: I hit issues with some PROG_TOOLS being defined as a *_TARGET. When those targets expanded to empty string, it would have PROG_TOOL expand to empty string (and error).

Specifically I think it was just two definitions that were causing the issue:

https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/ice40/boards.cmake#L51
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/ice40/boards.cmake#L69

Should they be changed to using their `${TINYFPGAB}` program name instead? Except if changing the `TINYFPGAB` to a full path that might be annoying?

make/env.cmake
nathanrossi: So if replace_with_env_if_set does not clear the variable if it found no environment to replace it with. Which resulted in the variable just being the binary name to execute.

I am not very familiar with CMake, but I could not find a function that would provide the full path to a program that worked like find_program. Do you know of a way to get the full path `find_program()` found the program at? As doing such would make it easy to solve the use of these property values as dependencies.

make/devices.cmake
litghost: Yes, the executable variables (e.g. ${VPR}, ${YOSYS}) should be full paths generally speaking.  There is an exception, which is that if the executable is built by the CMake enviroment, the variable should be the target name from add_executable.

make/devices.cmake
litghost: I guess you are right, they should be targets, so not required.

make/env.cmake
nathanrossi: I had been testing with find_program after the set, which in turn was skipping the searching and just leaving the variable to the value of the set.

make/env.cmake
```diff
@@ -91,8 +91,12 @@ function(SETUP_ENV)
     set_target_properties(env PROPERTIES USE_CONDA FALSE)
     foreach(binary ${MAYBE_CONDA_BINARIES})
       string(TOUPPER ${binary} binary_upper)
-      set(${binary_upper} ${binary})
+      find_program(${binary_upper} ${binary})
       replace_with_env_if_set(${binary_upper})
```
litghost: We should probably make this explicit:
if env var -> use env
else use find_program
if not found -> error

nathanrossi: I've pushed an updated commit that makes this change.

PR #316, opened 2019-01-07
Improve the Verilog to XML conversion process

acomodi:
This PR relates to the V2X tools to generate the XML models and pb_types necessary to have the architecture XML definition of the design.

The tasks to be done are the following:

- Generate direct pins to pin
- Better detection of clocks and combination output.
- Adding tests for the output (based on the VtR timing tutorial).
- Use the XML format change in SymbiFlow#183 to allow multiple instances of a cell type (done).


utils/deps_verilog.py
```diff
@@ -25,7 +25,27 @@
...
+    >>> list(read_dependencies(StringIO('''
+    ... `include "a.h"
+    ... `include "b.h" /* Cmt
+    ... `include "c.h"
+    ...    */
+    ... `include "d.h" // Cmt
+    ... // `include "e.h"
+    ...   `include "f.h"
+    ... ''')))
+    ['a.h', 'b.h', 'd.h', 'f.h']
```
mithro: Probably worth having some paths in here?

utils/vlog/tests/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+include(v2x_tests.cmake)
+
+add_subdirectory(dsp_combinational)
+add_subdirectory(dsp_inout_registered)
+add_subdirectory(dsp_inout_registered_dualclk)
```
mithro: These tests below should be eventually uncommented.

utils/vlog/tests/dsp_in_registered/golden.pb_type.xml
```diff
@@ -0,0 +1,22 @@
+<pb_type name="mem_sp" blif_model=".subckt single_port_ram_mixed" num_pb="1">
```
mithro: FYI This is obviously for a single port ram model.

utils/vlog/tests/dsp_inout_registered_dualclk/dsp_inout_registered_dualclk.sim.v
```diff
@@ -0,0 +1,34 @@
+`include "../dsp_combinational/dsp_combinational.sim.v"
+
+/* DSP Block with register on both the inputs and the output */
```
mithro: Which use different clocks...

utils/vlog/tests/dsp_out_registered/golden.pb_type.xml
```diff
@@ -0,0 +1,22 @@
+<pb_type name="mem_sp" blif_model=".subckt single_port_ram_mixed" num_pb="1">
```
mithro: Again a RAM example :-)

utils/vlog/tests/test.mk
```diff
@@ -0,0 +1,32 @@
+#	xsltproc --xinclude ../../../../common/xml/xmlsort.xsl $< > $@.tmp
```
mithro: This file should be removed.

utils/deps_verilog.py
acomodi: As example, right?

mithro: Yeah.

acomodi:
I have a question related to the <interconnect> tag.
Should it be in the form of 
`<interconnect><direct><port attrs/><port attrs/></direct>...</interconnect>`
or 
`<interconnect><direct attrs></interconnect>`?

utils/vlog/tests/test.mk
acomodi: We have to remove the file from the compilation process or just the comment?

mithro: `test.mk` should be removed as we use cmake now.

kgugala:
This may be useful for timing integration @acomodi @mithro can we make this PR merged into master?

kgugala:
I mean finishing what needs to be finished, resolve conflicts and then merge

litghost:
> This may be useful for timing integration @acomodi @mithro can we make this PR merged into master?

Conflicts need to be resolve, and CI needs to be green.

mithro:
@kgugala - Yes, lets try and get this merged.

.gitignore
```diff
@@ -8,4 +8,8 @@ __pycache__/
 
 # CMake output
 build
-*.orig
```
litghost: Why remove *.orig?

common/xml/xmlsort.xsl
```diff
@@ -12,10 +12,12 @@
     <xsl:copy>
       <!-- Sort the attributes by name and remove the xml:base attribute -->
       <xsl:for-each select="@*[name()!='xml:base']">
-	<xsl:sort select="name( . )"/>
+	<xsl:sort select="name( . )" order="ascending"/>
```
litghost: Why change this?

make/tools.cmake
```diff
@@ -0,0 +1,43 @@
...
+    ""
+    "${oneValueArgs}"
+    ""
+    ${ARGN}
+    )
+
+  set(DIFF_FILE_A ${DIFF_GOLDEN})
+  set(DIFF_FILE_B ${DIFF_ACTUAL})
+
+  get_file_target(DIFF_FILE_A_TARGET ${DIFF_FILE_A})
```
litghost: Use append_file_dependency over directly depending on get_file_target.  It solves transitive dependencies.

make/tools.cmake
```diff
@@ -0,0 +1,43 @@
...
+    OUTPUT ${DIFF_OUTPUT}
+    DEPENDS
+      ${DIFF_FILE_A_LOCATION}
+      ${DIFF_FILE_A_TARGET}
+      ${DIFF_FILE_B_LOCATION}
+      ${DIFF_FILE_B_TARGET}
+    COMMAND
+      diff -u ${DIFF_FILE_A_LOCATION} ${DIFF_FILE_B_LOCATION} > ${DIFF_OUTPUT} || true
+    COMMAND
+      diff -u ${DIFF_FILE_A_LOCATION} ${DIFF_FILE_B_LOCATION}
```
litghost: Should probably have a "WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}"

utils/vlog/tests/v2x_tests.cmake
```diff
@@ -0,0 +1,112 @@
...
+    "${multiValueArgs}"
+    ${ARGN}
+  )
+
+  set(V2X_TEST_BOTH_SRC ${V2X_TEST_BOTH_NAME}.sim.v)
+  add_file_target(FILE ${V2X_TEST_BOTH_SRC} SCANNER_TYPE verilog)
+  v2x(NAME ${V2X_TEST_BOTH_NAME} SRCS ${V2X_TEST_BOTH_SRC} TOP_MODULE ${V2X_TEST_BOTH_TOP_MODULE})
+
+  v2x_generic_test(NAME ${V2X_TEST_BOTH_NAME} TOP_MODULE ${V2X_TEST_BOTH_TOP_MODULE} TYPE pb_type)
+  v2x_generic_test(NAME ${V2X_TEST_BOTH_NAME} TOP_MODULE ${V2X_TEST_BOTH_TOP_MODULE} TYPE model)
```
litghost: Should add a ${NAME}_update_golden target which updates the golden target.

utils/vlog/tests/v2x_tests.cmake
```diff
@@ -0,0 +1,112 @@
...
+  add_custom_target(${V2X_TEST_PB_TYPE_NAME}_diff ALL DEPENDS ${V2X_TEST_PB_TYPE_NAME}_pb_type_diff)
+endfunction(V2X_TEST_PB_TYPE)
+
+function(V2X_TEST_BOTH)
+  # ~~~
+  # V2X_TEST_BOTH(
+  #   NAME name
+  #   TOP_MODULE name
+  #   )
+  # ~~~
```
litghost: Please expand documentation

utils/vlog/tests/v2x_tests.cmake
```diff
@@ -0,0 +1,112 @@
...
+  add_file_target(FILE ${V2X_GENERIC_TEST_GOLDEN_XML} SCANNER_TYPE xml)
+  xml_sort(NAME ${V2X_GENERIC_TEST_NAME}_${V2X_GENERIC_TEST_TYPE}_golden_sort FILE ${V2X_GENERIC_TEST_GOLDEN_XML} OUTPUT ${V2X_GENERIC_TEST_NAME}.${V2X_GENERIC_TEST_TYPE}.golden.xml)
+
+  set(V2X_GENERIC_TEST_ACTUAL_XML ${V2X_GENERIC_TEST_NAME}.${V2X_GENERIC_TEST_TYPE}.xml)
+  xml_sort(NAME ${V2X_GENERIC_TEST_NAME}_${V2X_GENERIC_TEST_TYPE}_actual_sort FILE ${V2X_GENERIC_TEST_ACTUAL_XML} OUTPUT ${V2X_GENERIC_TEST_NAME}.${V2X_GENERIC_TEST_TYPE}.actual.xml)
+
+  diff(NAME ${V2X_GENERIC_TEST_NAME}_${V2X_GENERIC_TEST_TYPE}_diff GOLDEN ${V2X_GENERIC_TEST_NAME}.${V2X_GENERIC_TEST_TYPE}.golden.xml ACTUAL ${V2X_GENERIC_TEST_NAME}.${V2X_GENERIC_TEST_TYPE}.actual.xml)
+
+  # add_dependencies(${NAME}_diff ${V2X_GENERIC_TEST_NAME}_${V2X_GENERIC_TEST_TYPE}_diff)
+  #add_dependencies(all_v2x_tests ${V2X_GENERIC_TEST_NAME}_${V2X_GENERIC_TEST_TYPE}_diff)
```
litghost: Should add this target

utils/vlog/vlog_to_pbtype.py
```diff
@@ -114,6 +114,7 @@
 def mod_pb_name(mod):
     """Convert a Verilog module to a pb_type name in the format documented here:
     https://github.com/SymbiFlow/symbiflow-arch-defs/#names"""
+    return mod.name
```
litghost: What's going on here?

utils/vlog/yosys/json.py
```diff
@@ -122,6 +162,38 @@ def cell_conns(self, cell, direction="input"):
                         conns.append(("{}[{}]".format(port, i), condata[i]))
         return conns
 
+    def cell_conn_list(self, cell, port):
+        """The connections of a cell in a given port as a list
+
+        Returns a list of connections:
+        -------
+        net : list
```
litghost: Please follow commenting style from https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings

mithro: I'm pretty sure we have be using the [numpy style for docstring comments](https://numpydoc.readthedocs.io/en/latest/format.html)?

litghost: > I'm pretty sure we have be using the [numpy style for docstring comments](https://numpydoc.readthedocs.io/en/latest/format.html)?

I'm actually happy to see either.   I think picking a convention (and writing it down) is a good plan.

FYI looks like https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html supports both conventions.  They list the following note:

```
NumPy style tends to require more vertical space, whereas Google style tends to use more horizontal space. Google style tends to be easier to read for short and simple docstrings, whereas NumPy style tends be easier to read for long and in-depth docstrings.

The Khan Academy recommends using Google style.
```

I will say that prjxray generally is missing a significant number of docstrings, so it cannot be a guide.  symbiflow-arch-defs currently has a mix of styles, but given that we haven't been enforcing docstring style, it is unclear if either convention is actually being used.  I suggest we consider adding a docstring lint pass.

utils/vlog/vlog_to_pbtype.py
acomodi: this most probably is making CI fail as the pb_type name will be wrongly translated, I will remove this `return` and check if that's the case. Probably it's a leftover from the WIP

make/tools.cmake
acomodi: You mean instead of getting the location of the file?

litghost: DIFF_OUTPUT is not absolute.  You still need to get the locations of DIFF_A and DIFF_B

utils/vlog/tests/v2x_tests.cmake
litghost: Documentation please

utils/vlog/tests/v2x_tests.cmake
litghost: Please add this

utils/vlog/tests/v2x_tests.cmake
litghost: Still need a _update_golden target

make/xml.cmake
```diff
@@ -37,3 +37,54 @@ function(XML_LINT)
...
+    COMMAND
+      ${XSLTPROC}
+      --nomkdir
+      --nonet
+      --xinclude
+      --output ${CMAKE_CURRENT_BINARY_DIR}/${XML_SORT_OUTPUT}
+      ${XML_SORT_XSL}
+      ${XML_SORT_INPUT_LOCATION}
+  )
+  add_file_target(FILE ${XML_SORT_OUTPUT} GENERATED)
```
kgugala: We should set INCLUDE_FILES property for the XML_SORT_OUTPUT target

add_file_target with GENERATED flag will not add this automatically, so we need to handle this here

utils/vlog/tests/v2x_tests.cmake
acomodi: Could you clarify on this point? These tests, as far as I understood, are made to check regressions and their purpose is to check the XML produced by `v2x` against a manually written golden `pb_type/model`. I do not think that there should be a target that updates the corresponding golden of a test.

utils/vlog/tests/v2x_tests.cmake
acomodi: Done

utils/vlog/tests/v2x_tests.cmake
acomodi: Done

utils/vlog/tests/v2x_tests.cmake
litghost: The target will not be an all target.  The idea being if we intentionally change v2x that affects the output , someone can run the specific update_golden target to update the test output.

utils/vlog/tests/CMakeLists.txt
litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/599

utils/vlog/tests/v2x_tests.cmake
acomodi: So you mean that, we update the golden automatically only if we make sure that v2x provides a correct output right?
It would be something like `make dsp_combinational_update_pb_type_golden` which will indeed update the golden `pb_type` of the dsp_combinational test, right?

litghost: Yes.  Then the PR that updates V2X would include both a code change and diff on the output, which can then be reviewed together.

acomodi: Ok, got it, thanks. I'll add the update_golden target than, and for now I'll leave it blank as it will be filled in the following PRs

acomodi: Done

make/xml.cmake
acomodi: I do think that in this case it is fine to leave it as it is as there shouldn't be files to be included in the `XML_SORT_OUTPUT` dependencies.

utils/deps_verilog.py
acomodi: Actually, I think that the comments that are already present are self-explanatory enough

make/xml.cmake
kgugala: makes sense

PR #333, opened 2019-01-18
re-work xc7 directory

acomodi:
This PR is to add support for zybo for what regards the blinky. It is very hacky and is an early WIP to get a blinky.

The idea is also to re-work the cmake structure so to easily select at compile time which part and board has to be used for future additions.
Any suggestions on how to proceed in the most correct way?


xc7/boards.cmake
```diff
@@ -1,10 +1,19 @@
 get_target_property_required(OPENOCD env OPENOCD)
 get_target_property_required(OPENOCD_TARGET env OPENOCD_TARGET)
 
+#define_board(
```
litghost: You don't need to comment out other boards, especially if they use different devices.

xc7/devices/CMakeLists.txt
```diff
@@ -18,9 +18,9 @@ add_conda_pip(
 list(APPEND DEVICE_RR_PATCH_DEPS intervaltree textx)
 
 define_device(
-  DEVICE xc7a50t
-  ARCH artix7
-  DEVICE_TYPE xc7a50t-roi-virt
+  DEVICE xc7z010
+  ARCH zynq7
+  DEVICE_TYPE xc7z010-roi-virt
   PACKAGES test
```
litghost: Rather than renaming these, move this to a function that takes the device and arch name

xc7/tests/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
...
 add_subdirectory(counter)
-add_subdirectory(ram_test)
-add_subdirectory(carry)
-add_subdirectory(simple_ff)
-add_subdirectory(dram)
-add_subdirectory(bram)
+#add_subdirectory(ram_test)
+#add_subdirectory(carry)
+#add_subdirectory(simple_ff)
+#add_subdirectory(dram)
```
litghost: Why comment these out?  If you don't want to build them, just don't run their targets?

xc7/tests/counter/CMakeLists.txt
```diff
@@ -1,6 +1,13 @@
+#add_fpga_target(
+#  NAME counter
+#  BOARD basys3
+#  INPUT_IO_FILE basys3.pcf
+#  SOURCES counter.v
+#  )
+
 add_fpga_target(
```
litghost: Probably worth creating a wrapper function "add_fpga_targets" that thats BOARD/INPUT tuples and calls add_fgpa_target

xc7/utils/prjxray_arch_import.py
```diff
@@ -145,7 +145,7 @@ def main():
     g = db.grid()
     x_min, x_max, y_min, y_max = g.dims()
 
-    name = 'xc7a50t-test'
+    name = 'xc7z010-test'
```
litghost: This should be an argument

xc7/zynq7.cmake
```diff
@@ -0,0 +1,78 @@
...
+  )
+
+set(ROI_PART xc7z010clg400-1)
+set(ROI_DIR ${PRJXRAY_DB_DIR}/zynq7/harness/zybo/swbut)
+
+add_subdirectory(tiles)
+add_subdirectory(blocks)
+
+
+define_arch(
```
litghost: Given that most of this is the same, should make a function to abstract the stuff that is the "same".  For example, the YOSYS_SCRIPT line is going to change for BRAM support, and both artix7 and zync7 should likely use the same YOSYS_SCRIPT line.

litghost:
> The idea is also to re-work the cmake structure so to easily select at compile time which part and board has to be used for future additions.
> Any suggestions on how to proceed in the most correct way?

add_fpga_target emits a target for one board/pindef.  Just emit a add_fpga_target per board/pindef as needed, either a different target name.

For example counter for basys3 should probably be
counter_basys3 therefore `make counter_basys3_bin` would make the counter test for basys3, etc.

acomodi:
Perfect, thank you for the tips and comments. I'll proceed  following your suggestions

litghost:
Which zybo part are you planning to target?  There are apparently 3 boards with that name.

acomodi:
The one with the xc7z010clg400c-1

litghost:
> The one with the xc7z010clg400c-1

There are two zybo's with that part :/  Do you have the VGA or the HDMI one?
https://reference.digilentinc.com/reference/programmable-logic/zybo-z7/migration-guide

acomodi:
> There are two zybo's with that part :/ Do you have the VGA or the HDMI one?

It is the HDMI one, 7010 version

xc7/make/arch_define.cmake
```diff
@@ -0,0 +1,77 @@
+function(ADD_ARCH_DEFINE)
+  set(options)
```
litghost: Maybe ADD_XC7_ARCH_DEFINE?

xc7/boards.cmake
```diff
@@ -1,19 +1,18 @@
...
+  PROG_CMD "${OPENOCD} -f ${PRJXRAY_DIR}/utils/openocd/board-digilent-basys3.cfg -c \\\"init $<SEMICOLON> pld load 0 \${OUT_BIN} $<SEMICOLON> exit\\\""
+)
 
 define_board(
   BOARD zybo
   DEVICE xc7z010
   PACKAGE test
   PROG_TOOL ${OPENOCD_TARGET}
-  #  PROG_CMD "${OPENOCD} -f ${PRJXRAY_DIR}/utils/openocd/board-digilent-basys3.cfg -c \\\"init $<SEMICOLON> pld load 0 \${OUT_BIN} $<SEMICOLON> exit\\\""
+  PROG_CMD "${OPENOCD} -f ${PRJXRAY_DIR}/utils/openocd/board-digilent-basys3.cfg -c \\\"init $<SEMICOLON> pld load 0 \${OUT_BIN} $<SEMICOLON> exit\\\""
```
litghost: Does this config work for the zybo?  If so, maybe rename it in PRJXRAY to board-digilent.cfg?

xc7/make/arch_define.cmake
```diff
@@ -0,0 +1,77 @@
...
+  )
+
+  set(ARCH ${ADD_ARCH_DEFINE_ARCH})
+  set(ROI_PART ${ADD_ARCH_DEFINE_ROI_PART})
+  set(ROI_DIR ${ADD_ARCH_DEFINE_ROI_DIR})
+  set(YOSYS_SCRIPT ${ADD_ARCH_DEFINE_YOSYS_SCRIPT})
+
+  define_arch(
+	  ARCH ${ARCH}
+    # -flatten is used to ensure that the output eblif has only one module.
```
litghost: This comment belongs with the YOSYS_SCRIPT definition.

xc7/archs/artix7.cmake
```diff
@@ -0,0 +1,13 @@
+set(ROI_PART xc7a35tcpg236-1)
```
litghost: Move these directly into the add_arch_define to avoid variable name collisions?

litghost: Same with the zync version

xc7/boards.cmake
acomodi: I still need to verify if it works, but I doubt it

xc7/make/arch_define.cmake
acomodi: ok, makes sense

xc7/make/arch_define.cmake
acomodi: Ok, I will move it to where YOSYS_SCRIPT is defined

xc7/archs/artix7.cmake
acomodi: Ok

litghost:
xc7/archs/artix7/blocks/* can be deleted, they are not used.

litghost:
The tiles between zync and artix are almost 100% the same.  Do not duplicate them.  It is our belief that 7-series parts share tile definitions that are common, plus some part specific ones.

xc7/archs/zynq7/tiles/unused.clblm_l/clblm_l.pb_type.xml
```diff
@@ -0,0 +1 @@
+<pb_type></pb_type>
```
litghost: Delete this directory?  Added by mistake?

xc7/archs/zynq7/devices/xc7z010-roi-virt/dummy/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+function(GENERATE_DUMMY)
+  set(DUMMY_FILES
+    CLBLL_R.xml CLBLM_L.xml CLBLM_R.xml
+    INT_L.xml INT_R.xml HCLK_L.xml HCLK_R.xml)
+
```
litghost: Why are you generating dummies for these?

litghost:
devices/xc7a50t-roi-virt/dummy/* can be deleted.

xc7/archs/zynq7/devices/xc7z010-roi-virt/dummy/CMakeLists.txt
acomodi: Actually I had a question about the purpose about this dummy folder

litghost: It's no longer used, safe to delete

litghost:
> I have modified quite a few things. In my opinion it is cleaner if each architecture has its own tiles, blocks, and devices subdirectories this because it is easier to handle different architecture specific components. What do you think?

Given that tiles are generally speaking auto-generated, I think there is some commonization that can be done at the tile level.  I've been thinking for a while that the tile directory structure should be flattened, and then the tile/CMakeLists.txt would just have a bunch of import tile calls.

"Blocks" are not being used in xc7, so that can be delete.
Device subdirectories should be specific.

xc7/archs/artix7/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+set(ROI_DIR ${PRJXRAY_DB_DIR}/${ARCH}/harness/basys3/swbut)
+
+project_xray_prepare_database(
+  PART ${ARCH}
+)
+
+add_subdirectory(tiles)
+add_subdirectory(blocks)
+
+add_xc7_arch_define(
```
litghost: How does this function not have an ARCH ${ARCH} argument?
Should project_xray_prepare_database just move within add_xc7_arch_define?

acomodi: The ${ARCH} variable is set in the specific archs folder (e.g. artix7 or zynq7) being them in different scopes the ${ARCH} variable should be consistent with each different architecture. I found it easier this way because that is a variable that is used consistently within all the various subdirectories

litghost: That's exactly what you shouldn't do.  Attach the variable to a target name if you must, but do not rely on scoped variables like that.  It is hard to debug and prevents refactoring.

acomodi: Ok, good to now, I have re-added the ARCH argument.

acomodi:
Regarding the PYTHONPATH of the last commit, there is a better way to solve this issue I think. There should be the textx.pip in the build directory, but for some reason `fasm.py` cannot find it.

xc7/make/arch_define.cmake
```diff
@@ -42,7 +42,7 @@ function(ADD_XC7_ARCH_DEFINE)
     BIN_EXTENSION bit
     FASM_TO_BIT ${PRJXRAY_DIR}/utils/fasm2frames.py
     FASM_TO_BIT_CMD "${CMAKE_COMMAND} -E env \
-    PYTHONPATH=${PRJXRAY_DIR}:${PRJXRAY_DIR}/third_party/fasm \
+    PYTHONPATH=${symbiflow-arch-defs_BINARY_DIR}/env/conda/lib/python3.7/site-packages:${PRJXRAY_DIR}:${PRJXRAY_DIR}/third_party/fasm \
```
litghost: This is wrong and indicates your PYTHON3 variable is incorrect.

acomodi: Yes, I thought the same, the problem is that I cannot find a way let it correctly compile.

litghost: Do you happen to have an env var PYTHON3?  The symbiflow CMake allows executable overrides via an env var of the name.

soc_tests/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
-add_subdirectory(murax)
```
litghost: Don't comment these out

xc7/archs/zynq7/devices/xc7z010-roi-virt/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+project_xray_arch(
```
litghost: Put this behind a function, and take the DEVICE and ROI_DIR argument rather using scoped variables

xc7/boards.cmake
litghost: Did this get resolved?

acomodi: No, but I would say that for now we could cut out the `PROG_TOOL` and `PROG_CMD` for now when taking into account the zybo. When there is something working we could think of this, I guess.

litghost: That's fine, remove both lines to just have the prog tool be disabled for now.

litghost: Also add a TODO comment and point to a new issue with symbiflow-arch-defs to get openocd working with zync.

soc_tests/CMakeLists.txt
acomodi: Fixed

xc7/archs/zynq7/devices/xc7z010-roi-virt/CMakeLists.txt
acomodi: Fixed

xc7/boards.cmake
acomodi: Done

acomodi:
I think for now it is good as it is. Probably it is not yet general enough to easily add other boards and devices but I believe it is a good starting point.

litghost:
This looks great!  Thanks.  What test programs have you run and on which platforms?

Ideally test buttons, counter, ram_test on both basys3 and zybo.  Minimum test those 3 only on basys3.  To test ram_test, set the USB serial port to 500k baud and turn off processing, and xxd should show 'L', 0x00, '\n', '\r' repeated over and over.

acomodi:
So, for what regards the zybo, no test is successfully building because of database issues.

For what regards the basys3 I have run the following:
- Buttons: successfully builds but there is an issue with the correlation between output leds and input switches which is wrong.
- Counter: it successfully builds but same issue as for the buttons. I have tested this one with the master and the old artix7 db and it works fine on the leftmost leds.
- Carry: successfully builds and does not present exceptional behaviours on hardware as for the buttons and counter.
- Ram test: successfully builds, but the output is quite strange, probably I am doing something wrong. `00022450: 4c00 0d0a 4c00 0d0a 4c00 0d0a 4c00 0d0a  L...L...L...L...`

I will run all the tests again with the current master and the current prjxray-db master and check the differences.

Those errors may be due to some kind of regression, but I still need to verify what is the cause

litghost:
> * Counter: it successfully builds but same issue as for the buttons. I have tested this one with the master and the old artix7 db and it works fine on the leftmost leds.



> So, for what regards the zybo, no test is successfully building because of database issues.
> 
> For what regards the basys3 I have run the following:
> 
> * Buttons: successfully builds but there is an issue with the correlation between output leds and input switches which is wrong.

An issue with buttons is usually an harness issue.  Try with the old harness?

> * Counter: it successfully builds but same issue as for the buttons. I have tested this one with the master and the old artix7 db and it works fine on the leftmost leds.

Left most LEDs are expected!  So that is successful.

> * Carry: successfully builds and does not present exceptional behaviours on hardware as for the buttons and counter.
> * Ram test: successfully builds, but the output is quite strange, probably I am doing something wrong. `00022450: 4c00 0d0a 4c00 0d0a 4c00 0d0a 4c00 0d0a  L...L...L...L...`

That output is correct actually.  4c00 0d0a is 'L', 0x00, '\n', '\r'.

acomodi:
> An issue with buttons is usually an harness issue. Try with the old harness?

I have built the test using the changes in this PR and the old database with the old harness and the issue persists.

 

> Left most LEDs are expected! So that is successful.

Actually when building with this PR changes the LEDs that are switched on are not on the left most part, but are spread through the array in random spots.

acomodi:
The only difference is in the fact that the counter and buttons tests makefiles are created using the `add_fpga_target_boards` in cmake instead of `add_fpga_target` as for the other tests which were successful. I don't know whether this could be the source of the issue, but I'll give it a look

litghost:
> The only difference is in the fact that the counter and buttons tests makefiles are created using the `add_fpga_target_boards` in cmake instead of `add_fpga_target` as for the other tests which were successful. I don't know whether this could be the source of the issue, but I'll give it a look

If ram_test is still us add_fpga_target and counter and buttons is using add_fpga_target_boards, definitely sounds like a regression.  However it narrows the problem source quiet a bit.

xc7/tests/buttons/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
+set(BOARDS basys3 zybo)
```
litghost: This doesn't fix the but with explicit IO files does it?

acomodi:
Found the bug. The .pcf file weren't added when considering using `add_fpga_target_boards`. I will test it now again and check if everything works fine with this fix

xc7/tests/buttons/CMakeLists.txt
acomodi: I have actually used the same paradigm as for the Ice40. The fix should be the `IMPLICIT_INPUT_IO_FILES`. Probably I should just remove the scoped variable

acomodi: Sure. So, As far as I got from debugging the should be a bug in the handling of the explicit `INPUT_IO_FILES`. I need to further check it, but for what I saw there is no use of explicit IO files in the repo. Maybe there should be further checks to guarantee that the `INPUT_IO_FILE` list is correctly parsed and passed to the actual `add_fpga_target` function. I will open the ticket right now

PR #352, opened 2019-02-04
XC7 tests fix

acomodi:
This PR is to fix some tests to make them available both for artix and zynq.


xc7/utils/prjxray_arch_import.py
```diff
@@ -147,6 +147,12 @@ def main():
     g = db.grid()
     x_min, x_max, y_min, y_max = g.dims()
 
+    # FIXME: There is an issue in the routing phase.
```
litghost: Link to a github issue

acomodi: Done

PR #354, opened 2019-02-04
Fix VPR_ARCH_ARGS

elmsfu:
VPR_ARCH_ARGS was being ignored in cmake define_arch()

ice40 tests should complete now.

Also added ability to specify conda package versions and builds. This was helpful in bisecting the failure and will be good for any release or milestone.

### DIFF
--- a/ice40/CMakeLists.txt
+++ b/ice40/CMakeLists.txt
@@ -12,6 +12,8 @@ define_arch(
   DEVICE_FULL_TEMPLATE \${DEVICE}-\${PACKAGE}
   VPR_ARCH_ARGS
     --clock_modeling route
+    --allow_unrelated_clustering off
+    --target_ext_pin_util 0.7
   RR_PATCH_TOOL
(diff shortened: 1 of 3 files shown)


make/env.cmake
```diff
@@ -156,7 +156,8 @@ function(ADD_CONDA_PACKAGE)
       DEPENDS conda ${CONDA_BIN}
       )
 
-    set(TARGET conda_${ADD_CONDA_PACKAGE_NAME})
+    # Allow specification of version and build, but strip it off for the package name
+    string(REGEX MATCH "^[^=]+" TARGET ${ADD_CONDA_PACKAGE_NAME})
```
litghost: This change seems unrelated?

elmsfu: It was an improvement during debugging. Do you want me to break out the one line change to a different PR?

litghost: Ya

PR #355, opened 2019-02-04
Add support for specifying version and build for conda packages

elmsfu:


litghost:
Please add documentation for what this does and how to use it

elmsfu:
Thanks @litghost Taking a closer look at the spec, I think an optional argument is more general and a better solution.

litghost:
You brought in some submodule changes, revert?

elmsfu:
> You brought in some submodule changes, revert?

nope. oops. will fix that.

PR #359, opened 2019-02-05
Convert yosys scripts to TCL.

litghost:
Per suggestion in #321


elmsfu:
This broke the build on Travis. Should we get it green and then turn it back to not allow merge if it fails on Travis?

litghost:
Was travis green prior to the PR?

elmsfu:
Yes. And you can confirm https://travis-ci.org/SymbiFlow/symbiflow-arch-defs/builds

Issue #385, opened 2019-02-19
BRAM test sometimes fails due to indeterminate cause


litghost:
I've got the BRAM test work with initial BRAM set to all 0's.  I've seen it sometimes fail with a signature that looks like no writes are happening.  I haven't been able to replicate this issue consistently, but it is something to look out for.

litghost:
This might be a timing issue?  Re-running PnR changes the result.  I'll try a couple regenerations to see probabilities.

litghost:
Another possibility has to do with cascading address interconnect.  There are some unknown bits that show's up when Vivado compiles the bram_test:
 - 26_170
 - 26_172

This appears to be related to cascade address lines, but I haven't been able to pinpoint what the bits are for.

litghost:
Ya, I've confirmed that using the cascade address lines appears to be part of the trouble.  In cases where VPR routes all through the normal address lines, the test passes.  In cases where VPR routes through the cascade address lines, the test fails.

I'll manually add 26_170 and 26_172 to the solution, and see if a failing test passes.

litghost:
Adding the "magic" bits 26_170 and 26_172 to the sourcing cascade address tile appears to fix the issue.  Solution is now to write a fuzzer that correctly documents those bits.

Issue #283, opened 2018-12-19
Add and test 7-series BRAM


litghost:
Currently BRAM's have bit-stream definitions, but the arch models don't use them.  First priority is to get BRAM's working in any mode.  Simplest mode to attempt is like 18k x 16-bit.

BRAM segbits: https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/segbits_bram_l.db
BRAM pb type: artix7/primitives/bram/ramb18e1.pb_type.xml

litghost:
Manual bram test appears to be working.  Next step is an automated R/W test and see about adding initial data support.

PR #386, opened 2019-02-19
Fix BRAM test.

litghost:
- Correct WE signals for TDP BRAM.
 - Add loop count to result report
 - Explicitly initialize UART Tx output state.

BRAM INIT contents is still not handled correctly.


litghost:
This begins to solve #283, barring solving #385.

litghost:
With 7ca1adf and a database that includes https://github.com/SymbiFlow/prjxray/pull/668, BRAM test appears to work.

litghost:
Fixes #385

Issue #382, opened 2019-02-15
The xc7/tests/counter example does not work from the current master branch (4b66516)


mkurc-ant:
The project builds but does not work on the actual hardware. All 4 LEDSs are off. Simulation of `top_synth.v` works fine. The last commit I've managed to successfully build and run is 79ab13f. I don't know if any other tests are affected. 

For all my tests I've set the `LOGIC_DELAY` parameter to 0

I've checked commits between 79ab13f and 4b66516 (master), here is the summary:

4b66516 (master):

The xc7/tests/counter builds but does not work. Got all 4 LEDs unlit.

50dca48:

Same as above.

ae9a635:

The build system fails with an error:

```
[ 66%] Generating pin_assignments.json
[ 66%] Built target file__home_mkurc_Repos_google-symbiflow-arch-defs_xc7_archs_artix7_channels.json
2019-02-15 17:02:28.175487 Reading channel data
Traceback (most recent call last):
  File "/home/mkurc/Repos/google-symbiflow-arch-defs/xc7/utils/prjxray_assign_tile_pin_direction.py", line 290, in <module>
    main()
  File "/home/mkurc/Repos/google-symbiflow-arch-defs/xc7/utils/prjxray_assign_tile_pin_direction.py", line 48, in main
    channels = json.load(f)
  File "/home/mkurc/Repos/google-symbiflow-arch-defs/build/env/conda/lib/python3.7/site-packages/simplejson/__init__.py", line 461, in load
    use_decimal=use_decimal, **kw)
  File "/home/mkurc/Repos/google-symbiflow-arch-defs/build/env/conda/lib/python3.7/site-packages/simplejson/__init__.py", line 518, in loads
    return _default_decoder.decode(s)
  File "/home/mkurc/Repos/google-symbiflow-arch-defs/build/env/conda/lib/python3.7/site-packages/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/home/mkurc/Repos/google-symbiflow-arch-defs/build/env/conda/lib/python3.7/site-packages/simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Invalid control character '\n' at: line 54604359 column 16 (char 945489370)
[...]

PR #392, opened 2019-02-21
[XC7] Add BRAM init

### LINKED ISSUE #283: Add and test 7-series BRAM

litghost:
Updates the ROI to included required fixtures, fixing #382.  This also completes initial BRAM support, fixing #283.


litghost:
Fixes #382

Issue #339, opened 2019-01-24
Merge the graph.py and graph2.py files in utils


mithro:
We have both;
 * https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/lib/rr_graph/graph.py
 * https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/lib/rr_graph/graph2.py

It would be good to merge them and add a decent test suite.

Fleker:
It looks like the files do a lot of similar things, but differently. For example, graph2.py defines a few enums like SwitchType as numbers, whereas SwitchType in graph.py uses strings. Is there a preference for this?

litghost:
In all cases prefer graph2 over graph.

elmsfu:
To give some background: graph is used for the ice40 portion and is tightly coupled with the XML. graph2 is for xc7 and avoids being tied with the XML format and is more optimised for larger FPGAs. I think the the optimizations (and separation from XML) are the reasons @litghost says prefer graph2.

It may make sense to have some unit tests at the higher level to allow quicker iteration. I'm actually adding some to for [ice40_import_layout_from_icebox.py](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/013388bf79c452659ded69283d5dc65e3f2d713d/ice40/utils/ice40_import_layout_from_icebox.py). So maybe some tests around [ice40_import_routing_from_icebox.py](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/013388bf79c452659ded69283d5dc65e3f2d713d/ice40/utils/ice40_import_routing_from_icebox.py#L90) is a good place to start.

I think this is a bigger task and could easily be broken down.

PR #399, opened 2019-02-25
Adds tests for graph.py

Fleker:
As part of #339, I started to write tests to verify existing behavior. During tests, I found some additional instructions to add to the README, and a fix to be made to the graph.py souce.

Let me know if the current changes look good, and if so I'll continue work on the issue.

### DIFF
--- a/utils/lib/rr_graph/tests/test_graph.py
+++ b/utils/lib/rr_graph/tests/test_graph.py
@@ -1,0 +1,772 @@
+#!/usr/bin/env python3
+# Run `python3 -m utils.lib.rr_graph.tests.test_graph`
+from .. import graph, P, Size
+from ..graph import (Pin, PinClass, PinClassDirection, Block, BlockGrid, BlockType, Segment,
+    Switch, SwitchType, RoutingGraph, RoutingGraphPrinter, RoutingNodeType)
+
+import lxml.etree as ET
+
+from ...asserts import assert_eq
+
+def test_parse_net():
+    # Fully specified
+    assert_eq(
+        ('a', 'b', [0]),
+        graph.parse_net('a.b[0]')
+    )
+    assert_eq(
+        ('c', 'd', [1]),
+        graph.parse_net('c.d[1]')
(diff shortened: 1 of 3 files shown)


utils/lib/rr_graph/tests/test_graph.py
```diff
@@ -0,0 +1,772 @@
+#!/usr/bin/env python3
```
litghost: Any reason not to use (unittest)[https://docs.python.org/3/library/unittest.html] or equivilent?

mithro:
Hi @Fleker,

Thank you for writing these tests! 

As Keith suggests we should use an existing testing framework. The unittest framework that is included with Python 3 is probably a good idea. `py.test` might be the other option.

Tim 'mithro' Ansell

utils/lib/rr_graph/tests/test_graph.py
Fleker: I don't have a particular reason. I can definitely reformat the file, I wanted to make sure I wasn't using an unexpected framework that wasn't wanted in the project.

mithro:
@Fleker, I have added a WIP title, feel free to remove once you have finished the reformat. 

Thanks once again!

Fleker:
I changed the tests to use Python3's unittest. A second PR will add this and more tests to the Travis build.

utils/lib/rr_graph/tests/test_graph.py
```diff
@@ -0,0 +1,754 @@
...
+        )
+        self.assertEqual(
+            P(x=0, y=0),
+            g.block_grid[P(0, 0)].position
+        )
+        self.assertEqual(
+            P(x=2, y=1),
+            g.block_grid[P(2, 1)].position
+        )
+        try:
```
mithro: `self.assertRaises`

Fleker: Updated.

mithro:
Thank you for your pull request!

Issue #372, opened 2019-02-11
VTR cannot pack slicem.carrychain


kgugala:
This can be reproduced by uncommenting the fifth counter in https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/tests/chain_packing/counter.v#L7
and using VPR from this PR https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/9

This will pack the fifth counter into SLICEM.CARRYCHAIN and VPR packer fails with:
```
Error 1: 
Type: Packing
File: vtr-verilog-to-routing/vpr/src/pack/cluster.cpp
Line: 2029
Message: Can not find any logic block that can implement molecule.
	Pattern BLK_TI-CLBLM_R.BLK_IG-SLICEM.CARRYCHAIN $auto$alumacc.cc:474:replace_alu$29.slice[0].carry
```
This happens only if a chain is packed into SLICEM

acomodi:
I have hacked this error by removing the CLBLM_[LR] definition from the `device_type` instantiation in `cmake` as in the example below:

```
add_xc7_device_define_type(
  ARCH artix7
  DEVICE xc7a50t
  ROI_DIR ${PRJXRAY_DB_DIR}/artix7/harness/basys3/swbut
  TILE_TYPES INT_R INT_L CLBLL_R CLBLL_L BRAM_L
)
```

This has the effect of not including `DRAM` blocks which are connected to the CLBLM_[LR] tiles.

litghost:
Ya, you completely removed CLBLM_[LR] tiles from the arch, which is not a solution.

acomodi:
Yes indeed, It Is just a temporary workaround, it should be properly solved

acomodi:
I have been studying this problem. Follows the steps to reproduce it:

- VPR version: https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/9
- Symbiflow-arch-defs: 79ab13f2b2ace5009b14fb11134f4e2a72fa9d63
- VPR options in [arch_define.cmake](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/4b6651608ab699a1a56bfc6a491c327e97ffd5db/xc7/make/arch_define.cmake#L27) to be added: `--clustering_pin_feasibility_filter off`
[...]

PR #402, opened 2019-02-25
d_dram.pb_type: update xml definition to solve slicem issue

acomodi:
This PR addresses issue https://github.com/SymbiFlow/symbiflow-arch-defs/issues/372. I have been dealing with the `slicem` issue and found a solution that involved the modification and addition of the `dual port` modes for the `d_dram.pb_type.xml` and `d_dram.128.pb_type.xml`. Even though `DLUT` should not support `dual port` mode, its addition let me successfully compile the `chain_packing` test.

The test was modified as follows:
- LOGIC_DELAY: set to 4 because of https://github.com/SymbiFlow/vtr-verilog-to-routing/issues/8;
- Addition of the 5th counter to include also the usage of SLICEMs.

VPR version was the following: https://github.com/antmicro/vtr-verilog-to-routing/commit/b99b9941697f37a36134711884bde4d211cbcba7

I believe that being there a discrepancy between the `d_dram` and the `a/b/c_dram` the modes cannot be selected accordingly, provoking the following error during the `cluster_routing`:
- `Differing modes for block.  Got LUTs previously and DRAMs for interconnect DO6`.

IMHO the problem is not 100% fixed because there should be a mechanism during clustering in VPR to select the `mode` that can satisfy all the molecules, and this is not yet happening in the correct way.

I have also checked whether the tests in the `xc7/tests/dram` directory were working on HW as well. 
A note about the dram tests: I have compared the one produced by master branch of `arch-defs` (https://github.com/SymbiFlow/symbiflow-arch-defs/commit/477f9aee350f75073c58b76a832ba2c766c44f45) and the ones produced with this PR. They work exactly the same on `HW` and they both present (IMO) an error in `HW` for what regards `2_32x1d` and `4_32x1s` tests.

### DIFF
--- a/xc7/primitives/slicem/Ndram/d_dram.pb_type.xml
+++ b/xc7/primitives/slicem/Ndram/d_dram.pb_type.xml
@@ -1,48 +1,128 @@
 <pb_type name="BLK_IG-D_DRAM" num_pb="1" xmlns:xi="http://www.w3.org/2001/XInclude">
   <clock  name="CLK" num_pins="1" />
   <input  name="A"   num_pins="6" />
+  <input  name="WA"  num_pins="6" />
   <input  name="WA7" num_pins="1" />
   <input  name="WA8" num_pins="1" />
   <input  name="DI1" num_pins="1" />
   <input  name="DI2" num_pins="1" />
   <input  name="WE"  num_pins="1" />
 
+  <output name="DO6"  num_pins="1" />
+  <output name="DO6_32"  num_pins="1" />
   <output name="SO6"  num_pins="1" />
   <output name="SO6_32"  num_pins="1" />
   <output name="O6"  num_pins="1" />
   <output name="O5"  num_pins="1" />
 
- <!-- TODO: Missing modes: SRL. -->
- <mode name="64_SINGLE_PORT">
-   <xi:include href="spram64.pb_type.xml" />
-   <interconnect>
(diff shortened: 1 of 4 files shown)


xc7/primitives/slicem/Ndram/d_dram.pb_type.xml
```diff
@@ -1,48 +1,128 @@
...
-     <direct name="O5" input="BLK_MM-SPRAM32.O5" output="BLK_IG-D_DRAM.O5" />
-     <direct name="SO6_32" input="BLK_MM-SPRAM32.O6" output="BLK_IG-D_DRAM.SO6_32" />
-   </interconnect>
- </mode>
-
- <metadata>
-  <meta name="fasm_prefix">DLUT</meta>
- </metadata>
+<!-- TODO: Missing modes: SRL. -->
+  <mode name="LUT">
```
litghost: You can't have the LUT mode in d_dram.  When any DRAM is used in a slice, the D element must be a RAM.

acomodi: Ok, this makes sense because, as far as I have seen, the DLUT inputs `DI` are governing all the other `LUTs`. I have tried to remove it, but the routing failed as I expected. I will look again into the `clusterer` to solve the issue

acomodi:
This can be closed as https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/29 solves the issue

PR #407, opened 2019-02-27
Remove tile level fasm_prefix

elmsfu:
It was previously ignored does not match features in db

This should merge once genfasm conda package has https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/12


litghost:
What tests were run?

elmsfu:
> What tests were run?

I compared a few generated fasm outputs and results were identical.

PR #426, opened 2019-03-05
ice40: Temp fix for build by running `abc` separately without -dress

elmsfu:
When the fix for libblifparse is merged and propagated to VPR this can be reverted

Signed-off-by: Elms <elms@freshred.net>

### DIFF
--- a/ice40/yosys/synth.tcl
+++ b/ice40/yosys/synth.tcl
@@ -1,6 +1,8 @@
 yosys -import
 
-synth_ice40 -nocarry
+synth_ice40 -nocarry -noabc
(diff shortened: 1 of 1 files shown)


ice40/yosys/synth.tcl
```diff
@@ -1,6 +1,8 @@
 yosys -import
 
-synth_ice40 -nocarry
+synth_ice40 -nocarry -noabc
+# TODO: revert when libblifparse fix propagates to vpr
```
litghost: Link to an issue so we can track the work?

elmsfu: Done #428

PR #435, opened 2019-03-06
WIP - Xilinx Series 7 split CLB tiles into two tiles types (SLICEL / SLICEM)

mkurc-ant:
This is a work in progress on splitting CLBs into individual SLICEs for the VPR.

The idea is to split each column of the FPGA tile grid so for each CLB column there will be two SLICE columns. There are 4 tile types in the grid. Each of them is split in two so in the end we have 8 tile types:

```
CLBLL_L -> CLBLL_L_SLICEL_X0Y0, CLBLL_L_SLICEL_X1Y0
CLBLL_R -> CLBLL_R_SLICEL_X0Y0, CLBLL_R_SLICEL_X1Y0
CLBLM_L -> CLBLM_L_SLICEM_X0Y0, CLBLM_L_SLICEL_X1Y0
CLBLM_R -> CLBLM_R_SLICEM_X0Y0, CLBLM_R_SLICEL_X1Y0
```

The split is implemented by introducing a kind of "overlay" to the prjxray database. There are new Python scripts added which do the split. These scripts take `tilegrid.json`, `tileconn.json` and relevant `tile_type_*.json` files, process them and saves them into a separate folder. All of the `prjxray_*.py` utility scripts were modified so they access the database through that overlay. So they operate on a splitted grid and new tile types.

The process of grid splitting is integrated into the build system. So far I've managed to generate a fasm file for the `xc7/tests/counter` project. The bitstream cannot be generated yet as new tile types are not recognized by the `fasm2frames.py` utility. It is still a WIP.


xc7/utils/db_overlay/db_overlay.py
```diff
@@ -0,0 +1,98 @@
...
+        self.overlay_root = overlay_root
+
+        # Load additional tile types
+        for f in os.listdir(self.overlay_root):
+            if f.endswith('.json') and f.startswith('tile_type_'):
+                tile_type = f[len('tile_type_'):-len('.json')].lower()
+
+                print("Overlay tile type: '%s'" % tile_type)
+
+                segbits = os.path.join(
```
litghost: There shouldn't be any segbits, pips or mask files in the overlay, so why is this here?

xc7/make/project_xray.cmake
```diff
@@ -212,15 +214,33 @@ function(PROJECT_XRAY_PREPARE_DATABASE)
   file(GLOB DEPS ${PRJXRAY_DB_DIR}/${PART}/*.json)
   file(GLOB DEPS2 ${PRJXRAY_DIR}/prjxray/*.py)
 
+  file(MAKE_DIRECTORY ${PRJXRAY_DB_OVERLAY_DIR}/${PART})
+
+  set(GRID_SPLITTER ${symbiflow-arch-defs_SOURCE_DIR}/xc7/utils/splitter/grid_splitter.py)
+
+  # FIXME: Move it somewhere
+  set(TILES_TO_SPLIT CLBLL_L CLBLL_R CLBLM_L CLBLM_R)
```
litghost: Can we make this config driven?  Then the database overlay knows exactly what files to expect?

xc7/make/project_xray.cmake
```diff
@@ -212,15 +214,33 @@ function(PROJECT_XRAY_PREPARE_DATABASE)
...
+
+  set(TILEGRID ${PRJXRAY_DB_OVERLAY_DIR}/${PART}/tilegrid.json)
+  set(TILECONN ${PRJXRAY_DB_OVERLAY_DIR}/${PART}/tileconn.json)
+  add_custom_command(
+  OUTPUT ${TILEGRID} ${TILECONN} # FIXME: The script will output not only these files
```
litghost: If the overlay is config driven, the output list can be generated from the config?

xc7/archs/artix7/tiles/CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
 add_subdirectory(bram_l)
 add_subdirectory(bram_r)
-add_subdirectory(clbll_l)
```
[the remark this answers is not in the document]
litghost: I suggested else where to make the tile split config driven.  Could use the existing directory structure + config to determine outputs

make/devices.cmake
```diff
@@ -565,7 +565,7 @@ endfunction()
 
 set(VPR_BASE_ARGS
     --min_route_chan_width_hint 100
-    --verbose_sweep on
+#    --verbose_sweep on
```
litghost: This was changes in master, can rebase and remove this

xc7/utils/splitter/conn_splitter.py
```diff
@@ -0,0 +1,505 @@
+#!/usr/bin/env python3
```
litghost: How was the new tileconn tested?  Have you considered an approach that wouldn't required writing a new tileconn.json?

litghost:
> The bitstream cannot be generated yet as new tile types are not recognized by the fasm2frames.py utility. It is still a WIP.

There should be no need to modify fasm2frames.py to accomplish this.  I will not accept a PR that requires a change in fasm2frames.py

mithro:
Should we end up with just a single `SLICEL` and `SLICEM` types? If the split means we still have multiple tile types then it is not going to achieve the outcome we are after.

xc7/archs/artix7/devices/xc7a50t-roi-virt/CMakeLists.txt
```diff
@@ -2,6 +2,6 @@ add_xc7_device_define_type(
   ARCH artix7
   DEVICE xc7a50t
   ROI_DIR ${PRJXRAY_DB_DIR}/artix7/harness/basys3/swbut
-  TILE_TYPES INT_R INT_L CLBLL_R CLBLL_L CLBLM_R CLBLM_L BRAM_L
+  TILE_TYPES INT_R INT_L CLBLL_L_SLICEL_X0Y0 CLBLL_L_SLICEL_X1Y0 CLBLL_R_SLICEL_X0Y0 CLBLL_R_SLICEL_X1Y0 CLBLM_L_SLICEM_X0Y0 CLBLM_L_SLICEL_X1Y0 CLBLM_R_SLICEM_X0Y0 CLBLM_R_SLICEL_X1Y0 BRAM_L
```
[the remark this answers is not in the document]
mithro: I was expecting something like;
```patch
-TILE_TYPES INT_R INT_L CLBLL_R CLBLL_L CLBLM_R CLBLM_L BRAM_L
+TILE_TYPES INT_R INT_L SLICEL SLICEM BRAM_L
```

xc7/archs/artix7/tiles/CMakeLists.txt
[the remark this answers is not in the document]
mkurc-ant: Yes, I'll think about that

xc7/utils/db_overlay/db_overlay.py
mkurc-ant: I just copied reading of all relevant tile data. It is not necessary so I'll remove it.

xc7/make/project_xray.cmake
mkurc-ant: Yes, That's why it is commented as "#FIXME"

xc7/make/project_xray.cmake
mkurc-ant: True.

make/devices.cmake
mkurc-ant: Yeah, my mistake. I did the rebase.

xc7/utils/splitter/conn_splitter.py
mkurc-ant: For "testing" I've written an utility that draws the FPGA grid along with connection using data in `tilegrid.json` and `tileconn.json`. I can visually examine if connection on the splitted grid are ok.

It is still WIP. For now there is one connection case which has not been converted and it is when carry chain cross clock region boundary.

mkurc-ant:
@litghost I don't plan to modify fasm2frames.py. I'm going to change fasm prefixes instead.

mkurc-ant:
@mithro Making only tile types `SLICEL` and `SLICEM` requires more effort. Even if now there are two identical SLICEs in CLB they have different wire names that come out of the CLB. I would need to rename all the wires.

litghost:
> @mithro Making only tile types `SLICEL` and `SLICEM` requires more effort. Even if now there are two identical SLICEs in CLB they have different wire names that come out of the CLB. I would need to rename all the wires.

I don't think this is a true statement, but instead a weakness of this particular approach.

xc7/utils/db_overlay/db_overlay.py
mkurc-ant: Ok, removed that.

litghost:
This PR it significantly more complicated than required.  There is a significantly simpler approach if you abandon the "overlay" approach as written.  Please pause and rethink your approach, this is not a fruitful direction.

mkurc-ant:
Right. I know that this is not the best option, however there are issues preventing me from taking a different approach. One of them is that there are many python scripts that do prjxray database import. And all those scripts make use of same information from the db (eg the tile grid) which has to be stored somewhere.

Tomorrow (my time) I'll prepare a document describing what I want to change and how I am planning to do it. I will share it with you.

mkurc-ant:
So after recent discussion we decided that the approach in this PR should not be progressed further. I'am closing this PR.

Issue #353, opened 2019-02-04
Error in VPR during routing step for Zynq


acomodi:
VPR exits with an error while routing when a zynq part is selected and while running the `counter` test in the `xc7` directory. The error is the following

```
Warning 7064: in check_rr_node: RR node: 230581 type: OPIN location: (117,1) pin: 88 pin_name: BLK_TI-CLBLM_R.CLBLM_M_COUT[0] with capacity 1 has no out-going edges.
Warning 7065: in check_rr_node: RR node: 240739 type: CHANX location: (89,51) <-> (91,51) track: 210 len: 2 longline: 0 seg_type: dummy dir: BI_DIR with capacity 1 has no out-going edges.
Warning 7066: in check_rr_node: RR node: 240804 type: CHANX location: (89,51) <-> (91,51) track: 199 len: 2 longline: 0 seg_type: dummy dir: BI_DIR with capacity 1 has no out-going edges.
Warning 7067: in check_rr_node: RR node: 240809 type: CHANX location: (89,51) <-> (91,51) track: 198 len: 2 longline: 0 seg_type: dummy dir: BI_DIR with capacity 1 has no out-going edges.
Warning 7068: in check_rr_node: RR node: 240813 type: CHANX location: (89,5
Error 1: 
Type: Routing
File: /tmp/really-really-really-really-really-really-really-really-really-really-really-really-really-long-path/conda/conda-bld/vtr_1546516453083/work/vpr/src/route/check_rr_graph.cpp
Line: 286
Message: in check_rr_node: CHANX out of range for endpoints (67,50) and (127,50)
```

To temporary solve the grid has to be expanded by `1` in the https://github.com/SymbiFlow/symbiflow-arch-defs/blob/1458f79c684c60554911c79b95d04caa7145c392/xc7/utils/prjxray_arch_import.py#L148.

litghost:
VPR only allows channels to run from 0 to dim-2.  To enable general channels within the grid, the underlying grid does need to be prxjray_dim+2.

PR #445, opened 2019-03-07
Initial SQLite3 XC7 routing graph implementation.

### LINKED ISSUE #292: Improve iteration time for 7-series
Currently database changes cause ~10-30 minutes of recomputation of the various intermediate products. 
Profile and determine if there are better algorithm/file output to reduction this

Goal is split into two priorities:
- Reduce time from arch.xml change to first pack
-  Reduce time from database change to first pack

litghost:
This converts the xc7 routing graph import from using channels.json to channels.db, a sqlite3 database.  This was done for several reasons:
 - Certain operations (e.g. pruning dead nodes) was fairly difficult in the previous data structure and layout.
 - Memory usage was very high as a side affect of keeping all data in memory
 - Parse times of channels.json (~800 MB) was high.
 - It was hard to see what was taking time, because database-like operations were in the code.  Now data lookup is explicit and can be optimized via SQL indicies or denormalized tables if shown to be required.

In addition, by moving to a database to keep relations between tiles, wires, nodes (prjxray node), graph_node (VTR rr_node), and graph_edges (VTR rr_edge) to the database, it should be easier to do optimizations on the graph, like merge nodes or splitting tiles (ala #415).

This also fixes #292 by avoiding recomputing the graph every time a pb_type XML is changed.  Instead the graph is computed only when prjxray DB changes.

This is fairly large rework of the routing import, but hopefully it is more understandable than the previous incarnation.

### DIFF
--- a/xc7/utils/prjxray_arch_import.py
+++ b/xc7/utils/prjxray_arch_import.py
@@ -147,17 +147,11 @@ def main():
     g = db.grid()
     x_min, x_max, y_min, y_max = g.dims()
 
-    # FIXME: There is an issue in the routing phase. (https://github.com/SymbiFlow/symbiflow-arch-defs/issues/353)
-    # if a zynq device is selected the grid must be expanded by 1
-    if args.device == 'xc7z010':
-        x_max += 1
-        y_max += 1
-
     name = '{}-test'.format(args.device)
     fixed_layout_xml = ET.SubElement(layout_xml, 'fixed_layout', {
             'name': name,
-            'height': str(y_max+1),
-            'width': str(x_max+1),
+            'height': str(y_max+2),
+            'width': str(x_max+2),
     })
 
     only_emit_roi = False
--- a/xc7/utils/prjxray_arch_import.py
+++ b/xc7/utils/prjxray_arch_import.py
@@ -109,9 +109,9 @@ def main():
     parser.add_argument(
             '--use_roi', required=False)
     parser.add_argument(
-            '--synth_tiles', required=False)
-    parser.add_argument(
             '--device', required=True)
+    parser.add_argument(
+            '--synth_tiles', required=False)
 
     args = parser.parse_args()
 
@@ -161,8 +161,6 @@ def main():
     })
 
     only_emit_roi = False
-    roi_inputs = []
-    roi_outputs = []
 
     synth_tiles = {}
     synth_tiles['tiles'] = {}
@@ -171,6 +169,9 @@ def main():
         with open(args.use_roi) as f:
             j = json.load(f)
 
+        with open(args.synth_tiles) as f:
+            synth_tiles = json.load(f)
+
         roi = Roi(
                 db=db,
                 x1=j['info']['GRID_X_MIN'],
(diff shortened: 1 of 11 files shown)


elmsfu:
Looks like this approach could have a lot of benefits. I may be out of the loop on other discussions around this. @litghost What alternatives have been considered?

As it gets closer I think a description of the tables and overall approach should be added before we merge.

litghost:
> @litghost What alternatives have been considered?

Do you mean alternative data stores?  Or do you mean alternative approaches?

elmsfu:
> > @litghost What alternatives have been considered?
> 
> Do you mean alternative data stores? Or do you mean alternative approaches?

Both

litghost:
> > > @litghost What alternatives have been considered?
> > 
> > 
> > Do you mean alternative data stores? Or do you mean alternative approaches?
> 
> Both

In terms of alternative data stores, SQLite is built into python3 distributions, requires no servers, is file oriented and is fairly battle tested.  Most of the operations that were giving me the feeling that I was re-implementing a poor mans database fit neatly into SQL.  Most of operations are relational oriented except for the final graph build in prjxray_create_edges, so I felt that a SQL db versus say a graph DB or NoSQL.  There is an open question of whether there should be two databases, one the is SQL for the relation queries (e.g. node<->wire<->pip) versus the graph itself.  However there are still some relational type queries done on the constructed graph (e.g. is this graph node alive?) that works very nicely in the SQL format.

If you have other concrete data store suggestions, I'm open to examining them.

In terms of alternative approaches, the initial xc7 implementation is an alternative in of itself.  The problems of that implementation (repeated work, expensive data passing, poor data querying) led me to the conclusion that a fully bespoke approach had reached its limit.  In particular, two tasks that were fairly challenging in the previous implement (dead node pruning and tile split) were signalling that another approach was required.  In terms of a third way, I don't have any bright ideas besides some external data store.  As I mentioned above, there is potentially value in fragmenting the data stores, but I don't believe we have good justification to do that at this time.

litghost:
> As it gets closer I think a description of the tables and overall approach should be added before we merge.

Added.

xc7/utils/prjxray_assign_tile_pin_direction.py
```diff
@@ -5,17 +5,202 @@
...
+        reason = NodeClassification(classification)
+
+        if reason == NodeClassification.NULL:
+            for (_, tile_type, wire) in yield_wire_info_from_node(conn, node_pkey):
+                null_tile_wires.add((tile_type, wire))
+
+        if reason != NodeClassification.EDGES_TO_CHANNEL:
+            continue
+
+        c2 = conn.cursor()
```
mithro: I would say the SQL queries here are inconsistently formatted.

litghost: Took another pass, take a look.

xc7/utils/prjxray_arch_import.py
```diff
@@ -147,22 +147,14 @@ def main():
...
-    if args.device == 'xc7z010':
-        x_max += 1
-        y_max += 1
-
     name = '{}-test'.format(args.device)
     fixed_layout_xml = ET.SubElement(layout_xml, 'fixed_layout', {
             'name': name,
-            'height': str(y_max+1),
-            'width': str(x_max+1),
+            'height': str(y_max+2),
```
elmsfu: Please add a comment on #353 explaining why it should always be +2 and close it, if appropriate.

litghost:
> Any initial measurement of runtime/memory improvement?

Right now everything is actually about the same.  However the new structure should be amendable to some easy optimizations.

Prjxray database change:
Before runtime:  20:54.88
After runtime: 22:12.28
Memory usage before: 25845448 kB
Memory usage after: 28814828 kB

Routing import only:
Before runtime: 7:05.24
After runtime: 6:57.06
Memory usage before: 25845448 kB
Memory usage after: 28814828 kB

I've started looking into what is going on with the memory usage during routing import in the new method, and with incremental XML generation, memory usage is down to 5075020 kB and runtime is down a minute.  There are further gains that are easy now that a database is storing the data throughout, so with some further PR's, can probably lower the runtime and memory foot print even further.

litghost:
With incremental edge import, memory usage on routing import is down to 3632632 kB with a small increase in runtime.  Probably not getting the memory usage down any more without further increases in runtime.

PR #457, opened 2019-03-13
SQLite database cache

mkurc-ant:
While working with arch defs I've noticed that recent changes regarding SQLite introduction dramatically slowed down process of prjxray database import. I've narrowed it down to problem with frequent writing of small chunks of data to the SQL database on disk. So I've created a wrapper which creates a temporary copy of the database in memory and all subsequent operations are performed on it.

I also added missing Python dependencies to the CMake build scripts.

### DIFF
--- a/xc7/utils/prjxray_assign_tile_pin_direction.py
+++ b/xc7/utils/prjxray_assign_tile_pin_direction.py
@@ -25,6 +25,8 @@ import progressbar
 import sqlite3
 import datetime
 
+from prjxray_db_cache import DatabaseCache
+
 now = datetime.datetime.now
 DirectConnection = namedtuple('DirectConnection', 'from_pin to_pin switch_name x_offset y_offset')
 
@@ -259,7 +261,8 @@ def main():
                 assert key not in edge_assignments, key
                 edge_assignments[key] = []
 
-    conn = sqlite3.connect(args.connection_database)
+    db_cache = DatabaseCache(args.connection_database)
+    conn = db_cache.get_connection()
 
     direct_connections = set()
     print('{} Processing direct connections.'.format(now()))
@@ -406,6 +409,9 @@ SELECT pkey, track_pkey FROM node WHERE classification = ?;
             'pin_directions': pin_directions,
             'direct_connections': [d._asdict() for d in direct_connections],
         }, f, indent=2)
+
+    print('{} Flushing database back to file "{}"'.format(now(), args.connection_database))
+    db_cache.close()
 
 if __name__ == '__main__':
     main()
--- a/xc7/utils/prjxray_assign_tile_pin_direction.py
+++ b/xc7/utils/prjxray_assign_tile_pin_direction.py
@@ -261,7 +261,7 @@ def main():
                 assert key not in edge_assignments, key
                 edge_assignments[key] = []
 
-    db_cache = DatabaseCache(args.connection_database)
+    db_cache = DatabaseCache(args.connection_database, read_only=True)
     conn = db_cache.get_connection()
 
     direct_connections = set()
--- a/xc7/utils/prjxray_assign_tile_pin_direction.py
+++ b/xc7/utils/prjxray_assign_tile_pin_direction.py
@@ -261,156 +261,155 @@ def main():
                 assert key not in edge_assignments, key
                 edge_assignments[key] = []
 
-    db_cache = DatabaseCache(args.connection_database, read_only=True)
-    conn = db_cache.get_connection()
+    with DatabaseCache(args.connection_database, read_only=True) as conn:
 
-    direct_connections = set()
-    print('{} Processing direct connections.'.format(now()))
-    handle_direction_connections(conn, direct_connections, edge_assignments)
+        direct_connections = set()
+        print('{} Processing direct connections.'.format(now()))
+        handle_direction_connections(conn, direct_connections, edge_assignments)
 
-    wires_not_in_channels = {}
-    c = conn.cursor()
-    print('{} Processing non-channel nodes.'.format(now()))
-    for node_pkey, classification in progressbar.progressbar(c.execute("""
-SELECT pkey, classification FROM node WHERE classification != ?;
-""", (NodeClassification.CHANNEL.value,))):
-        reason = NodeClassification(classification)
+        wires_not_in_channels = {}
+        c = conn.cursor()
+        print('{} Processing non-channel nodes.'.format(now()))
+        for node_pkey, classification in progressbar.progressbar(c.execute("""
+    SELECT pkey, classification FROM node WHERE classification != ?;
+    """, (NodeClassification.CHANNEL.value,))):
+            reason = NodeClassification(classification)
 
-        for (tile, tile_type, wire) in yield_wire_info_from_node(conn, node_pkey):
-            key = (tile_type, wire)
+            for (tile, tile_type, wire) in yield_wire_info_from_node(conn, node_pkey):
+                key = (tile_type, wire)
 
-            # Sometimes nodes in particular tile instances are disconnected,
-            # disregard classification changes if this is the case.
-            if reason != NodeClassification.NULL:
-                if key not in wires_not_in_channels:
-                    wires_not_in_channels[key] = reason
-                else:
-                    other_reason = wires_not_in_channels[key]
-                    assert reason == other_reason, (tile, wire, reason, other_reason)
+                # Sometimes nodes in particular tile instances are disconnected,
+                # disregard classification changes if this is the case.
+                if reason != NodeClassification.NULL:
+                    if key not in wires_not_in_channels:
+                        wires_not_in_channels[key] = reason
+                    else:
+                        other_reason = wires_not_in_channels[key]
+                        assert reason == other_reason, (tile, wire, reason, other_reason)
 
-            if key in wires_in_tile_types:
-                wires_in_tile_types.remove(key)
+                if key in wires_in_tile_types:
+                    wires_in_tile_types.remove(key)
 
-    # List of nodes that are channels.
-    channel_nodes = []
+        # List of nodes that are channels.
+        channel_nodes = []
 
-    # Map of (tile, wire) to track.  This will be used to find channels for pips
-    # that come from EDGES_TO_CHANNEL.
-    channel_wires_to_tracks = {}
+        # Map of (tile, wire) to track.  This will be used to find channels for pips
+        # that come from EDGES_TO_CHANNEL.
+        channel_wires_to_tracks = {}
 
-    # Generate track models and verify that wires are either in a channel
-    # or not in a channel.
-    print('{} Creating models from tracks.'.format(now()))
-    for node_pkey, track_pkey in progressbar.progressbar(c.execute("""
-SELECT pkey, track_pkey FROM node WHERE classification = ?;
-""", (NodeClassification.CHANNEL.value,))):
-        assert track_pkey is not None
+        # Generate track models and verify that wires are either in a channel
+        # or not in a channel.
+        print('{} Creating models from tracks.'.format(now()))
+        for node_pkey, track_pkey in progressbar.progressbar(c.execute("""
+    SELECT pkey, track_pkey FROM node WHERE classification = ?;
+    """, (NodeClassification.CHANNEL.value,))):
+            assert track_pkey is not None
 
-        tracks_model, _ = get_track_model(conn, track_pkey)
-        channel_nodes.append(tracks_model)
-        channel_wires_to_tracks[track_pkey] = tracks_model
+            tracks_model, _ = get_track_model(conn, track_pkey)
+            channel_nodes.append(tracks_model)
+            channel_wires_to_tracks[track_pkey] = tracks_model
 
-        for (tile, tile_type, wire) in yield_wire_info_from_node(conn, node_pkey):
-            tileinfo = grid.gridinfo_at_tilename(tile)
-            key = (tileinfo.tile_type, wire)
-            # Make sure all wires in channels always are in channels
-            assert key not in wires_not_in_channels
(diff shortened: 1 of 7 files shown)


litghost:
Can you actually profile the difference?  I ran similar tests, and the previous and current methods both took about the same time.

xc7/utils/prjxray_assign_tile_pin_direction.py
```diff
@@ -259,7 +261,8 @@ def main():
                 assert key not in edge_assignments, key
                 edge_assignments[key] = []
 
-    conn = sqlite3.connect(args.connection_database)
+    db_cache = DatabaseCache(args.connection_database)
```
litghost: assign_tile_pin_direction is read-only, the cache should have no effect.

mkurc-ant:
Well, before the change, on my system a single run of the `prjxray_create_edges.py` script took about 30min and I've noticed heavy HDD load (in terms of "busy time"). I'will do the tests again and post the results here.

litghost:
> Well, before the change, on my system a single run of the `prjxray_create_edges.py` script took about 30min and I've noticed heavy HDD load (in terms of "busy time"). I'will do the tests again and post the results here.

What was the time before?  On my machine (SSD), it took about ~20 minutes front to back for both methods.  It's possible that the SSD's write cache is deep enough to mask the difference.  I'll post numbers on my system for all three (old method, SQLite no memory, SQLite in memory), and we can compare with your numbers on the same three.

xc7/utils/prjxray_db_cache.py
```diff
@@ -0,0 +1,53 @@
+"""
+This is a database cache. Its a kind of proxy object for Python sqlite3 connection
+which operates on a memory copy of the database.
+
+Upon object creation the database is "backed up" to memory. All subsequent
+operations are then pefromed on this copy which yields in performance increase.
+
+It is important to explicitly call the close() method once database connection
+is no longer needed. The method "backs up" the database back to disk.
+"""
```
litghost: The typical solution to this problem is to use a context manager.  I suggest we take that approach here.

mkurc-ant:
Yeah, I'll do the timing measurement. I also tried the database on SSD. It performed better but not as good as when in RAM.

Actually I started this issue because I needed to debug some Python scripts and it took ages for them to reach the point which I was debugging.

xc7/utils/prjxray_db_cache.py
```diff
@@ -0,0 +1,53 @@
...
+import sqlite3
+
+# =============================================================================
+
+class DatabaseCache(object):
+
+    def __init__(self, file_name):
```
litghost: Depending on the results, let's make the cache optional in a transparent way.  Doing this will likely increase the memory usage significantly, but I'd like to get profile results.

xc7/utils/prjxray_db_cache.py
```diff
@@ -0,0 +1,53 @@
...
+    def close(self):
+
+        print("Dumping database to '{}'".format(self.file_name))
+        self.memory_connection.backup(self.file_connection, pages=1, progress=self._progress)
+        print("")
+
+        self.memory_connection.close()
+        self.file_connection.close()
+
+    def _progress(self, status, remaining, total):
```
litghost: Use progressbar.Progressbar?

xc7/utils/prjxray_db_cache.py
```diff
@@ -0,0 +1,53 @@
...
+class DatabaseCache(object):
+
+    def __init__(self, file_name):
+
+        self.file_name         = file_name
+        self.memory_connection = sqlite3.connect(":memory:")
+        self.file_connection   = sqlite3.connect(file_name)
+
+        print("Loading database from '{}'".format(self.file_name))
+        self.file_connection.backup(self.memory_connection, pages=1, progress=self._progress)
```
litghost: Probably want a larger page count?

litghost:
> Yeah, I'll do the timing measurement. I also tried the database on SSD. It performed better but not as good as when in RAM.
> 
> Actually I started this issue because I needed to debug some Python scripts and it took ages for them to reach the point which I was debugging.

SQLite Disk:

Built target buttons_basys3_bin
	Command being timed: "make buttons_basys3_bin"
	User time (seconds): 1552.84
	System time (seconds): 121.33
	Percent of CPU this job got: 154%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 18:03.73
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 6598532
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 12800402
	Voluntary context switches: 538230
	Involuntary context switches: 184411
	Swaps: 0
	File system inputs: 608
	File system outputs: 25199216
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

SQLite with In-memory cache:

	Command being timed: "make buttons_basys3_bin"
	User time (seconds): 1529.56
	System time (seconds): 62.60
	Percent of CPU this job got: 185%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 14:19.86
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 6599448
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 15410683
	Voluntary context switches: 223523
	Involuntary context switches: 7311
	Swaps: 0
	File system inputs: 448
	File system outputs: 15396752
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

Saving 4 minutes on an SSD, with no increase in peak memory usage, is a nice win.  Address the comments, and we can merge it.

xc7/utils/prjxray_db_cache.py
litghost: Given that peak memory usage didn't increase when using the in-memory cache, I think we can skip making it optional.

xc7/utils/prjxray_assign_tile_pin_direction.py
litghost: You can also enforce read-only with:
https://www.sqlite.org/pragma.html#pragma_query_only

mithro:
It might be worth checking if [libeatmydata](https://www.flamingspork.com/projects/libeatmydata/) makes a difference. It disables fsync which is probably the cause of the small writes.

litghost:
> It might be worth checking if [libeatmydata](https://www.flamingspork.com/projects/libeatmydata/) makes a difference. It disables fsync which is probably the cause of the small writes.

Given how cheap the "backup" call appears to be, the in-memory copy is likely sufficient.

xc7/utils/prjxray_db_cache.py
mkurc-ant: I'd rather not implement `__enter__ / __exit__` as that would require wrapping the whole code into `with` statement. I can make a destructor `__del__` (which is I suppose how file objects implement closing).

xc7/utils/prjxray_db_cache.py
mkurc-ant: Done.

xc7/utils/prjxray_db_cache.py
mkurc-ant: Done.

xc7/utils/prjxray_db_cache.py
mkurc-ant: I made a destructor that does the writeback.

mkurc-ant:
@litghost  Did timing tests on my machine:

**HDD without cache (SHA dfc4c53)** 
> 	Command being timed: "make buttons_basys3_bin -j 1"
	User time (seconds): 1498.16
	System time (seconds): 49.45
	Percent of CPU this job got: 64%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 39:55.78
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 6043044
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 1
	Minor (reclaiming a frame) page faults: 10676976
	Voluntary context switches: 502059
	Involuntary context switches: 102247
	Swaps: 0
	File system inputs: 8
	File system outputs: 25195720
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

**SSD without cache (SHA dfc4c53)** 
> 	Command being timed: "make buttons_basys3_bin -j 1"
	User time (seconds): 1539.01
	System time (seconds): 49.99
	Percent of CPU this job got: 168%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 15:41.42
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 6043472
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 4
	Minor (reclaiming a frame) page faults: 10662231
	Voluntary context switches: 423899
	Involuntary context switches: 138442
	Swaps: 0
	File system inputs: 0
	File system outputs: 25224216
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

**HDD with cache (SHA: 78ad2a9)**
> 	Command being timed: "make buttons_basys3_bin -j 1"
	User time (seconds): 1489.68
	System time (seconds): 24.17
	Percent of CPU this job got: 206%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 12:14.00
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 6043096
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 4
	Minor (reclaiming a frame) page faults: 11601844
	Voluntary context switches: 37175
	Involuntary context switches: 69340
	Swaps: 0
	File system inputs: 0
	File system outputs: 15398400
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

**SSD with cache (SHA: 78ad2a9)**
>	Command being timed: "make buttons_basys3_bin -j 1"
	User time (seconds): 1602.21
	System time (seconds): 24.83
	Percent of CPU this job got: 222%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 12:09.63
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 6043156
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 11932767
	Voluntary context switches: 36157
	Involuntary context switches: 125058
	Swaps: 0
	File system inputs: 104
	File system outputs: 15692136
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

So the conclusion is that without the cache when working on HDD its terribly slow. And when using the cache there is no difference between disk storage type.

xc7/utils/prjxray_assign_tile_pin_direction.py
mkurc-ant: Implemented the read-only option.

xc7/utils/prjxray_db_cache.py
litghost: Please implement __enter__ / __exit__, that is the way this should be handled.  __del__ has problems.  If you are concerned by the with statement, use an contextlib.ExitStack.

xc7/utils/prjxray_db_cache.py
```diff
@@ -0,0 +1,66 @@
...
+        self.file_connection.close()
+
+    def _progress(self, status, remaining, total):
+        """
+        Prints database copy progress.
+        """
+        if self.bar is None:
+            self.bar = ProgressBar(max_value = total)
+        else:
+            self.bar.update(total - remaining)
```
litghost: I believe this should just be update(remaining)?

mkurc-ant: Not really, according to the example from progressbar documentation (https://progressbar-2.readthedocs.io/en/latest/) it is fine.
```
import time
import progressbar

with progressbar.ProgressBar(max_value=10) as bar:
    for i in range(10):
        time.sleep(0.1)
        bar.update(i)
```

It also works correctly.

xc7/utils/prjxray_db_cache.py
mkurc-ant: Ok. So i'll do it in the `__enter__` / `__exit__` way

xc7/utils/prjxray_db_cache.py
litghost: Yep, I'm used to a count up, remaining counts down.

xc7/utils/prjxray_db_cache.py
mkurc-ant: Done.

xc7/utils/prjxray_create_edges.py
```diff
@@ -748,137 +750,140 @@ def main():
...
+        else:
+            use_roi = False
 
-    for loc in progressbar.progressbar(grid.tile_locations()):
-        gridinfo = grid.gridinfo_at_loc(loc)
-        tile_name = grid.tilename_at_loc(loc)
+        output_only_nodes = set()
+        input_only_nodes = set()
 
-        # Not a synth node, check if in ROI.
```
litghost: Some code was deleted?

Issue #338, opened 2019-01-24
Merge the channel.py and channel2.py files in utils


mithro:
We have both;
 * https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/lib/rr_graph/channel.py
 * https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/lib/rr_graph/channel2.py

It would be good to merge them and add a decent test suite.

Fleker:
Good evening,
I'm interested in getting involved in this project, and I can work on this issue.

Fleker:
Looks like channel2.py is only used by graphs2.py

mithro:
Merging grpahs.py and graphs2.py is also on the TODO list.

mithro:
There might also be useful information [in this Google doc](https://docs.google.com/document/d/17QBzXAul8ioGL7WeTXk0awwspGIDZBZgTYUKy5oT_Gg/edit?usp=drivesdk)

Fleker:
The easiest way to do this would be to just copy and paste, although it seems like Channel2.py was created due to some deep-seated bugs in Channel.py track layout. Would it make more sense to fix the Channel.py bug, or just merge the files?

litghost:
channel.py is simply wrong, so I wouldn't both to fix it.

mithro:
@litghost Do you remember what made you think that channel.py was wrong?

litghost:
> @litghost Do you remember what made you think that channel.py was wrong?

From channels2.py:

```
channel.py is actually broken for some track configuration, and generates
an excessive number of dummy tracks to fill empty space (channel.Channel = >2M
versus channel2.Channel ~70k).
```

There is also the fact that channels.py is ~1141 lines long and channels2.py is 40 lines long.

mithro:
channels.py is missing a lot of the pretty printing functionality,
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/7ad5525b8b5d8128ef5aed84bf000689712821f6/utils/lib/rr_graph/channel.py#L1051-L1068

https://github.com/SymbiFlow/symbiflow-arch-defs/blob/7ad5525b8b5d8128ef5aed84bf000689712821f6/utils/lib/rr_graph/channel.py#L1113-L1122

litghost:
[...]

PR #460, opened 2019-03-14
Created unit tests for Channel.py

Fleker:
As part of tackling the issue #338, this prepares by adding unit tests


mithro:
Looks like conda is having issues. Will rety this build in a couple of hours.

.travis/script.sh
```diff
@@ -68,3 +68,13 @@ make all
 end_section "symbiflow.build_all_demos"
 
 $SPACER
+
+start_section "symbiflow.run_python_tests" "Run Python unit tests"
+# Reset working directory
+cd /home/travis/build/SymbiFlow/symbiflow-arch-defs
+pip install -r ./requirements.txt
+python -m unittest utils.lib.rr_graph.tests.test_channel
+python -m unittest utils.lib.rr_graph.tests.test_graph
```
elmsfu: I think it'd be good to run tests via a build system target. We want to make sure we are testing against the python and libs from conda. I see we already have a [pytest](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/4101e067a4174e5bb7f54e0cb3f9bad222c14e0e/utils/CMakeLists.txt) target

I realize the dependencies aren't clean, so created #463 to improve the situation with python dependencies.

elmsfu: @Fleker Just a heads up, I probably just caused a conflict by moving this block up in #480

PR #462, opened 2019-03-14
Add arty support

acomodi:
This PR is to add arty support. It relies to the harness that can be generated with https://github.com/SymbiFlow/prjxray/pull/724.
Currently it is still under WIP because there is the use of an environmental variable which is a workaround to select the different fabric parts. We need to modify the CMake structure to allow a same `ARCH` to have different `PARTs`. Right now it is not possible to have same `ARCH` with different `PARTs`, therefore, the `if ("$ENV{BOARD}" STREQUAL "arty")`  statement is currently necessary.
To allow the `arty` board usage do the following in the root directory of `arch-defs`:
- `BOARD=arty make env`

There is also an addition of a simple `uart loopback` test and a counter test specific only to arty (this because only one led could be enabled).


acomodi:
Update on this PR:
to use the arty board `cmake` has to be run with an additional option:
`cmake -DBOARD=arty ..`.

The `cmake` build system has to be modified to allow a cleaner way in using different `parts` for the same architecture. IMO this PR is good to go for the short term.

litghost:
> The `cmake` build system has to be modified to allow a cleaner way in using different `parts` for the same architecture. IMO this PR is good to go for the short term.

Ya, once the ROI is removed sharing the architecture should be significantly easier, as just the ioplacement changes between arty and basys3.

acomodi:
Updated cmake build system to allow same `archs` to have different `parts`. This makes easier the definition of new `devices` and `boards`

litghost:
Assuming this has been tested on both basys3 and arty3, LGTM!

acomodi:
> Assuming this has been tested on both basys3 and arty3, LGTM!

Tested on both devices and both correctly compile and the basic tests to work on HW.
I'll wait for green travis

PR #471, opened 2019-03-15
utils: Move the connection db schema into a file.

mithro:
Means you get syntax highlighting and generate docs from the file.


utils/lib/connection_database.sql
```diff
@@ -0,0 +1,247 @@
...
+CREATE TABLE x_list(
+    idx INT,
+    info INT
+);
+CREATE TABLE y_list(
+    idx INT,
+    info INT
+);
+
+INSERT INTO
```
litghost: Leave the initial switch names in the python, this will become dynamic in the future.

mithro: Done.

PR #473, opened 2019-03-17
WIP: Initial 7-series FASM to verilog conversion

litghost:
This is the initial version of a FASM to verilog converter.  Requires that all bits from the original bitstream were converted into FASM tags.


mithro:
Conflict in `xc7/utils/prjxray_assign_tile_pin_direction.py`?

litghost:
> Conflict in `xc7/utils/prjxray_assign_tile_pin_direction.py`?

Fixed

litghost:
Because the ROI leads to output (sink) nets that sometimes have no connections, VPR design bitstreams should use `--allow_orphan_sinks` until the ROI is removed.

litghost:
Note that fasm2verilog requires that all bits are solved when converting from bit2fasm.  This does imply a database built with https://github.com/SymbiFlow/prjxray/pull/727

Issue #463, opened 2019-03-14
Revisit python dependencies and requirements file


elms:
We essentially have python package dependencies scoped in the cmake files, but also have a requirements file at the top level. We should only have a single method of tracking dependencies.

Do we need to scope any packages and have them installed only if you are building a target that depends on them?

litghost:
No, but there does need to be some link between dependency installation and targets that need them.

elmsfu:
We have several libraries and several executable modules. One idea is to wrap these up as packages with a setup.py so they can be installed in the env and tested independently. The granularity of dependencies would be less, but easier to describe in CMake (I think). For example, if a build needs something from xc7/utils you will probably need them all so no harm in that level of dependency.

mithro:
@elmsfu Which libraries / executables are you talking about?

I agree that we should probably revisit requirements.txt verse inside cmake files verse setup.py, etc. 

I definitely think the [third_party/prjxray](third_party/prjxray) ones should become proper Python modules (with setup.py and such). Unsure about the local stuff in this repo.

elmsfu:
These are top of mind:
 * [utils](https://github.com/SymbiFlow/symbiflow-arch-defs/tree/daee11d750e6ba6948f0c398690abd4dbb301189/utils)
 * [utils/lib](https://github.com/SymbiFlow/symbiflow-arch-defs/tree/daee11d750e6ba6948f0c398690abd4dbb301189/utils/lib)
 * [ice40/utils](https://github.com/SymbiFlow/symbiflow-arch-defs/tree/daee11d750e6ba6948f0c398690abd4dbb301189/ice40/utils)
 * [xc7/utils](https://github.com/SymbiFlow/symbiflow-arch-defs/tree/daee11d750e6ba6948f0c398690abd4dbb301189/xc7/utils)
 * [utils/vlog/yosys](https://github.com/SymbiFlow/symbiflow-arch-defs/tree/daee11d750e6ba6948f0c398690abd4dbb301189/utils/vlog/yosys)

mithro:
Just going through those directories;

 * `xc7/utils` are all command line tools
 * `ice40/utils` is mostly command line tools, with one library file `pcf.py`
 * `utils/*.py` seem to be mostly command line tools. Seems to be one library file `eblif.py`.
 * `utils/lib` these actually seem to be library / support functions with `utils/lib/rr_graph` and `utils/lib/rr_graph_xml` been strongly libraries.
[...]

PR #480, opened 2019-03-20
travis: use env variable to run python tests and run earlier

elmsfu:


.travis/script.sh
```diff
@@ -29,6 +29,16 @@ end_section "info.conda.config"
 
 $SPACER
 
+start_section "symbiflow.run_python_tests" "Run Python unit tests"
+# Reset working directory
+pushd $TRAVIS_BUILD_DIR
+pip install -r ./requirements.txt
+python -m unittest utils.lib.rr_graph.tests.test_graph
```
litghost: Should we use pytest here?  I don't think we explicit specific python paths on the prjxray python tests?

elmsfu: This change is not addressing that. It is just moving them earlier in travis script and using an environment variable instead of a hard-coded path so it will pass on forks.

We need to address #463 before we can start calling all the tests. We should also have them run as build target.

litghost:
Merge assuming travis is green.

PR #484, opened 2019-03-21
Forward make targets to build/ directory.

mithro:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,3 +10,8 @@ env:
 	git submodule init
 	git submodule update --init --recursive
 	mkdir -p build && cd build && cmake ..
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -10,3 +10,8 @@ env:
 	git submodule init
 	git submodule update --init --recursive
 	mkdir -p build && cd build && cmake ..
+
+.PHONY: Makefile
```
elmsfu: Not clear why we need this phony target. Maybe a comment?

mithro: It is actually unclear to me too. It was copied from the sphinx Makefile which does similar forwarding. Removing it means it stops working...

PR #490, opened 2019-03-22
Scalable test design

mkurc-ant:
I've created a scalable test design. The design consists of a ROM with data (an initialized BRAM), a configurable number of parallel "processing units" and output module which sends the processed data through UART. For now each processing unit is a 16x16 multiplier implemented using LUTs.

Basically the design reads N 32-bit words from an initialized BRAM, processes them in parallel using N processing units and then sends the results through the UART.

The whole design is artificially slowed down to <1MHz (100MHz / 128) to rule out (hopefully) timing issues of VPR. The UART works on 9600 bps.

The test design comes up with a python script which receives the data through UART and compares it with pre-computed expected pattern.

I did some tests for for different values of N (number of parallel workers). Here are the results:

N=1: Working fine
N=2: Working fine
N=3: Working fine
N=4: ASCII characters being sent are messed up
N=5: Working, but there are spurious one-bit errors in the computation results.
N=6: Working, but the computation results are totally wrong.

The tests has been done for:
- symbiflow-arch-defs SHA: f4ce9e1
- Yosys SHA: 87846729 (built locally)
- VPR SHA: f725c5ac0 (built locally)

I did also a test in which I used Yosys + Vivado and the design worked perfectly for N from 1 to 8.

I am sharing the design here because in my opinion it may allow to identify potential problems with VPR by gradually increasing design complexity rather than jumping from a simple counter to a SoC.


litghost:
Waiting for green travis, then good to merge.

elmsfu:
Is there any reason this shouldn't live in the top layer tests?

litghost:
> Is there any reason this shouldn't live in the top layer tests?

No, that's valid.  Would be good to test N=1-6 on ice40 too.  Let's move it to <root>/tests then.

mkurc-ant:
So, shall I move it to `<root>/tests` ?

litghost:
> So, shall I move it to `<root>/tests` ?

Yep, and then I think this is LGTM.

litghost:
Still showing a conflict with xc7/tests/CMakeLists.txt, probably just revert it?

mkurc-ant:
Ok, I'll try

litghost:
CI is failing, https://travis-ci.org/SymbiFlow/symbiflow-arch-defs/builds/511590955?utm_source=github_status&utm_medium=notification

mkurc-ant:
Yep it is failing, however on my local machine it passes the CMake... I'll investigate that.

acomodi:
@mkurc-ant could it be due to https://github.com/SymbiFlow/symbiflow-arch-defs/pull/462?

elmsfu:
@mkurc-ant I think this is because we are [skipping inclusion of xc7](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/CMakeLists.txt#L78-L80) for CI (i.e. Travis) so it doens't know about basys3 when you include it. May need to update [here](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/tests/CMakeLists.txt#L2-L4)

litghost:
xc7 is not used on CI right now because it exceeds the maximum memory of the CI.

elmsfu:
> xc7 is not used on CI right now because it exceeds the maximum memory of the CI.

Correct. Currently the scalable test is trying to run on xc7 on Travis and we can't do that. The scalable test will have to be skipped on CI for now, similar to the link to picosoc I included.

mkurc-ant:
Ahhhh... I forgot about that basys3 target is not available in CI. So I disabled the design for CI builds as it is done with the PicoSoC.

PR #492, opened 2019-03-22
Use shorts instead of delayless switches when joining two channels together

litghost:


mithro:
Just FYI - I have an issue around the switches here https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/354

litghost:
> Just FYI - I have an issue around the switches here [verilog-to-routing/vtr-verilog-to-routing#354](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/354)

Do you want to hold this PR for that?

mithro:
No, but maybe we should add a comment pointing to it?

PR #502, opened 2019-03-26
Modify BRAM and DRAM tests to use all top level pins.

litghost:
This is required for clean fasm2bel usage.


litghost:
> Looks fine to me. Please give some detail on why it's required for clean fasm2bel. Context can help others solve similar issues.

fasm2bel by default requires that all sinks (e.g. OBUF in this case) have a source.  In the ROI, if the IO is not used, then the sink lacks a source, etc.  This change ensures that all sinks have a source.  This can be backed out when we stop using the ROI.

PR #501, opened 2019-03-26
WIP: Fixes for based FASM output found via fasm2bels.

litghost:
- Output CEUSEDMUX and SRUSEDMUX only when required (e.g. CE or SR
  signal is used).
- Only output FFSYNC if FDSE or FDRE is used.
- Only output DLUT.RAM if a DRAM is actually used.
- Change order of DRAM packing to pack into DLUT before CLUT, etc.

PR #503, opened 2019-03-26
fasm2bels implementation.

litghost:
This correctly works on designs with the following restrictions:
 - IO's must not use IDDR, ODDR, ISERDES, or OSERDES
 - IO's must be single ended
 - Only sites from IOB33, IOI33, CLBxx_x, BRAM_x, BUFGCTRL and BUFHCE tiles are supported.
 - RAMB36 is not supported
 - RAMB18 designs will have different routing for constants connected to the WEBWE and WEA ports.  This results in a significant different in routing pips, but should result in functionally identical output.
 - Constant routes are not explicitly specified, resulting in a small feature variations, but no functional differences.
 - VPR designs should use all top-level ports, or significant differences will result.

### DIFF
--- a/make/devices.cmake
+++ b/make/devices.cmake
@@ -1143,41 +1143,6 @@ function(ADD_FPGA_TARGET)
 
     add_custom_target(${NAME}_bit ALL DEPENDS ${OUT_BITSTREAM})
 
-    get_target_property_required(NO_BIT_TO_V ${ARCH} NO_BIT_TO_V)
-    if(NOT ${NO_BIT_TO_V})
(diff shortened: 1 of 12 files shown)


litghost:
For VPR designs, it is highly recommended to use designs that have https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/33 and https://github.com/SymbiFlow/symbiflow-arch-defs/pull/501 merged into the local branch.  https://github.com/SymbiFlow/symbiflow-arch-defs/pull/502 modifies the DRAM and BRAM tests to conform with the "use all top-level port" suggestion.

PR #504, opened 2019-03-27
Move output writing into Yosys tcl script.

litghost:
This is required to enable 7-series constant generator support using the
write_blif -true/-false arguments.


mithro:
I don't see `-true`/`-false` arguments?

litghost:
> I don't see `-true`/`-false` arguments?

That's coming later, I wanted to refactor and merge this change first.

mithro:
Okay, LGTM I guess...

PR #510, opened 2019-03-28
Selectable ROM implementation (BRAM or DRAM) for the "9-scalable_test" design

mkurc-ant:
I added support for explicitly instantiated DRAMs (RAM64X1D) to the `9-scalable_test` design. The ROM implementation and its content is generated by a Python script. I had to limit size of the ROM data when using DRAM to 64 words in order not to use too many DRAMs and logic for muxes.

The CMake now generates targets `top_bram_xxx` and `top_dram_xxx`


litghost:
LGTM!

PR #511, opened 2019-03-28
Initial constant routing support for 7-series

litghost:
This implements the routing modifications required for constant support in 7-series parts.  It is not currently working because the router cannot handle U-shaped routes.  Investigations will continue in a seperate branch to isolate the failure and come up with a proposed modifications to this PR and to modify VPR if required.


xc7/archs/artix7/devices/xc7a50t-arty-roi-virt/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ add_xc7_device_define_type(
   DEVICE xc7a50t-arty
   ROI_PART xc7a35tcsg324-1
   ROI_DIR ${PRJXRAY_DB_DIR}/artix7/harness/arty-a7/uart
-  TILE_TYPES INT_R INT_L CLBLL_R CLBLL_L CLBLM_R CLBLM_L BRAM_L
+  TILE_TYPES CLBLL_R CLBLL_L CLBLM_R CLBLM_L BRAM_L
```
mkurc-ant: Why remove importing of INT_L and INT_R ?

litghost: Because the only thing located in the INT_L/INT_R tile is the TIEOFF site.  With this modification the TIEOFF site doesn't get used, instead those site pins are tied directly into the global constant channel.

litghost:
> Why remove importing of INT_L and INT_R ?

Answered in the thread.

litghost:
This is now working with the latest VPR from conda.  That VPR will be integrated into symbiflow-arch-defs in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/518, which is currently failing CI, but will then be merged.

There is a latent bug in how the CE graph is expressed that I'm working on.

litghost:
I've convinced myself that there is NOT a latent bug in the CE connection, but I'm going to add a huge comment about it.

xc7/primitives/slicem/slicem.pb_type.xml
```diff
@@ -629,7 +617,17 @@
     <direct name="A5" input="BLK_IG-SLICEM.A5" output="BLK_IG-SLICEM_MODES.A5" />
     <direct name="A6" input="BLK_IG-SLICEM.A6" output="BLK_IG-SLICEM_MODES.A6" />
 
-    <direct name="CK2" input="BLK_IG-SLICEM.CLK" output="BLK_IG-SLICEM_MODES.CLK"/>
+    <direct name="CK2" input="BLK_IG-SLICEM.CLK" output="BLK_IG-SLICEM_MODES.CLK">
+        <!-- The DLUT must be in RAM-mode for an of the RAM's to work.
```
mithro:
```suggestion
        <!-- The DLUT must be in RAM-mode for any of the RAM's to work.
```

xc7/primitives/slicem/slicem.pb_type.xml
```diff
@@ -629,7 +617,17 @@
...
+             As a corrolary, a DRAM requires the clock, so only turn on the
```
mithro:
```suggestion
             As a corollary, a DRAM requires the clock, so only turn on the
```

xc7/utils/prjxray_constant_site_pins.py
```diff
@@ -0,0 +1,66 @@
...
+also be emitted.
+
+WIRE_TIE_OPTIONS is a map of wires connected to site pins to the list of tie
+options available for the pin.
+
+"""
+
+
+# Using these wires requires a mux to be enabled.
+FEATURES_WHEN_ROUTED = {
```
mithro: I'm not sure these `XXUSEDMUX` are correct.

Shouldn't these just be the following inside the `<pb_type>`;
```xml
<interconnect>
<mux name="SRUSEDMUX" input="BLK_IG-SLICE_X1.CE" output="SLICE.CE" >
  <metadata>
    <meta name="fasm_mux">
      COMMON_SLICE_X1.CE : SRUSEDMUX
    </meta>
  </metadata>
</mux>
</interconnect>
```

xc7/yosys/synth.tcl
```diff
@@ -15,5 +15,9 @@ opt_clean
 
 setundef -zero -params
 stat
-write_blif -attr -cname -param $::env(OUT_EBLIF)
+write_blif -attr -cname -param \
+ -true VCC VCC \
+ -false GND GND \
+ -undef VCC VCC \
```
mithro: I'm unsure if undefined actually be mapped to VCC? Can you add a comment about why it is a good idea?

xc7/utils/prjxray_form_channels.py
```diff
@@ -676,13 +687,106 @@ def form_tracks(conn):
...
+    track_pkeys = insert_tracks(conn, tracks_to_insert)
+    vcc_track_pkey = track_pkeys[vcc_idx]
+    gnd_track_pkey = track_pkeys[gnd_idx]
+
+    c.execute("""
+INSERT INTO constant_sources(vcc_track_pkey, gnd_track_pkey) VALUES (?, ?)
+        """, (vcc_track_pkey, gnd_track_pkey,))
+
+    conn.commit()
+
```
mithro: Maybe split the stuff below into its own function?
```suggestion
    convert_hardpins_to_constant_network(....)


def convert_hardpins_to_constant_network(...):
```

xc7/yosys/synth.tcl
litghost: Because that is what Xilinx's tools do?

xc7/utils/prjxray_constant_site_pins.py
litghost: That doesn't allow support for tying of SR and CE to VCC.  What you have is what there was before, which requires a routed VCC rather than the HARD1 inside of the site.

litghost: It does make me wonder whether a techmap to disconnect SR and CE might work better.

litghost:
> Only major question is about the `XXUSEDMUX`, otherwise LGTM.

I've removed the XXUSEDMUX change for now.  I think the original approach taken here was wrong, and needs another go.

xc7/utils/prjxray_constant_site_pins.py
litghost: Removed the XXUSEDMUX changes in light of found complexity.  Will take another pass at them in after merging the other changes.

PR #515, opened 2019-03-28
Add 10x10 test arch and add const sources.

litghost:
This is required to explore global routing errors found in #511.


mithro:
I think this looks okay for now. However, it is unclear to me if we want **all** the testarchs to have constant sources? I think we possibly want testarchs which explore the different types of constant routing options?

One problem I ran into with the testarch stuff is how quickly it became hard to manage the different aspects and combinations of things we wanted to explore....

mithro:
Oh there was some attempt at doing constant generation previously that should probably be removed - see https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/testarch/primitives/cgen/cgen.pb_type.xml#L11-L16

litghost:
> I think this looks okay for now. However, it is unclear to me if we want **all** the testarchs to have constant sources? I think we possibly want testarchs which explore the different types of constant routing options?
> 
> One problem I ran into with the testarch stuff is how quickly it became hard to manage the different aspects and combinations of things we wanted to explore....

Ya.  We could make multiple versions of the yosys tcl script, but I don't have a good idea on how to do that.  However I think having the test arch have constant sources in the layout makes more sense than not, and we can explore how to leverage them in different ways, e.g. have yosys emit a LUT, have yosys emit a subckt, and compare relative performance.

PR #519, opened 2019-04-01
Add target to use yapf to check all python files

elmsfu:


litghost:
Should we just add the start of the make format target too?

elmsfu:
> Should we just add the start of the make format target too?

I don't understand what you are suggesting. Can you clarify? a link or suggestion would help.

litghost:
> > Should we just add the start of the make format target too?
> 
> I don't understand what you are suggesting. Can you clarify? a link or suggestion would help.

In prjxray, this check is implemented as:
make format
git diff 
if $? != 0 -> needs format

https://github.com/SymbiFlow/prjxray/blob/master/Makefile#L62

I believe yapf -i is "update file in place".

Anyways, I like the idea of a check format target (good for pre-push hooks), but the more important target is "make format", e.g. the thing that will fix the formatting.

mithro:
Do we already have a yapf config file?
We use the following file for [Project X-Ray](https://github.com/SymbiFlow/prjxray/blob/master/.style.yapf)

elmsfu:
> Do we already have a yapf config file?
> We use the following file for [Project X-Ray](https://github.com/SymbiFlow/prjxray/blob/master/.style.yapf)

I don't see one anywhere. I can use the same style and add that to #522

mithro:
Let's add the config file.

mithro:
DCO.

PR #520, opened 2019-04-01
Reenable pytest

elmsfu:


utils/lib/rr_graph/channel.py
```diff
@@ -1053,7 +1053,7 @@ def test_pad():
     >>> docprint(test_pad())
     ['| 0  | 1  | 2  | 3  | 4  | 5  |',
      '       AA>---------------->AA ',
-     '       BB>------>BB   XX   XX ',
```
elmsfu: I think this is the wrong change, but thought it'd be easier to ask in review than look closer last week.

utils/CMakeLists.txt
```diff
@@ -1,7 +1,16 @@
 get_target_property_required(PYTEST env PYTEST)
 
+# TODO: figure out a better way instead of this hack. Maybe add_test
+# just sucks or maybe we should write a wrapper test driver.
+add_test(
```
litghost: CTest never builds targets.  The solution is something like:
make prepare_for_tests
make test

elmsfu: That sucks and confirms my thought from the comment.

I saw that if the test uses a target added by `add_executable`[see `add_test` COMMAND](https://cmake.org/cmake/help/v3.10/command/add_test.html) it will use the once built.

litghost: True, but it doesn't link the file dependency, just replaces the executable with the true path.

litghost: https://stackoverflow.com/questions/733475/cmake-ctest-make-test-doesnt-build-tests

elmsfu: This makes be conclude that ctest is not a good solution and that we should just wrap tests in shell script and call as normal custom targets.

Links related to open feature request to fix this in CMake:
 * [old mantis bug](https://cmake.org/Bug/view.php?id=8774)
 * [current tracking issue](https://gitlab.kitware.com/cmake/cmake/issues/8774)

utils/lib/rr_graph/channel.py
elmsfu: @mithro thoughts on this?

mithro: Does the test now pass?

mithro: It looks like the `g.fill_empty(0, name='XX')` isn't working?

utils/pytest.ini
```diff
@@ -1,3 +0,0 @@
-[pytest]
```
mithro: Why is this file being removed?

elmsfu: Its scope was limited and the purpose is redundant.

mithro: It looks like it has lost the `norecursedirs = vlog` setting? Is that no longer needed...

elmsfu: Correct. It seems to do fine on everything so whatever required it in the [commit](https://github.com/SymbiFlow/symbiflow-arch-defs/commit/e6c4952ce7c2c8278aa57e01d71c8bedf48c97a3) seems to be resolved.

utils/lib/rr_graph/channel.py
elmsfu: Fixed properly now. Using `yield` with calls that modify state is a bad combination.

litghost:
Did the submodule changes get added by accident?

elmsfu:
> Did the submodule changes get added by accident?

Yes. Fixed now. Good thing we have reviews when I'm context switching.

utils/pytest.ini
mithro: Great!

utils/lib/rr_graph/tests/test_graph.py
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
-# Run `python3 -m unittest utils.lib.rr_graph.tests.test_graph`
+# Run `python3 -m unittest lib.rr_graph.tests.test_graph`
 import unittest
 
-from .. import graph, P, Size
```
mithro: Why have you changed these to absolute imports?

elmsfu: pytest failed on these imports. I think this is because of how the modules are laid out. I haven't dug into it but I believe we need to separate some of our code some and make sure we take care of some details around module best practices.

mithro: I think we want lib/rr_graph to be its own library and then we want to add `utils/lib` to python path and then import this as `import rr_graph`or `pytest rr_graph` style thing?

PR #521, opened 2019-04-02
WIP: add RAMB36 support

tmichalak:
Signed-off-by: Tomasz Michalak <tmichalak@antmicro.com>


tmichalak:
@litghost I saw that in the database for RAMB36 we have the config bits (in segbits_bramx.db), but for the content we only have RAMB18. Do we need explicit RAMB36 or is what we have now sufficient? I mean RAMB36 is constructed of RAMB18 anyway.

utils/mux_gen.py
```diff
@@ -67,8 +67,7 @@
     '--name-inputs',
     type=str,
     default=None,
-    help=
-    "Comma deliminator list for the name of each input to the mux (implies --split-inputs)."
+    help="Comma deliminator list for the name of each input to the mux (implies --split-inputs)."
```
litghost: I think you ingested a revert commit, can you clean up the PR?

litghost:
> @litghost I saw that in the database for RAMB36 we have the config bits (in segbits_bramx.db), but for the content we only have RAMB18. Do we need explicit RAMB36 or is what we have now sufficient? I mean RAMB36 is constructed of RAMB18 anyway.

RAMB36 is simply the two underlying RAMB18's merged.  So you should have all the bits required to make this change.

litghost:
Does this work on hardware?

tmichalak:
It doesn't work on hardware neither it passes routing. I still have to work on the interconnections part.

tmichalak:
@litghost: I chose to add a dedicated ramb36e1 model, but I am wondering where the rambfifo36e1 model fits into the flow. The rambfifo36e1.pb_type.xml refers to RAMB36E1_PRIM module which has not been implemented so far. What is the use purpose of  rambfifo36e1.pb_type.xml and do we want to use the newly created RAMB36E1_VPR cell definition for the BRAM mode of rambfifo36?

tmichalak:
@litghost: After adding the bram36 model and pb_type I got stuck on vpr traying to pack the molecule:
<pre>
Circuit Statistics:
  Blocks: 2407
    .input      :      18
    .output     :      17
    0-LUT       :       2
    5-LUT       :    1176
    CARRY       :     188
    CARRY0      :      73
    FDRE_ZINI   :     502
    MUXF6       :     315
    MUXF7       :      88
    MUXF8       :      27
    RAMB36E1_VPR:       1
  Nets  : 2497
    Avg Fanout:     3.6
    Max Fanout:   786.0
    Min Fanout:     1.0
  Netlist Clocks: 1
...
Error 1: 
Type: Packing
/tmp/really-really-really-really-really-really-really-really-really-really-really-really-really-long-path/conda/conda-bld/vtr_1552671508380/work/vpr/src/pack/cluster.cpp
Line: 1973
Message: Can not find any logic block that can implement molecule.
	Atom bram.ram.0.0.0 (RAMB36E1_VPR)
</pre>
Do you happen to know at which stage vpr figures out it has to use two BRAM18 primitives instead of BRAM36? If it's the pb_type then I will have to modify the ramb36 pb_type I've added.

litghost:
You need to implement RAMB36E1_VPR, not RAMB36E1.

tmichalak:
Right, I am not sure if I have done that. Could you explain the difference?

litghost:
RAMB36E1 is the Xilinx tech library version, RAMB36E1_VPR is our VPR specific black box.  It has some differences in parameters mostly, but we change some connections around to match the underlying hardware.

tmichalak:
Then I guess the error message I get in VPR I pasted few post before, results from some incorrectly specified connections?

litghost:
> Then I guess the error message I get in VPR I pasted few post before, results from some incorrectly specified connections?

You mean:
```
Message: Can not find any logic block that can implement molecule.
	Atom bram.ram.0.0.0 (RAMB36E1_VPR)
```

It could mean a number of things, one of which is incorrect connections.  Enable VPR verbosity to determine which your model wasn't used.

tmichalak:
Ok, I understand. Last question then. Since eventually in fasm we will see bram18_y0 and bram18_y1 is the bram36 just a pack pattern consisting of bram18s? I want to understand how the bram36_vpr model gets translated into molecules that eventually become bram18 in fasm.

litghost:
> Ok, I understand. Last question then. Since eventually in fasm we will see bram18_y0 and bram18_y1 is the bram36 just a pack pattern consisting of bram18s? I want to understand how the bram36_vpr model gets translated into molecules that eventually become bram18 in fasm.

It is possible to emit the BRAM36 as two BRAM18's, but I don't believe it the prudent path.  Instead have VPR explicitly map BRAM36 features to each of the  BRAM18 features for the BRAM36, for example DOA_REG -> RAMB18_Y0.DOA_REG and DOA_REG -> RAMB18_Y1.DOA_REG.

tmichalak:
superseded by #727

PR #522, opened 2019-04-02
Mass yapf reformat

elmsfu:


mithro:
Do we already have a yapf config file?
We use the following file for [Project X-Ray](https://github.com/SymbiFlow/prjxray/blob/master/.style.yapf)

elmsfu:
> #519 should be merged first right?

Yep hence the WIP.

.style.yapf
```diff
@@ -4,3 +4,5 @@ split_before_expression_after_opening_paren = True
 split_before_first_argument = True
 split_complex_comprehension = True
 split_penalty_comprehension = 2100
+dedent_closing_brackets=True
```
elmsfu: @mithro split_before_first_argument was already true from the style.yapf in prjxray. If this is the style should we make it consistent across Symbiflow projects?

mithro: My suggestions were to add;
```python
DEDENT_CLOSING_BRACKETS=True
INDENT_DICTIONARY_VALUE=True
SPLIT_BEFORE_FIRST_ARGUMENT=True 
```

And you are correct, we should keep them consistent across SymbiFlow repositories. So that looks like;
 * https://github.com/SymbiFlow/symbiflow-arch-defs
 * https://github.com/SymbiFlow/prjxray
 * https://github.com/SymbiFlow/fasm
 * https://github.com/SymbiFlow/fpga-tool-perf
?

elmsfu: I'm not sure I understand. Are they case sensitive?

mithro: No, your version is fine. Are all three values set?

elmsfu: Yes

elmsfu:
> This all looks good to me - you have submodule changes again though?

uggh I keep falling back to a bad habit that results in this. Thanks. I rebased to resolve merge conflict and updated the sha in the `.git-ignore-blame-revs`. Will merge once travis shows green.

mithro:
You might want to try the following git config to help with that;
```
	git config status.submodulesummary 1
	git config push.recurseSubmodules check
	git config diff.submodule log
	git config checkout.recurseSubmodules 1
	git config alias.sdiff '!'"git diff && git submodule foreach 'git diff'"
	git config alias.spush 'push --recurse-submodules=on-demand'
```

PR #529, opened 2019-04-03
WIP: Initial version of generic test fixture.

litghost:
This is untested, but I wanted to get this out for Verilog review.


elmsfu:
Looks good to this non expert.

elmsfu:
There is a similar approach that I mentioned in Symbiflow/ideas#25  Maybe we can consider leveraging/contributing upstream https://github.com/daveshah1/hilotof

PR #537, opened 2019-04-05
Tile grid location mapping

mkurc-ant:
I've implemented a generic mechanism which allows to map tile locations in an arbitrary way. In this implementation new tables are added to the `channels.db` database:

 - `phy_tile`: This is the same table as `tile` but now it stores the tile grid as imported from the prjxray database. The `tile` table now stores the tile grid to be used with VPR. Primary keys of corresponding entries in those tables are equal.
- `grid_loc_map`: The table stores pairs of (x,y) coordinates thus provides a map between the physical grid (`phy_tile` table) and the VPR grid (the `tile`) table.

It is worth noting that in general the grid mapping can be one-to-many as it is meant to be used for splitting a CLB at one location into two SLICEs which will have two separate locations.

There is also the `grid_splitter.py` script which, for testing purposes, inserts an empty column east of each CLB  column. Therefore it simulates splitting CLBs into SLICEs but without doing the actual split.

Wile working on the problem of location mapping I managed to remove prjxray database dependency from `prjxray_assign_tile_pin_direction.py`, `prjxray_create_edges.py` and `prjxray_routing_import.py`. All information necessary for these scripts are already in `channels.db`

### DIFF
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -262,3 +262,11 @@ CREATE TABLE constant_sources(
     FOREIGN KEY(vcc_track_pkey) REFERENCES track(pkey),
     FOREIGN KEY(gnd_track_pkey) REFERENCES track(pkey)
 );
+
+-- Grid location map.
+CREATE TABLE grid_loc_map(
+    phy_tile_pkey INT,
+    vpr_tile_pkey INT,
+    FOREIGN KEY(phy_tile_pkey) REFERENCES phy_tile(pkey),
+    FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey)
+);
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -51,7 +51,17 @@ CREATE TABLE site_type(
   name TEXT
 );
 
--- Tile table, contains type and name of tile and location in grid.
+-- Tile table, contains type and name of tile and location in the physical grid.
+CREATE TABLE phy_tile(
+  pkey INTEGER PRIMARY KEY,
+  name TEXT,
+  tile_type_pkey INT,
+  grid_x INT,
+  grid_y INT,
+  FOREIGN KEY(tile_type_pkey) REFERENCES tile_type(pkey)
+);
+
+-- Tile table, contains type and name of tile and location in the VPR grid.
 CREATE TABLE tile(
   pkey INTEGER PRIMARY KEY,
   name TEXT,
--- a/utils/lib/grid_mapping.py
+++ b/utils/lib/grid_mapping.py
@@ -6,37 +6,109 @@ import itertools
 
 # =============================================================================
 
+
 class GridLocMap(object):
     """
     This class preforms forward (physical -> VPR) and backward (VPR -> physical)
     grid location mapping.
     """
 
-    def __init__(self, db_conn):
+    def __init__(self, fwd_loc_map, bwd_loc_map):
+        self.fwd_loc_map = fwd_loc_map
+        self.bwd_loc_map = bwd_loc_map
 
-        # Create a database cursor
-        self.db_cursor = db_conn.cursor()
+    @staticmethod
+    def generate_one_to_one_map(extent):
+        """
+        Generates a one-to-one map for specified location range.
+        :param extent:
+        :return:
+        """
 
-        # Get the whole table
-        grid_loc_map = self.db_cursor.execute(
-            "SELECT grid_phy_x, grid_phy_y, grid_vpr_x, grid_vpr_y FROM grid_loc_map"
-        ).fetchall()
+        xmin, ymin, xmax, ymax = extent
+
+        fwd_loc_map = {}
+        bwd_loc_map = {}
+
+        # Make a one-to-one map
+        for x, y in itertools.product(range(xmin, xmax + 1),
+                                      range(ymin, ymax + 1)):
+
+            fwd_loc_map[(x, y)] = [(x, y)]
+            bwd_loc_map[(x, y)] = [(x, y)]
+
+        # Return map object
+        return GridLocMap(fwd_loc_map, bwd_loc_map)
+
+    @staticmethod
+    def generate_shift_map(extent, shift_x, shift_y):
+        """
+        Generates a one-to-one map for specified location range. For debugging
+        purposes.
+        :param extent:
+        :param shift_x:
+        :param shift_y:
+
+        :return:
+        """
+
+        xmin, ymin, xmax, ymax = extent
+
+        fwd_loc_map = {}
+        bwd_loc_map = {}
+
+        # Make a one-to-one map
+        for x, y in itertools.product(range(xmin, xmax + 1),
+                                      range(ymin, ymax + 1)):
+
+            phy_loc = (x, y)
+            vpr_loc = (x + shift_x, y + shift_y)
+
+            fwd_loc_map[phy_loc] = [vpr_loc]
+            bwd_loc_map[vpr_loc] = [phy_loc]
+
+        # Return map object
+        return GridLocMap(fwd_loc_map, bwd_loc_map)
+
+    @staticmethod
+    def load_from_database(conn):
+        """
+        Loads grid location mapping from a SQL database. Returns a GridLocMap
+        object.
+        :param conn:
+        :return:
+        """
+
+        c = conn.cursor()
+
+        # Query the grid map from the database
+        grid_loc_map = c.execute("""
+SELECT phy.grid_x, phy.grid_y, vpr.grid_x, vpr.grid_y
+FROM phy_tile phy
+INNER JOIN grid_loc_map map
+ON phy.pkey = map.phy_tile_pkey
+INNER JOIN tile vpr
+ON vpr.pkey = map.vpr_tile_pkey
+""").fetchall()
 
         # Build maps
-        self.fwd_loc_map = {}
-        self.bwd_loc_map = {}
+        fwd_loc_map = {}
+        bwd_loc_map = {}
 
         for loc_pair in grid_loc_map:
             phy_loc = loc_pair[0:2]
             vpr_loc = loc_pair[2:4]
 
-            if phy_loc not in self.fwd_loc_map.keys():
-                self.fwd_loc_map[phy_loc] = []
-            self.fwd_loc_map[phy_loc].append(vpr_loc)
+            if phy_loc not in fwd_loc_map.keys():
+                fwd_loc_map[phy_loc] = []
+            fwd_loc_map[phy_loc].append(vpr_loc)
 
-            if vpr_loc not in self.bwd_loc_map.keys():
-                self.bwd_loc_map[vpr_loc] = []
-            self.bwd_loc_map[vpr_loc].append(phy_loc)
+            if vpr_loc not in bwd_loc_map.keys():
+                bwd_loc_map[vpr_loc] = []
+            bwd_loc_map[vpr_loc].append(phy_loc)
+
+        # Return map object
+        return GridLocMap(fwd_loc_map, bwd_loc_map)
 
     def get_vpr_loc(self, grid_loc):
         return tuple(self.fwd_loc_map[grid_loc])
@@ -45,6 +117,7 @@ class GridLocMap(object):
         return tuple(self.bwd_loc_map[grid_loc])
 
 # =============================================================================
+
 
 def create_tables(conn):
     """
@@ -57,6 +130,7 @@ def create_tables(conn):
         cursor = conn.cursor()
         cursor.executescript(fp.read())
         conn.commit()
+
 
 def get_phy_grid_extent(conn):
     """
@@ -76,6 +150,7 @@ def get_phy_grid_extent(conn):
 
     return (xmin, ymin, xmax, ymax)
 
+
 def get_vpr_grid_extent(conn):
     """
     Returns a tuple with (xmin, ymin, xmax, ymax) which defines
@@ -94,53 +169,8 @@ def get_vpr_grid_extent(conn):
 
     return (xmin, ymin, xmax, ymax)
 
-def initialize_one_to_one_map(conn):
-    """
-    Initializes a one to one mapping of grid coordinates. The grid must be
-    imported to the database before calling this function.
-    """
-
-    cursor = conn.cursor()
-
-    # Clear the table (just in case)
-    cursor.executescript("DELETE FROM grid_loc_map; VACUUM;")
-
-    # Get the physical grid extent
-    extent = get_phy_grid_extent(conn)
-
-    # Make a one-to-one map
-    for x, y in itertools.product(range(extent[0], extent[2]+1), range(extent[1], extent[3]+1)):
-        cursor.execute("INSERT INTO grid_loc_map VALUES (?, ?, ?, ?);", (x, y, x, y))
-
-    cursor.execute("COMMIT TRANSACTION;")
-    cursor.connection.commit()
-
 # =============================================================================
 
-def main():
-    """
-    When executed adds grid location map table(s) to the database and initializes
-    a "dummy" one-to-one map.
-    """
-
-    # Parse arguments
-    parser = argparse.ArgumentParser()
-    parser.add_argument("--db", type=str, required=True, help="Database file")
-
-    args = parser.parse_args()
-
-    # Open the DB
-    conn = sqlite3.Connection(args.db)
-
-    # Create table
-    create_tables(conn)
-
-    # Initialize one-to-one mapping
-    initialize_one_to_one_map(conn)
-
-    conn.close()
-
-# =============================================================================
 
 if __name__ == "__main__":
     main()
--- a/utils/lib/grid_mapping.py
+++ b/utils/lib/grid_mapping.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python3
+import sqlite3
+import argparse
+import os
+import itertools
+
+# =============================================================================
+
+class GridLocMap(object):
+    """
+    This class preforms forward (physical -> VPR) and backward (VPR -> physical)
+    grid location mapping.
+    """
+
+    def __init__(self, db_conn):
+
+        # Create a database cursor
+        self.db_cursor = db_conn.cursor()
+
+        # Get the whole table
+        grid_loc_map = self.db_cursor.execute(
+            "SELECT grid_phy_x, grid_phy_y, grid_vpr_x, grid_vpr_y FROM grid_loc_map"
+        ).fetchall()
+
+        # Build maps
+        self.fwd_loc_map = {}
+        self.bwd_loc_map = {}
+
+        for loc_pair in grid_loc_map:
+            phy_loc = loc_pair[0:2]
+            vpr_loc = loc_pair[2:4]
+
+            if phy_loc not in self.fwd_loc_map.keys():
+                self.fwd_loc_map[phy_loc] = []
+            self.fwd_loc_map[phy_loc].append(vpr_loc)
+
+            if vpr_loc not in self.bwd_loc_map.keys():
+                self.bwd_loc_map[vpr_loc] = []
+            self.bwd_loc_map[vpr_loc].append(phy_loc)
+
+    def get_vpr_loc(self, grid_loc):
+        return tuple(self.fwd_loc_map[grid_loc])
+
+    def get_phy_loc(self, grid_loc):
+        return tuple(self.bwd_loc_map[grid_loc])
+
+# =============================================================================
+
+def create_tables(conn):
+    """
+    Creates database tables related to grid location mappings
+    """
+
+    sql_file = os.path.join(os.path.dirname(__file__), "grid_mapping.sql")
+
+    with open(sql_file, "r") as fp:
+        cursor = conn.cursor()
+        cursor.executescript(fp.read())
+        conn.commit()
+
+def get_phy_grid_extent(conn):
+    """
+    Returns a tuple with (xmin, ymin, xmax, ymax) which defines
+    the grid extent.
+    """
+
+    # Get coordinates of all tiles
+    cursor = conn.cursor()
+    coords = list(cursor.execute("SELECT grid_x, grid_y FROM phy_tile"))
+
+    # Determine extent
+    xmin = min([loc[0] for loc in coords])
+    xmax = max([loc[0] for loc in coords])
+    ymin = min([loc[1] for loc in coords])
+    ymax = max([loc[1] for loc in coords])
+
+    return (xmin, ymin, xmax, ymax)
+
+def get_vpr_grid_extent(conn):
+    """
+    Returns a tuple with (xmin, ymin, xmax, ymax) which defines
+    the grid extent.
+    """
+
+    # Get coordinates of all tiles
+    cursor = conn.cursor()
+    coords = list(cursor.execute("SELECT grid_x, grid_y FROM tile"))
+
+    # Determine extent
+    xmin = min([loc[0] for loc in coords])
+    xmax = max([loc[0] for loc in coords])
+    ymin = min([loc[1] for loc in coords])
+    ymax = max([loc[1] for loc in coords])
+
+    return (xmin, ymin, xmax, ymax)
+
+def initialize_one_to_one_map(conn):
+    """
+    Initializes a one to one mapping of grid coordinates. The grid must be
+    imported to the database before calling this function.
+    """
+
+    cursor = conn.cursor()
+
+    # Clear the table (just in case)
+    cursor.executescript("DELETE FROM grid_loc_map; VACUUM;")
+
+    # Get the physical grid extent
+    extent = get_phy_grid_extent(conn)
+
+    # Make a one-to-one map
+    for x, y in itertools.product(range(extent[0], extent[2]+1), range(extent[1], extent[3]+1)):
+        cursor.execute("INSERT INTO grid_loc_map VALUES (?, ?, ?, ?);", (x, y, x, y))
+
+    cursor.execute("COMMIT TRANSACTION;")
+    cursor.connection.commit()
+
+# =============================================================================
+
+def main():
+    """
+    When executed adds grid location map table(s) to the database and initializes
+    a "dummy" one-to-one map.
+    """
+
+    # Parse arguments
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--db", type=str, required=True, help="Database file")
+
+    args = parser.parse_args()
+
+    # Open the DB
+    conn = sqlite3.Connection(args.db)
+
+    # Create table
+    create_tables(conn)
+
+    # Initialize one-to-one mapping
+    initialize_one_to_one_map(conn)
+
+    conn.close()
+
+# =============================================================================
+
+if __name__ == "__main__":
+    main()
+
--- a/utils/lib/grid_mapping.py
+++ b/utils/lib/grid_mapping.py
@@ -31,8 +31,8 @@ class GridLocMap(object):
         bwd_loc_map = {}
 
         # Make a one-to-one map
-        for x, y in itertools.product(range(xmin, xmax + 1),
-                                      range(ymin, ymax + 1)):
+        for x, y in itertools.product(range(xmin, xmax + 1), range(ymin,
+                                                                   ymax + 1)):
 
             fwd_loc_map[(x, y)] = [(x, y)]
             bwd_loc_map[(x, y)] = [(x, y)]
@@ -58,8 +58,8 @@ class GridLocMap(object):
         bwd_loc_map = {}
 
         # Make a one-to-one map
-        for x, y in itertools.product(range(xmin, xmax + 1),
-                                      range(ymin, ymax + 1)):
+        for x, y in itertools.product(range(xmin, xmax + 1), range(ymin,
+                                                                   ymax + 1)):
 
             phy_loc = (x, y)
             vpr_loc = (x + shift_x, y + shift_y)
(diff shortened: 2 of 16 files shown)


utils/lib/connection_database.sql
```diff
@@ -51,7 +51,17 @@ CREATE TABLE site_type(
   name TEXT
 );
 
--- Tile table, contains type and name of tile and location in grid.
+-- Tile table, contains type and name of tile and location in the physical grid.
+CREATE TABLE phy_tile(
+  pkey INTEGER PRIMARY KEY,
+  name TEXT,
+  tile_type_pkey INT,
```
litghost: I just want to talk through the mapping.   Upon a tile split, one physical location (phy_tile.pkey) will be presented by multiple VPR tiles (tile.pkey).  Therefore the tile table should have the foreign key, rather than phy_tile?  Yes?

utils/lib/grid_mapping.sql
```diff
@@ -0,0 +1,12 @@
+-- Grid location map
+-- This table maps locations from the input (physical) grid to the VPR grid
+-- locations and vice versa. If there are multiple entries for the same physical
+-- location (grid_phy_x, grid_phy_y) then it means that a single physical tile
```
litghost: Why not make this table using primary keys?  Denormalization exists for performance, I'm not sure it is required here.

xc7/utils/prjxray_arch_import.py
```diff
@@ -166,27 +182,37 @@ def main():
...
+        vpr_loc = GridLoc(vpr_loc[0], vpr_loc[1])
+
         if tile in synth_tiles['tiles']:
             synth_tile = synth_tiles['tiles'][tile]
 
             assert len(synth_tile['pins']) == 1
 
             vpr_tile_type = synth_tile_map[synth_tile['pins'][0]['port_type']]
-        elif only_emit_roi and not roi.tile_in_roi(loc):
+        elif only_emit_roi and not roi.tile_in_roi(vpr_loc):
```
litghost: I think this is a case where we want to map from VPR location to physical location, and leave the ROI definition alone?  Thoughts?

xc7/utils/prjxray_assign_tile_pin_direction.py
```diff
@@ -230,45 +227,41 @@ def handle_edges_to_channels(conn, null_tile_wires, edge_assignments, channel_wi
...
-        for wire in type_obj.get_wires():
-            wires_in_tile_types.add((tile_type, wire))
+        c = conn.cursor()
 
-        for site in type_obj.get_sites():
-            for site_pin in site.site_pins:
-                if site_pin.wire is None:
-                    continue
+        # List tile wires and site wires
+        tiles = c.execute("SELECT pkey, name FROM tile_type").fetchall()
```
litghost: No need for the fetchall, you can just do for x in c.execute() and stream the results.  The inner for loop just needs to use a difference Cursor object.  This is recommended because it will stream the results from the database.

xc7/utils/prjxray_create_edges.py
```diff
@@ -450,9 +448,8 @@ def find_connector(wire_pkey, node_pkey):
 
     return find_connector
 
-def make_connection(conn, input_only_nodes, output_only_nodes,
-        find_wire, find_pip, find_connector,
-        tile_name, loc, tile_type, pip, switch_pkey):
+
+def make_connection(conn, input_only_nodes, output_only_nodes, loc, tile_pkey, src_wire_pkey, dst_wire_pkey, pip_pkey, switch_pkey, find_connector):
```
litghost: Why did this get reverted?  The previous format was conforming to the max line length of 80.

xc7/utils/prjxray_create_edges.py
```diff
@@ -463,14 +460,14 @@ def make_connection(conn, input_only_nodes, output_only_nodes,
...
-        find_wire (function): Return value from create_find_wire.
-        find_pip (function): Return value from create_find_pip.
-        find_connector (function): Return value from create_find_connector.
-        tile_name (str): Name of tile pip belongs too.
-        loc (prjxray.grid_types.GridLoc): Location of tile.
-        pip (prjxray.tile.Pip): Pip being connected.
+        tile_pkey (int): pkey of the tile in the tile table
+        src_wire_pkey (int): pkey of the source wire in the wire table
+        dst_wire_pkey (int): pkey of the destination wire in the wire table
+        loc (tuple): Location of tile.
```
litghost: Whenever we pass the location of a tile, lets be specific about physical or VPR coordinates for clarity.

xc7/utils/prjxray_create_edges.py
```diff
@@ -485,35 +482,32 @@ def make_connection(conn, input_only_nodes, output_only_nodes,
 
     """
 
-    src_wire_pkey, tile_pkey, src_node_pkey = find_wire(tile_name, tile_type, pip.net_from)
-    sink_wire_pkey, tile_pkey2, sink_node_pkey = find_wire(tile_name, tile_type, pip.net_to)
+    c = conn.cursor()
 
-    assert tile_pkey == tile_pkey2
+    # Get node pkeys
+    src_node_pkey = c.execute("SELECT node_pkey FROM wire WHERE tile_pkey = (?) AND wire_in_tile_pkey = (?)", (tile_pkey, src_wire_pkey)).fetchone()[0]
```
litghost: Probably worth making a utility method for "get_node_from_wire"

xc7/utils/prjxray_create_edges.py
```diff
@@ -822,37 +815,34 @@ def main():
...
-        for loc in progressbar.progressbar(grid.tile_locations()):
-            gridinfo = grid.gridinfo_at_loc(loc)
-            tile_name = grid.tilename_at_loc(loc)
+        c2 = conn.cursor()
+        for tile_pkey, tile_type_pkey, grid_x, grid_y in progressbar.progressbar(c.execute("SELECT pkey, tile_type_pkey, grid_x, grid_y FROM tile")):
+            loc = (grid_x, grid_y)
 
             # Not a synth node, check if in ROI.
-            if use_roi and not roi.tile_in_roi(loc):
+            if use_roi and not (roi_range[0] <= loc[0] and loc[0] <= roi_range[2] and roi_range[1] <= loc[1] and loc[1] <= roi_range[3]):
```
litghost: Again, for the ROI, probably best to iterate over VPR tiles, but map back to phy coordinates and check against the phy ROI

xc7/utils/prjxray_create_edges.py
```diff
@@ -822,37 +815,34 @@ def main():
...
-            for pip in tile_type.get_pips():
-                if pip.is_pseudo:
-                    continue
-
-                if not pip.is_directional:
-                    # TODO: Handle bidirectional pips?
-                    continue
+                # No pseudo pips and bidirectional pips should be present here
+                # as they were skipped during import to the SQLite database
+                # in prjxray_form_channels.py
```
litghost: I recommend we add the pseudo and bidirectional column's to the database and filter them here still.

xc7/utils/prjxray_form_channels.py
```diff
@@ -139,15 +141,10 @@ def build_tile_type_indicies(c):
     c.execute("CREATE INDEX site_pin_index ON site_pin(name, site_type_pkey);")
     c.execute("CREATE INDEX wire_name_index ON wire_in_tile(name, tile_type_pkey);")
     c.execute("CREATE INDEX wire_site_pin_index ON wire_in_tile(site_pin_pkey);")
-    c.execute("CREATE INDEX tile_type_index ON tile(tile_type_pkey);")
```
litghost: Why removed these indices?

utils/lib/connection_database.sql
mkurc-ant: It works like that: First the prjxray grid is imported to the `phy_tile` table. Then the mapping physical <-> VPR is generated. Once done, entires from the `phy_tile` table are "copied" to the `tile` table with modified locations and the same pkey (actually didn't think if the pkey needs to stay the same).

The correspondence between these two tile tables is held in the `grid_loc_map` table. It is done by keeping location map rather than direct `tile` <-> `phy_tile` via foreign key.

xc7/utils/prjxray_create_edges.py
mkurc-ant: I didn't know that I should keep line lengths < 80. I will fix that.

xc7/utils/prjxray_create_edges.py
mkurc-ant: You mean changing the table name from `tile` to eg. `vpr_tile` ?

xc7/utils/prjxray_create_edges.py
mkurc-ant: I think this should be done in a separate PR.

xc7/utils/prjxray_form_channels.py
mkurc-ant: These indices were created for the `tile` table right after grid import. Now the grid is imported to the `phy_tile` table which is not used anywhere else but during the location mapping. Once all entries are copied to the `tile` table those indices that you are referring to are created. See lines 745-747 in this file.

utils/lib/grid_mapping.sql
mkurc-ant: You mean not to store coordinates but pkeys (as foreign keys) of `phy_tile` and `tile` ?

mkurc-ant:
@litghost So regarding the ROI: I will make the location mapping reverse so tile locations will be first converted to physical and then compared against the ROI.

utils/lib/grid_mapping.sql
litghost: Yes.

xc7/utils/prjxray_create_edges.py
litghost: We just add back on auto-formatting with the target "make python_format", and the CI is checking formatting for green.

xc7/utils/prjxray_create_edges.py
litghost: No, I just mean be specific when passing a location as a function argument.

xc7/utils/prjxray_form_channels.py
litghost: Okay

xc7/utils/prjxray_create_edges.py
litghost: Probably worth doing that prior to merge this one then.  Should be a small PR.

utils/lib/grid_mapping.sql
litghost: Every tile and phy_tile should appear in this map at least once.  That doesn't change my conclusion about the coordinates vs foreign keys.

utils/lib/connection_database.sql
mkurc-ant: So the conclusion is that I will go with the foreign keys.

xc7/utils/prjxray_arch_import.py
mkurc-ant: I already changed that in `prjxray_create_synth_tiles.py`. I will do the same for this script.

xc7/utils/prjxray_create_edges.py
mkurc-ant: I did the formatting.

xc7/utils/prjxray_create_synth_tiles.py
```diff
@@ -106,44 +151,42 @@ def main():
...
-    vbrk_loc = None
-    vbrk_tile = None
-    vbrk2_loc = None
-    vbrk2_tile = None
-    for tile in g.tiles():
-        if tile in vbrk_in_use:
-            continue
+    # Find two VBRK's in the corner of the fabric to use as the synthetic VCC
+    loc_min = GridLoc(g.dims()[0], g.dims()[2])
+    loc_max = GridLoc(g.dims()[1], g.dims()[3])
```
litghost: Keep them to the left side of the grid, e.g. use min x dim for both.

mkurc-ant: Done.

xc7/make/project_xray.cmake
```diff
@@ -199,26 +207,26 @@ function(PROJECT_XRAY_ARCH)
...
+      --connection_database ${GENERIC_CHANNELS_LOCATION}
       --output-arch ${CMAKE_CURRENT_BINARY_DIR}/arch.xml
       --tile-types "${TILE_TYPES_COMMA}"
       --pin_assignments ${PIN_ASSIGNMENTS}
       --device ${DEVICE}
       ${ROI_ARG}
     DEPENDS
     ${ARCH_IMPORT}
-    ${DEPS}
+    ${DEPS} ${CMAKE_CURRENT_BINARY_DIR}/channels.db
```
litghost: Never depend directly on generated files, always use append_file_dependency, otherwise parallel brings may break.

mkurc-ant: Fixed.

xc7/utils/prjxray_assign_tile_pin_direction.py
mkurc-ant: Yes, but no here. The cursor "c" is used inside the loop so it will break the iteration (I did such mistakes before). I will create a second cursor eg. "c2" for iteration.

xc7/utils/prjxray_create_edges.py
mkurc-ant: Ok.

mkurc-ant:
Clarification: As for `6272c276` the code shifts the whole grid by 2 to the east. And the VPR fails on routing with the message while implementing the `xc7/tests/counter` :

`
Cannot route from BLK_SY-GND.GND[0] (RR node: 37040 type: SOURCE location: (20,51) class: 0) to BLK_TI-CLBLL_R.CLBLL_LL_DX[0] (RR node: 136753 type: SINK location: (48,19) class: 29) -- no possible path
Failed to route connection from '$false' to '$auto$alumacc.cc:474:replace_alu$6.slice[16].carry' for net '$false' (#2)
`

mkurc-ant:
I have updated my locally built VPR to `1cf4f1b9b` (I've been using an "old" one). And now routing of the `xc7/tests/counter` succeeds. I can shift the grid toward the X direction with no error.

The problem was related to: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/520

litghost:
CI is red on travis, ice40 and xc7.

utils/lib/grid_mapping.sql
```diff
@@ -0,0 +1,8 @@
+-- Grid location map
+CREATE TABLE IF NOT EXISTS grid_loc_map(
+    phy_tile_pkey INT,
+    vpr_tile_pkey INT,
+    FOREIGN KEY(phy_tile_pkey) REFERENCES phy_tile(pkey),
+    FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey)
+);
```
litghost: Why not just have this in utils/lib/connection_database.sql?

utils/lib/grid_mapping.py
```diff
@@ -0,0 +1,179 @@
...
+# =============================================================================
+
+
+class GridLocMap(object):
+    """
+    This class preforms forward (physical -> VPR) and backward (VPR -> physical)
+    grid location mapping.
+    """
+
+    def __init__(self, fwd_loc_map, bwd_loc_map):
```
litghost: Should this just take a connection to the database?

utils/lib/grid_mapping.py
```diff
@@ -0,0 +1,179 @@
...
+
+            if vpr_loc not in bwd_loc_map.keys():
+                bwd_loc_map[vpr_loc] = []
+            bwd_loc_map[vpr_loc].append(phy_loc)
+
+        # Return map object
+        return GridLocMap(fwd_loc_map, bwd_loc_map)
+
+    def get_vpr_loc(self, grid_loc):
+        return tuple(self.fwd_loc_map[grid_loc])
```
litghost: tuple call here seems unnecessary?

utils/lib/grid_mapping.py
```diff
@@ -0,0 +1,179 @@
...
+            vpr_loc = (x + shift_x, y + shift_y)
+
+            fwd_loc_map[phy_loc] = [vpr_loc]
+            bwd_loc_map[vpr_loc] = [phy_loc]
+
+        # Return map object
+        return GridLocMap(fwd_loc_map, bwd_loc_map)
+
+    @staticmethod
+    def load_from_database(conn):
```
litghost: Should have a "insert_into_database"?  I assume this object will eventually gain the other grid manipulation features?

utils/lib/grid_mapping.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        Loads grid location mapping from a SQL database. Returns a GridLocMap
+        object.
+        :param conn:
```
litghost: Please follow the following style guide: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods

utils/lib/grid_mapping.py
```diff
@@ -0,0 +1,179 @@
...
+        return tuple(self.fwd_loc_map[grid_loc])
+
+    def get_phy_loc(self, grid_loc):
+        return tuple(self.bwd_loc_map[grid_loc])
+
+
+# =============================================================================
+
+
+def create_tables(conn):
```
litghost: Ya, please delete this function and just move the contents of grid_mapping.sql to the main DB schema file

xc7/utils/prjxray_create_edges.py
```diff
@@ -1031,39 +1046,37 @@ def main():
...
-                    loc=loc,
-                    tile_type=gridinfo.tile_type,
-                    pip=pip,
-                    switch_pkey=switch_pkey,
-                    delayless_switch_pkey=delayless_switch_pkey,
-                    const_connectors=const_connectors
+                    conn, input_only_nodes, output_only_nodes, loc, tile_pkey,
+                    pip_src_wire_pkey, pip_dst_wire_pkey, pip_pkey,
+                    switch_pkey, delayless_switch_pkey, find_connector,
+                    const_connectors
```
litghost: Please continue to use kwarg calling style when there are this many arguments.

xc7/utils/prjxray_create_edges.py
litghost: This comment still stands

xc7/utils/prjxray_create_synth_tiles.py
```diff
@@ -3,12 +3,66 @@
...
+from prjxray_db_cache import DatabaseCache
+from lib.grid_mapping import GridLocMap
+
+
+def find_vbrk_closest_to(grid, roi, loc, loc_in_use):
+    """
+    Finds a VBRK tile (optionally within the ROI) which is located closest
+    to a given location. Checks if such a tile is not already used by
+    another synth tile.
+    :param grid:
```
litghost: Comment style again

xc7/utils/prjxray_form_channels.py
```diff
@@ -207,38 +203,36 @@ def import_grid(db, grid, conn):
...
-        tile_type = db.get_tile_type(gridinfo.tile_type)
 
-        c.execute(
-            """SELECT pkey, tile_type_pkey FROM tile WHERE name = ?;""",
-            (tile, )
-        )
-        tile_pkey, tile_type_pkey = c.fetchone()
-        tiles[tile] = (tile_pkey, tile_type_pkey)
+    for tile_name, tile_pkey, tile_type_pkey in progressbar.progressbar(
+            c1.execute("SELECT name, pkey, tile_type_pkey FROM tile")):
```
litghost: This should be phy_tile?

litghost:
I think this PR is looking pretty good.  Please address comments and make CI green, and I think we can merge.

mkurc-ant:
Actually the CI failed because of `"CondaHTTPError: HTTP 502 BAD GATEWAY for url <https://conda.anaconda.org/symbiflow/noarch/repodata.json>"`

I hope it will pass once I push fixes related to  comments.

utils/lib/grid_mapping.py
mkurc-ant: There is a static method `load_from_database` which loads the grid map from database and returns an instance of the `GridLocMap` class.

utils/lib/grid_mapping.py
mkurc-ant: Since the grid location mapping is implemented as a table with key pairs that connect two other tables with tiles, it is not possible to just store the mapping in the database. The mapped tile grid needs to be created first and this is done in function `remap_tile_grid` in `prjxray_form_channels.py`

utils/lib/grid_mapping.py
mkurc-ant: Ok.

xc7/utils/prjxray_create_edges.py
mkurc-ant: Done.

xc7/utils/prjxray_form_channels.py
mkurc-ant: Actually when there is no tile split yet it does not matter. I still haven't figured out though what to do with all connections when CLBs are split.

xc7/utils/prjxray_create_edges.py
mkurc-ant: I'll look into it.

xc7/utils/prjxray_create_synth_tiles.py
mkurc-ant: Fixed.

utils/lib/grid_mapping.sql
mkurc-ant: At first I thought there will be more tables related to the grid map. But yest it can be moved to the `utils/lib/connection_database.sql` file.

utils/lib/grid_mapping.py
mkurc-ant: Done.

utils/lib/grid_mapping.py
mkurc-ant: Right, fixed.

utils/lib/grid_mapping.py
```diff
@@ -0,0 +1,179 @@
...
+        grid_loc_map = c.execute(
+            """
+SELECT phy.grid_x, phy.grid_y, vpr.grid_x, vpr.grid_y
+FROM phy_tile phy
+INNER JOIN grid_loc_map map
+ON phy.pkey = map.phy_tile_pkey
+INNER JOIN tile vpr
+ON vpr.pkey = map.vpr_tile_pkey
+"""
+        ).fetchall()
```
litghost: Should iterate over cursor rather than using fetch all.

xc7/utils/prjxray_create_edges.py
```diff
@@ -1031,39 +1046,45 @@ def main():
...
-            tile_name = grid.tilename_at_loc(loc)
+        c2 = conn.cursor()
+        for tile_pkey, tile_type_pkey, grid_x, grid_y in \
+                progressbar.progressbar(
+                c.execute(
+                    "SELECT pkey, tile_type_pkey, grid_x, grid_y FROM tile")):
+            loc = (grid_x, grid_y)
 
             # Not a synth node, check if in ROI.
-            if use_roi and not roi.tile_in_roi(loc):
```
litghost: Why implement this comparison vs using the ROI object?  They should be equivilant?

litghost: Why is this using VPR coordinates?  The ROI is defined using prjxray coordinates?  This should iterate over VPR tiles, and get the prjxray coordinates for the ROI check.

xc7/utils/prjxray_form_channels.py
```diff
@@ -207,38 +203,36 @@ def import_grid(db, grid, conn):
...
         c.execute(
             """
-INSERT INTO tile(name, tile_type_pkey, grid_x, grid_y)
+INSERT INTO phy_tile(name, tile_type_pkey, grid_x, grid_y)
 VALUES
   (?, ?, ?, ?)""",
             (tile, tile_types[gridinfo.tile_type], loc.grid_x, loc.grid_y)
         )
 
-    build_other_indicies(c)
```
litghost: Aren't we missing the index on phy_tile(name) and phy_tile(grid_x, grid_y)?

xc7/utils/prjxray_form_channels.py
litghost: My comment absolutely stands.  Connections using the prjxray DB will always use the physical grid, never the VPR grid.

litghost: As a concrete example, on line 235, the tile_type must be the prjxray tile type name, not the VPR tile type name.  SLICEL is not a tile type, CLBLL_L is.

xc7/utils/prjxray_create_edges.py
mkurc-ant: That is another possible solution. However, the `prjxray_create_edges.py` does not get the roi directly. Instead it reads `synth_tiles.json` which contains the ROI range and synthetic tiles in VPR coordinates (already mapped). So it is easier to do a simple check whether a VPR location fits within the remapped ROI range read from `synth_tiles.json` rather than having the physical (prjxray) ROI definition as another input of the script. Both of them contain the same information.

xc7/utils/prjxray_form_channels.py
mkurc-ant: I will add them. But they are barely needed as the phy_tile table is accessed only once inside the `remap_tile_grid` function.

utils/lib/grid_mapping.py
mkurc-ant: Right, done.

xc7/utils/prjxray_form_channels.py
mkurc-ant: Ok, you are right. This has to be a phy_tile. In this function tile location isn't concerned, just the type.

xc7/utils/prjxray_create_edges.py
litghost: > Both of them contain the same information.

This is not generally speaking true.  The physical ROI limit has a 1:1 relationship to the underlying bitstream, the same does not have to be true for the remapped ROI.  Do not make the assumption that they are the same.

litghost:
CI is red:

```
/ |#                                                  | 0 Elapsed Time: 0:00:00Loading database from '/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/xc7/archs/artix7/channels.db'
2019-04-16 04:51:53.425892 Processing direct connections.
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc7/utils/prjxray_assign_tile_pin_direction.py", line 499, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc7/utils/prjxray_assign_tile_pin_direction.py", line 332, in main
    conn, direct_connections, edge_assignments
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc7/utils/prjxray_assign_tile_pin_direction.py", line 141, in handle_direction_connections
    source_wire)].append((source_dir, ))
KeyError: ('CLBLL_L', 'CLBLM_M_COUT')

                                                                               

| | #                                               | 284 Elapsed Time: 0:00:00
xc7/archs/artix7/CMakeFiles/file_xc7_archs_artix7_pin_assignments.json.dir/build.make:246: recipe for target 'xc7/archs/artix7/pin_assignments.json' failed
```

mkurc-ant:
Yeah :( Working on it...

xc7/utils/prjxray_create_edges.py
mkurc-ant: What I mean is that `synth_tiles.json` contains the ROI which has been re-mapped from physical to VPR coordinates. So instead of passing the ROI from prjxray db to `prjxray_create_edges.py` (in physical coordinates) it is possible to read the roi in VPR coordinates (this is already done through reading `synth_tiles.json`) and re-map it back to physical coordinates. The check can be performed then. I never meant to assume 1:1 relation.

mkurc-ant: @litghost I can add a "--roi" argument if you want and use it to pass the physical ROI from the prjxray. But it will only complicate the whole flow (which is in my opinion complex enough).

mkurc-ant:
@litghost I rebased with master. The CI now fails at:

> Traceback (most recent call last):
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc7/fasm2bels/__main__.py", line 4, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc7/fasm2bels/fasm2bels.py", line 239, in main
    net_map = load_net_list(conn, args.rr_graph, args.route_file)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc7/fasm2bels/fasm2bels.py", line 155, in load_net_list
    read_xml_file(rr_graph_file), need_edges=False
TypeError: __init__() got an unexpected keyword argument 'need_edges'

I'll check master but I don't think that the failure is related to this PR.

mkurc-ant:
Yep, and it fails on master as well...

xc7/utils/prjxray_create_edges.py
mkurc-ant: @litghost I've created a PR with the change you suggested: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/604 Once merged it will require integration with this one (I marked the place in code as FIXME).

utils/lib/grid_mapping.sql
mkurc-ant: So in the end there is a table with pkey pairs.

xc7/utils/prjxray_create_edges.py
mkurc-ant: As you suggested I added a method "get_node_from_wire"

xc7/utils/prjxray_create_edges.py
mkurc-ant: @litghost Another solution that comes into my mind is to simply store physical ROI range in `synth_tiles.json`.

litghost: That was what I was suggesting

PR #538, opened 2019-04-05
Fix multi-process bugs in CMake build system.

litghost:
- Add lock files for conda and pip installations to prevent those tools
   from being invoked in parallel.
 - Change some file dependencies to target dependencies, which is
   important when multiple process builds occur.
 - Remove eblif and bit files from ALL, instead only having the conda environment on all.

Fixes #448, fixes #337

PR #540, opened 2019-04-05
Add 7-series rrgraph to CI

litghost:
Remaining tasks:
- [x] Wait for https://github.com/SymbiFlow/symbiflow-arch-defs/pull/538 to be merged
- [ ] Determine "-j<N>" level that enables acceptable run time.  Adding 7-series rrgraph import to CI will add ~30 minutes, which is likely too much


litghost:
Current runtime before this PR: 37 min 27 sec

elmsfu:
The [default tier](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) for travis is 2 CPUs, so not sure we will get much for higher than `-j 2`. see [travis makefile optimization](https://docs.travis-ci.com/user/speeding-up-the-build/#makefile-optimization)

litghost:
> The [default tier](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) for travis is 2 CPUs, so not sure we will get much for higher than `-j 2`. see [travis makefile optimization](https://docs.travis-ci.com/user/speeding-up-the-build/#makefile-optimization)

I think trying for -j2 on the longer portions of the build is worth trying, under the assumption that the current CI run completes successfully.

litghost:
Time limit 50 minutes, closing this.

elmsfu:
I think that is per Travis job. We could look at splitting it up in different ways like a travis job per architecture. But it's probably not worth our time and should focus on running on a larger build server (ie kokoro).

PR #543, opened 2019-04-08
Optimize routing resources on unused signals.

litghost:
Unused reset signals are tied to VCC by default, which would cause
these reset signals to be tied high.  Use the inverter to tie these signals
low with no additional routing.


mithro:
LGTM!

Issue #539, opened 2019-04-05
Define and implement reasonable all target


litghost:
Current default all target as of c66597570807a07734e59a98fe0456b2b2a39ff0 is to build every bitstream.  This is almost certainly not the right thing to do as the default all target.  This issue about defining and implementing a reasonable all target.

elmsfu:
I agree that all bitstreams is probably not correct. Maybe "small" tests would be a good goal with the aim to cover of all the blocks and modes. But that leaves any stress test, but not sure if travis is a good match for that.

litghost:
I think it problably time to get the kokoro instances going to symbiflow, because they are beefier machines, and will be able to run @ -j16 and get results in a reasonable amount of time.

PR #547, opened 2019-04-08
Adds route grouping targets and all targets for major variants.

litghost:
@mithro and @elmsfu Does this address https://github.com/SymbiFlow/symbiflow-arch-defs/issues/539 ?

This doesn't tie them into the true all target, but I don't think we want to.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,3 +123,72 @@ add_custom_command(
   TARGET format_python
   COMMAND ${symbiflow-arch-defs_SOURCE_DIR}/common/yapf_helper.sh -f -y ${YAPF}
   )
+
+
+add_custom_target(all_ice40_demos)
+add_dependencies(all_ice40_demos
+    all_icevision_bin
+    all_icestick_bin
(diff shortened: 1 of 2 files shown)


elmsfu:
> @mithro and @elmsfu Does this address #539 ?
> 
> This doesn't tie them into the true all target, but I don't think we want to.

Did you mean to merge this? You only requested a review about an hour ago.

I do not think this addresses #539. We discuss what the all target should be on that issue and then implement it.

CMakeLists.txt
```diff
@@ -123,3 +123,72 @@ add_custom_command(
...
+
+
+if (NOT DEFINED ENV{CI} OR NOT $ENV{CI})
+add_dependencies(all_7series_route_tests
+    all_dummy_artix7_xc7a50t-arty_test_route
+    all_dummy_artix7_xc7a50t-basys3_test_route
+    )
+endif()
+
+add_custom_target(all_7series_demos)
```
mithro: To be consistent with naming - should `7series` be `xc7`?

Issue #550, opened 2019-04-08
[XC7] Picosoc requires tile split


litghost:
Round robin packing is disabled for now, so until the tile split (#415) is complete, picosoc will not place due to resource limitations.

Issue #549, opened 2019-04-08
[XC7] RAM128X1D test is no longer working


litghost:

HackerFoo:
```
Results from re-enabling the dram_1_128x1d FPGA target and running _bin:

# Packing
Begin packing '/home/dusty/src/symbiflow-arch-defs/build/xc7/tests/dram/dram_1_128x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.eblif'.

After removing unused inputs...
	total blocks: 26, total nets: 25, total inputs: 17, total outputs: 2
Begin prepacking.
Finish prepacking.
Using inter-cluster delay: 6.74304e-10
Packing with pin utilization targets: BLK_TI-CLBLL_R:0.8,1 BLK_TI-CLBLL_L:1,1 BLK_TI-CLBLM_R:1,1 BLK_TI-CLBLM_L:1,1 BLK_TI-BRAM_L:1,1 BLK_SY-INPAD:1,1 BLK_SY-OUTPAD:1,1 BLK_SY-VCC:1,1 BLK_SY-GND:1,1
Not enough resources expand FPGA size to (116 x 158)
Complex block 0: ram0.ram3, type: BLK_TI-CLBLM_R ....
Complex block 1: ram0.f7a_mux, type: BLK_TI-CLBLM_L .Error 1: 
Type: Packing
File: /build/source/vpr/src/pack/cluster.cpp
Line: 1976
Message: Can not find any logic block that can implement molecule.
	Pattern DRAM128_DP ram0.f7a_mux


Routing was impossible!
Net 'ram0.blut_o6' is impossible to route within proposed BLK_TI-CLBLM_L cluster
Failed route at end, repack cluster trying detailed routing at each stage.
Routing was impossible!
Net 'ram0.blut_o6' is impossible to route within proposed BLK_TI-CLBLM_L cluster
Routing was impossible!
Net 'ram0.blut_o6' is impossible to route within proposed BLK_TI-CLBLM_R cluster

```

mithro:
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/6850f2c665a8fabb12a771067d6d2a2d7ea266c0/xc7/primitives/slicem/slicem.pb_type.xml#L569-L574

mithro:
[...]

PR #551, opened 2019-04-08
Disable targets that are broken to get initial green CI.

litghost:
The intention is to get kokoro green with the targets that are working.  3 issues have been filed to track the targets disabled in this PR.

Issues filed #548, #549, #550


litghost:
> Why is test_dram_packing not on CI?
> 
> Otherwise LGTM?

CI really means travis, which cannot run 7-series or large ice40 parts.

mithro:
Then maybe should check for `TRAVIS=true` - https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

PR #556, opened 2019-04-08
Enable kokoro on symbiflow-arch-defs

litghost:
Example kokoro run: https://github.com/SymbiFlow/symbiflow-arch-defs-kokoro/pull/4


.github/kokoro/ice40.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+CALLED=$_
```
elmsfu: @mithro I think I'm missing the point of this. Why do we set this and not use it? Is this from a recipe/example?

.github/kokoro/ice40.sh
```diff
@@ -0,0 +1,27 @@
...
+[[ "${BASH_SOURCE[0]}" != "${0}" ]] && SOURCED=1 || SOURCED=0
```
elmsfu: nit: Prefer `if` block over short-circuit assignment. (Maybe also from recipe or example?)

mithro: I think it comes from https://stackoverflow.com/questions/2683279/how-to-detect-if-a-script-is-being-sourced
> ```shell
> [[ $_ != $0 ]] && echo "Script is being sourced" || echo "Script is a subshell"
> ```

litghost: It's unclear if we need the SOURCED variable at all.  Should we just remove this?

.github/kokoro/ctest2junit.xsl
```diff
@@ -0,0 +1,120 @@
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
```
elmsfu: Forgive my ignorance, why do we need this file?

.github/kokoro/kokoro-cfg.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
```
elmsfu: Let's add a comment for what purpose this script serves. Looks to generate the continuous and presubmit configs from a template? A helper/developer script?

.github/kokoro/kokoro-cfg.py
```diff
@@ -0,0 +1,46 @@
...
+
+db_full = """\
+# Format: //devtools/kokoro/config/proto/build.proto
+
```
elmsfu: Maybe add a comment that it is generated from this script.

.github/kokoro/steps/git.sh
```diff
@@ -0,0 +1,27 @@
...
+echo "========================================"
+echo "Git log"
+echo "----------------------------------------"
+echo "----------------------------------------"
+
+echo
+echo "========================================"
+echo "Git fetching tags"
+echo "----------------------------------------"
+git fetch --tags || true
```
elmsfu: Are these `|| true` so that kokoro continues even if the commands fail? Maybe comment worthy that we want to continue even if the commands fail.

.github/kokoro/ctest2junit.xsl
litghost: Removed.

.github/kokoro/database.cfg
```diff
@@ -0,0 +1,14 @@
+build_file: "symbiflow-prjxray-db/.github/kokoro/database.sh"
```
mithro: This file shouldn't exist for symbiflow-arch-defs (bad copy from prjxray).

litghost: Removed

.github/kokoro/steps/git.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -e
+
+echo
+echo "========================================"
+echo "Git log"
+echo "----------------------------------------"
```
mithro:
```suggestion
echo "----------------------------------------"
git log -n5 --stat
```

PR #557, opened 2019-04-08
Add FF SR/CE signal test.

litghost:
This is the beginning of an equivilance test for FF SR and CE control
signals.

### DIFF
--- a/xc7/tests/ff_sr_ce/ff_ce_sr.v
+++ b/xc7/tests/ff_sr_ce/ff_ce_sr.v
@@ -1,0 +1,61 @@
+module top(
+    input clk,
+    input rx,
+    output tx,
+    input [15:0] sw,
+    output [15:0] led
+);
+
+assign tx = rx;
+
+// Generate some signals for the CE and reset lines.
+wire ce = ^sw;
+wire reset = &sw;
+
+// 4 FF's with SR tied to GND, and CE tied to VCC
+// 4 FF's with SR tied to input and CE tied to VCC
+// 4 FF's with SR tied to GND and CE tied to input
+// 4 FF's with SR tied to input and CE tied to input
+generate for(i = 0; i < 3; i++) begin:ff
+    FDRE #(
+        .INIT(0),
+    ) vcc_gnd (
+        .Q(led[4*i+0]),
+        .C(clk),
+        .D(sw[4*i+0]),
+        .CE(1),
+        .R(0)
+    );
+
+    FDRE #(
+        .INIT(0),
+    ) s_gnd (
+        .Q(led[4*i+1]),
+        .C(clk),
+        .D(sw[4*i+1]),
+        .CE(ce),
+        .R(0)
+    );
+
+    FDRE #(
+        .INIT(0),
+    ) s_s (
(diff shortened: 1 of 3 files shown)


xc7/tests/ff_sr_ce/ff_ce_sr.v
```diff
@@ -0,0 +1,61 @@
...
+
+// Generate some signals for the CE and reset lines.
+wire ce = ^sw;
+wire reset = &sw;
+
+// 4 FF's with SR tied to GND, and CE tied to VCC
+// 4 FF's with SR tied to input and CE tied to VCC
+// 4 FF's with SR tied to GND and CE tied to input
+// 4 FF's with SR tied to input and CE tied to input
+generate for(i = 0; i < 3; i++) begin:ff
```
mithro: I think it might make sense to split this into 4 generate statements so a comment can go above each FF loop with the given type?

litghost: I think moving the comment makes sense, but creating 4 for loops is a terrible idea.

xc7/tests/ff_sr_ce/ff_ce_sr.v
```diff
@@ -0,0 +1,62 @@
...
+);
+
+assign tx = rx;
+
+// Generate some signals for the CE and reset lines.
+wire ce = ^sw;
+wire reset = &sw;
+
+// 4 FF's of each SR/CE varient to check packing behavior.
+generate for(i = 0; i < 3; i++) begin:ff
```
mithro: It also might be a good idea to do 3 elements to make sure that it doesn't try and pack the wrong things together?

xc7/tests/ff_sr_ce/ff_ce_sr.v
```diff
@@ -0,0 +1,62 @@
...
+    output [15:0] led
+);
+
+assign tx = rx;
+
+// Generate some signals for the CE and reset lines.
+wire ce = ^sw;
+wire reset = &sw;
+
+// 4 FF's of each SR/CE varient to check packing behavior.
```
elmsfu: How will the packing behavior be confirmed?

litghost: I'll add a "ASSERT_USAGE" once I get the equiv tests working.  This will look a little strange until the SLICEL/SLICEM split is complete.

elmsfu: Can you share a link to documentation of `ASSERT_USAGE`? I'm not familiar with it.

litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/tests/dram/CMakeLists.txt#L6

litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/usage.py

elmsfu: Thanks.

PR #560, opened 2019-04-09
Add support for naming top-level port names based on pcf file.

litghost:
This is useful for binding to test benches and equivilence miter
circuits.

### DIFF
--- a/utils/lib/parse_pcf.py
+++ b/utils/lib/parse_pcf.py
@@ -1,0 +1,28 @@
+from collections import namedtuple
+import re
+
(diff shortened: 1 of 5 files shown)


litghost:
> LGTM.
> 
> Random note: It is a bit weird to be using pcf files for Xilinx parts, they generally use xdc files for these types of constraints. pcf is a Lattice thing...

I just used PCF because that was what ice40 used.  We can always change it.  xdc files are basically tcl, and I'm not super excited to deal with that.

Issue #559, opened 2019-04-08
Rely on external package management as much as possible


HackerFoo:
Automatically loading dependencies via Conda and PIP causes difficulty when trying to integrate with the system package manager, or when working offline.

In particular, Nix (and hence NixOS) controls dependencies to ensure repeatable builds, so it would be better to allow Nix to handle the dependencies.

The apparent approach would be to use `cmake -DUSE_CONDA=FALSE ..`, but this seems to be currently broken.

mithro:
@HackerFoo - External package managers (IE Debian / Ubuntu apt, Fedora, etc) don't have new enough versions or our custom patched versions. The other option would be to compile from scratch.

conda gives us a way to distribute (rather than compile from scratch) binary versions of our tooling.

litghost:
Conda setup should only occur the first time you build, much like a system package manager.  If conda is getting invoked after initial setup, that is likely a bug.

HackerFoo:
This works okay for development, but may cause problems when deploying to e.g. test servers. It's best to leave the packaging to the package managers, and either distribute packages directly or set up repositories (e.g. PPAs for Ubuntu.) Another option is to use Nix, or use a separate script that can be run before building.

I've also seen a single build download Miniconda at least 6 times, even if it has just been run. I think it was `buttons_basys3_prog`.

litghost:
> I've also seen a single build download Miniconda at least 6 times, even if it has just been run. I think it was `buttons_basys3_prog`.

Then that's simply a bug.

elmsfu:
I would have written a [shorter reply if I had the time](https://en.wikiquote.org/wiki/Blaise_Pascal#Pascal_plus_longue)

I think there is room to improve around this area, but we are still not close to stable and rely on integration of several repos that are in flux. I don't think this is an area we should attack yet. There is probably a bigger question for deploying for non-development.

## My understanding of current (2019-04-10) state
Currently we have 4 methods that I know of for external dependencies (or at least externally based if we have a custom fork).

1. conda (vpr, yosys)
1. submodule (icestorm, netlistsvg, fasm)
1. pip (should probably wrap all python as installable modules and have finer grain deps)
1. requirements.txt (not used by build system by default)

My experience has been that most projects choose to list dependencies in documentation and maybe provide a script to run for development/ease-of-use for popular OS/distro. CI can then run that script as part of setup.

## Potential options

1. Status Quo
[...]

PR #563, opened 2019-04-09
Add a shell.nix to allow developing with Nix

HackerFoo:
`nix-shell shell.nix` and get to work.

Signed-off-by: Dusty DeWeese <dustin.deweese@gmail.com>


shell.nix
```diff
@@ -0,0 +1,99 @@
...
+      verilog
+      sqlite
+      nodejs
+      pkg-config
+      libftdi
+      fasm
+      icestorm
+      gtkwave
+
+      silver-searcher
```
mithro: What is this?

litghost:
I don't know much about nix, but aren't you missing yosys?  Also unclear why this belongs as part of the symbiflow repository directly?

shell.nix
```diff
@@ -0,0 +1,99 @@
...
+          name = "vtr-${version}";
+          version = "git";
+          src = fetchGit {
+            url = "https://github.com/SymbiFlow/vtr-verilog-to-routing.git";
+            rev = "1cf4f1b9b2ef3e815f58a770cfad7f0aa209d423";
+          };
+
+          doCheck = false;
+
+          buildInputs = [ ctags bison flex cmake ];
```
mithro: Does vtr actually require ctags?

shell.nix
HackerFoo: [silver-searcher](https://github.com/ggreer/the_silver_searcher) and [fzf](https://github.com/junegunn/fzf) are useful tools, but maybe I should remove these because they aren't necessary.

HackerFoo:
> I don't know much about nix, but aren't you missing yosys? Also unclear why this belongs as part of the symbiflow repository directly?

`yosys` is listed on line 63. This is to get people using Nix started quickly by just typing `nix-shell shell.nix` and they have everything they need installed and ready to go.

shell.nix
HackerFoo: Apparently not. Maybe the upstream vtr failed without it.

HackerFoo:
I should note that this is also a work in progress, because I wasn't able to get `cmake -DUSE_CONDA=FALSE ..` to work, so this actually uses a [lightweight sandbox](https://nixos.org/nixpkgs/manual/#sec-fhs-environments) to allow Conda to work.

I hope to eventually not need the build system to set up the environment. See #559.

shell.nix
```diff
@@ -0,0 +1,96 @@
...
+      vtr
+      python37
+      git
+      cmake
+      gcc
+      clang
+      gnumake
+      cmake
+      conda
+      yosys
```
mithro: Is this going to be Yosys from git?

shell.nix
```diff
@@ -0,0 +1,96 @@
...
+      openocd
+      libxml2
+      libxslt
+      git
+      wget
+      zlib
+      binutils
+      xxd
+      verilog
```
mithro: What is `verilog`? Did you mean `iverilog`?

shell.nix
```diff
@@ -0,0 +1,96 @@
...
+      sqlite
+      nodejs
+      pkg-config
+      libftdi
+      fasm
+      icestorm
```
mithro: Should there be an override like the fasm one above which uses icestorm from the local submodule?

shell.nix
```diff
@@ -0,0 +1,96 @@
...
+      progressbar2
+      simplejson
+      pyserial
+      setuptools
+      textx
+      pytest
+      intervaltree
+      yapf
+      numpy
+      tox
```
mithro: What about prjxray stuff from third_party?

shell.nix
HackerFoo: Yeah. It is just called verilog in nixpkgs.

shell.nix
HackerFoo: It's the latest version in nixpkgs. Should it be pinned to the version in CMakeLists.txt?

shell.nix
HackerFoo: prjxray doesn't seem to get installed in any way. It's still in third_party.

shell.nix
HackerFoo: Done.

mithro:
@HackerFoo Looks like you have overwritten your master branch this pull request came from with the RAM32X1D change.

I was going to say we can merged the nix file if we move it out of the top level directory.

mithro:
@HackerFoo - Are you going to resubmit your nix stuff as a new pull request?

HackerFoo:
Yeah, I can do that.

PR #565, opened 2019-04-10
Modify the "buttons" project to change switch/led behavior.

HackerFoo:
The previous behavior (out = in) was confusing, because the switches
behaved exactly as they did before the BASYS3 was programmed.

Signed-off-by: Dusty DeWeese <dustin.deweese@gmail.com>

### DIFF
--- a/xc7/tests/buttons/buttons_basys3.v
+++ b/xc7/tests/buttons/buttons_basys3.v
@@ -2,5 +2,5 @@ module top(
 	input [15:0] in,
 	output [15:0] out
 );
-  assign out = in;
+  assign out = in + 16'h5A69;
 endmodule


litghost:
The point of the buttons test was specifically to test that we can route each switch and LED, nothing more.  You will note that the 8-segment LED's change behavior, etc.

HackerFoo:
Makes sense. I found "xc7/tests/counter" which is a better test that programming worked.

litghost:
> Makes sense. I found "xc7/tests/counter" which is a better test that programming worked.

Counter is a better example, but does required "more" to work.  "buttons" requires no clock (e.g. combintorial only).  Counter requires both clock and carry chain to work.

PR #567, opened 2019-04-10
Add initial FF equiv tests

litghost:
This adds an initial testbench and failing equiv tests for the FF CE/SR test.  I'm going to continue to expand the initial testbench to try to see if I can determine why the equiv test is failing.  It has something to do the "^Q", "|Q" and "&Q" assign statements.  Unclear if this is a bug in fasm2verilog or a bug in VPR routing or what.


xc7/tests/ff_sr_ce/CMakeLists.txt
```diff
@@ -1,9 +1,52 @@
...
+    add_file_target(FILE ff_ce_sr_${num_ff}_tb.v GENERATED)
+    get_file_target(TB_SOURCE_TARGET ff_ce_sr_testbench.v)
+    get_file_target(TB_TARGET ff_ce_sr_${num_ff}_tb.v)
+    get_target_property(INCLUDE_FILES ${TB_SOURCE_TARGET} INCLUDE_FILES)
+    set_target_properties(${TB_TARGET} PROPERTIES INCLUDE_FILES
+        ${INCLUDE_FILES})
+
+    add_fpga_target(
+        NAME ff_ce_sr_${num_ff}
+        BOARD basys3
```
elmsfu: Will we want to run this on other boards? If so should the board and pcf file be parameters to this function?

litghost: I think this is little value in running this on other boards, all it will do is inflate the CI runtime for little value.  In addition because of the ROI, each top file will vary based on pinouts.  Once the ROI is gone we can revisit this.

PR #574, opened 2019-04-11
WIP [DNM]: generate arch.xml from verilog templates including timing info

kgugala:
This PR is to track and discuss the timing data import into arch.xml file. The idea is to update all the BEL sim.v files to include timing templates (see https://github.com/antmicro/symbiflow-arch-defs/blob/25cb1d212d4890fb2a22a5576a56ec3a0e50fd05/xc7/primitives/common_slice/Nlut/ntemplate.Nlut.sim.v)

 The script (https://github.com/antmicro/symbiflow-arch-defs/blob/25cb1d212d4890fb2a22a5576a56ec3a0e50fd05/utils/ntemplate_timings.py) can update such file with data read from sdf file.

Next, v2x tool can be used to generate xml file.

Initial tests are promising. The script called with:

```
ntemplate_timings.py --template ntemplate.Nlut.sim.v --sdf CLBLL_L.sdf --npattern a alut.sim.v  --bel LUT5 --bel LUT6 --process slow --delay max --instance SLICEL/A5LUT --instance SLICEL/A6LUT
```

the following verilog file - https://pastebin.com/TSffCRKZ
v2x generated the following xml - https://pastebin.com/SmbucBn9


mithro:
It might be better / more flexible to do this at the XML level rather than the verilog level?

elmsfu:
> It might be better / more flexible to do this at the XML level rather than the verilog level?

There was some brief discussion([logs here](https://logs.timvideos.us/%23symbiflow/%23symbiflow.2019-04-10.log.html#t2019-04-10T14:05:22)) on which way to take on this.

<details>
<summary>Relevant conversation</summary>

|user| msg| time|
| - | - | - |
sf-slack | <kgugala> @litghost @mithro I'm looking at the possibilities of adding the timing information to architecture definition and there seem to be two options | 14:05
sf-slack | <kgugala> one is to implement a script to read the arch.merged.xml file and add/update timing entries basing on sdfs generated with timining fuzzer | 14:06
sf-slack | <kgugala> this option is pretty straight forward - it will add one step between xml merging and carry chains specialization | 14:07
sf-slack | <kgugala> I'm not sure if this scales, though | 14:07
sf-slack | <kgugala> the second option is to generate all the arch xmls from verilog (this is something @mithro mentioned some time ago) | 14:08
sf-slack | <kgugala> this is much more work, because we'd need to reimplment all the arch.xml generation | 14:09
sf-slack | <kgugala> but this would allow us to create verilog templates for every BEL and CLB structure | 14:09
sf-slack | <kgugala> a separate script can use the verlilog templates and merge them with timings from sdf | 14:10
sf-slack | <kgugala> in the end v2x tool can generate arch xml | 14:10
sf-slack | <kgugala> this approach is much more versatile, but requires more work | 14:11
sf-slack | <kgugala> what do you think | 14:11
sf-slack | <mkurc> I wonder whether it is possible in the second option to include the timings in Verilog models and use them for simulation. (just a random thought) | 14:16
litghost | hackerfoo: What changed?  I'm also working on writing a testbench for the FF SR/CE test should we should probably commonize some things, at a minimum an assert of some kind | 14:55
litghost | kgugala: I think we should proceed with a v2x solution, with the follow caveat: Let's write tests as we god | 14:58
litghost | go* | 14:58
litghost | kgugala: I am working on some FF tests, which also does a little combintorial testing | 14:58

</details>

I can see the value in both approaches. Using `v2x` may mean expanding the attributes to support the different timing information and will require a lot more work. Also it's not clear to me what we will do with interconnect timing information.

@kgugala What is the concern with scaling in patching the xml? If we have to do that during routing import/generation, may be we should try to be consistent.

kgugala:
We can patch XMLs using similar approach like I did for verilog - create a bunch of templates and just fill them with data extracted from sdfs.

@elmsfu  v2x claims it can handle a lot of different parameters, so it shouldn't be a problem (see: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/vlog_to_pbtype.py)

I think interconnects will have to be updated directly in xmls.

PR #575, opened 2019-04-11
Use SymbiYosys to prove FF CE/SR tests

litghost:
This PR adds SymbiYosys to third_party and uses it to prove equivalence on the FF CE/SR tests.  This PR also expands the FF CE/SR test to the other 3 FD types (FDSE, FDPE, FDCE).

This PR is dependent on https://github.com/SymbiFlow/symbiflow-arch-defs/pull/570 (which is blocked by https://github.com/SymbiFlow/prjxray-db/pull/7).

This takes a first attempt at #572 by using submodules.  This is fine for the "PASS" case, but in the failure case, another SMT2 solver needs to be installed.  However for the purposes of getting an equivalence testing working on the FF CE/SR test, this does work.

PR #576, opened 2019-04-11
Add explicit CEUSEDMUX and SRUSEDMUX support.

litghost:
This has been tested with https://github.com/SymbiFlow/symbiflow-arch-defs/pull/575

### DIFF
--- a/xc7/primitives/common_slice/common_slice.model.xml
+++ b/xc7/primitives/common_slice/common_slice.model.xml
@@ -1,5 +1,15 @@
 <models xmlns:xi="http://www.w3.org/2001/XInclude">
- <xi:include href="carry/carry.model.xml"         xpointer="xpointer(models/child::node())" />
- <xi:include href="carry/carry0.model.xml"        xpointer="xpointer(models/child::node())" />
- <xi:include href="../ff/ff.model.xml"            xpointer="xpointer(models/child::node())" />
+  <model name="CE_VCC">
+    <output_ports>
+      <port name="VCC" />
+    </output_ports>
+  </model>
+  <model name="SR_GND">
+    <output_ports>
+      <port name="GND" />
+    </output_ports>
+  </model>
+  <xi:include href="carry/carry.model.xml"         xpointer="xpointer(models/child::node())" />
+  <xi:include href="carry/carry0.model.xml"        xpointer="xpointer(models/child::node())" />
+  <xi:include href="../ff/ff.model.xml"            xpointer="xpointer(models/child::node())" />
 </models>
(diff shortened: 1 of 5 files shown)


litghost:
This breaks on non-trival examples, investigating.

litghost:
> This breaks on non-trival examples, investigating.

Looks like pack patterns were required, but they are nasty.  I'll commit the "working" version, and begin thinking about how to make the pack pattern generation suck less.

litghost:
> > This breaks on non-trival examples, investigating.
> 
> Looks like pack patterns were required, but they are nasty. I'll commit the "working" version, and begin thinking about how to make the pack pattern generation suck less.

Using pack patterns solves the packing issue at the cost of the typical VPR packing problem, which is that it no longer packs a tightly.  Investigating ...

mithro:
The pack pattern is a be horrific but otherwise looks okay I think?

litghost:
> The pack pattern is a be horrific but otherwise looks okay I think?

Amazingly the CI shows green, so that means murax pack/placed/routed!  I want to make those pack patterns generated (rather than hand written), and I also need to verify resource usage.  It's unclear if VPR can pack all 8 FF's into a SLICE any more with this PR.

mkurc-ant:
@litghost So with this PR the VPR will be able to use `VCC` and `GND` local to `CEUSEDMUX` and `SEUSEDMUX` ?

BTW I like the comment "... to make the pack pattern generation suck less" ;)

litghost:
> @litghost So with this PR the VPR will be able to use `VCC` and `GND` local to `CEUSEDMUX` and `SEUSEDMUX` ?

Exactly!

litghost:
Using https://github.com/SymbiFlow/symbiflow-arch-defs/pull/591 I've determined that this PR does need increase sum CLB usage for any of the basys3 test bitstreams.  Given that, I'll focus on making the pack pattern output be less terrible.  This is probably a good time to migrate to v2x, and hopefully I can compactly express the required pack pattern there without too much trouble using generate statements.

litghost:
> Using #591 I've determined that this PR does need increase sum CLB usage for any of the basys3 test bitstreams. Given that, I'll focus on making the pack pattern output be less terrible. This is probably a good time to migrate to v2x, and hopefully I can compactly express the required pack pattern there without too much trouble using generate statements.

https://gist.github.com/litghost/5208eb5304e441cd281ec08aa4a0c975 has the data in CSV format.  Data points:
- ff_pack: Results from this PR
- no_ff_pack: Results from master
- round_robin_ff_pack: Results from this PR with round robin turned on

Issue #582, opened 2019-04-12
Add tiles tag to XML architecture file


acomodi:
With https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/36 the current XML architecture definitions will produce an error due to the fact that `tiles` tag is absent.

There is the need to create a script that can generate the `<tiles>` tag containing all the top level `pb_types`

acomodi:
Solved by https://github.com/SymbiFlow/symbiflow-arch-defs/pull/653

PR #583, opened 2019-04-12
WIP: added script to add tiles tag and equivalent tiles

acomodi:
This PR is an initial effort to change the XML structure and move the tile-specific tags from the top level `pb_types` to the `tiles` tag (solving https://github.com/SymbiFlow/symbiflow-arch-defs/issues/582).

Moreover, the script automatically tries to add the equivalent tiles specified by the user (right now their specification is hard-coded in the script).


mithro:
@acomodi You'll need to update the XML schema here -> https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/xml/fpga_architecture.xsd

mithro:
The [xmlsort.xsl](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/xml/xmlsort.xsl) script is already use to "post process" the XML we have. Maybe it could be used for doing the XML rewriting too?

litghost:
CI is reporting errors on all architectures.

litghost:
@acomodi In the case of xc7, the arch file is generated.  Why not modify the way it is generated?  And for ice40 and testarch, use this method as it option on define_arch?

acomodi:
So, I should modify the initial generation of the arch.xml, instead of manipulating it afterwards, right?

litghost:
> So, I should modify the initial generation of the arch.xml, instead of manipulating it afterwards, right?

For xc7, yes.  That also solves the hard coding of the equivilance in the manipulation script.  For reasons that you've already stated, having a generic arch.old.xml to arch.new.xml is required for VPR to merge the tile change, so we need to develop that tool anyways.  It can be used on ice40 and testarch for now.

acomodi:
All right, I will proceed by deleting the `equivalent tile` part from this script and leave only the swap from top level `pb_type` to `tiles` and add it as a dependency option for ice40 and testarch.
Then I'll create another PR with the modifications for the xc7 architectures

litghost:
@acomodi I think we can close this PR?

acomodi:
@litghost Yep, closing

PR #584, opened 2019-04-12
rr_graph tests and docstrings

elmsfu:


utils/lib/rr_graph/graph2.py
```diff
@@ -299,7 +301,7 @@ def get_nodes_for_pin(self, loc, pin_name):
 
         return self.loc_pin_map[(loc[0], loc[1], pin_idx)]
 
-    def create_edge(self, src_node, sink_node, switch_id, name=None, value=''):
+    def _create_edge(self, src_node, sink_node, switch_id, name=None, value=''):
```
litghost: This is not an internal method, so don't prefix it with _

litghost:
CI is red?

utils/lib/rr_graph/tests/test_channel.py
```diff
@@ -2,7 +2,7 @@
 # Run `python3 -m unittest utils.lib.rr_graph.tests.test_channel`
 import unittest
 
-from lib.rr_graph.channel import Track, ChannelGrid
+from ..channel import Track, ChannelGrid
```
litghost: Weren't we using the relative import less in the test files?

utils/lib/rr_graph/tests/test_tracks.py
```diff
@@ -0,0 +1,60 @@
+import unittest
+from ..tracks import *
```
litghost: Do not use wildcard imports, it breaks linting by polluting the name space.  tracks only has two objects, Track and Tracks, so it isn't a crazy import line anyways.

mithro:
@HackerFoo - FYI

utils/lib/rr_graph/tests/test_channel.py
elmsfu: I did switch to full imports to get tests up and running in #520. I realized that the tests directories needed a `__init__.py` and is better solution.

elmsfu:
> CI is red?

Yeah still a WIP, but wanted to get some early feedback. Will update this afternoon with green and some cleanup.

litghost:
> > CI is red?
> 
> Yeah still a WIP, but wanted to get some early feedback. Will update this afternoon with green and some cleanup.

Okay, I reviewed what there is so far.  Seems pretty reasonable, plus/minus the test module, which I think seems weird, but it's unclear what the standard convention is.

utils/lib/rr_graph/tests/test_channel.py
litghost: I think I disagree that adding __init__.py is the right idea.  If we move to the python setup.py installation of modules, that implies this file will be "installed" too, which seems wrong.  I believe the traditional structure is:

<module>/__init__.py ...
tests
doc

etc.

I don't have a problem with the tests being located in utils/lib/rr_graph/tests, but I do think it's weird that lib/rr_graph/tests is part of the module.

utils/lib/rr_graph/tests/test_tracks.py
elmsfu: Done

elmsfu:
I think I still have a few tests to write and then another rebase with DCO signoffs. I'll request review when it's ready for more thorough review.

utils/lib/rr_graph/points.py
```diff
@@ -62,6 +63,7 @@ def NP(x, y, *n):
 
 class StraightSegment(list):
     class Type(enum.Enum):
+        """ Vertical, Horizontal, Stub"""
```
mithro:
```python
        V = '|'
        VERTICAL = V
        
        H = '-'
        HORIZONTAL = H
        
        S = 'o'
        STUB = S
```

Or maybe
```python
        V = '|'
        V__doc__ = """Vertical"""
        
        H = '-'
        H__doc__ = """Horizontal"""

        S = 'o'
        STUB = S
```
Or https://stackoverflow.com/questions/52062831/how-do-i-properly-document-python-enum-elements

utils/lib/rr_graph/points.py
```diff
@@ -138,6 +141,9 @@ def has_at(self, pos):
             return False
 
     def replace(self, pos):
+        """Replace point at same x and y. Essentially updating the names.
+        If no match is found, no change is occured and returns False
```
mithro:
```suggestion
        If no match is found, list remains unmodified and the function returns False.
```

utils/lib/rr_graph/points.py
```diff
@@ -1147,15 +1163,3 @@ def print_tracks(ret):
 
     print('x = {}'.format(x_tracks))
     print('y = {}'.format(y_tracks))
-
```
mithro: Why is this removed?

elmsfu: It's not used and I think having removing main() methods from library type modules is a good step toward having them as proper packages.

mithro:
@elmsfu Looks like your missing the DCO on a commit?

elmsfu:
> @elmsfu Looks like your missing the DCO on a commit?

I tried Githubs commit from suggestions. Didn't think about the DCO on that commit.

utils/lib/rr_graph/points.py
mithro: I normally find it useful to be able to quickly test a whole file by running the library file...

utils/lib/rr_graph/tests/test_channel.py
mithro: Let's solve this later.

Issue #578, opened 2019-04-11
Improve xc7 fasm2verilog output 


litghost:
With https://github.com/SymbiFlow/symbiflow-arch-defs/pull/577, fasm2v is good enough to be useful, however it could be more useful when debugging correctness issues with some more features.

Ideas to improve output, along with difficulty:
- [ ] (easy) Combine carry chain inputs and outputs into wire bus for easier viewing in during simulation.  This is a pretty straight forward verilog parsing task, basically join the various CARRY4's that are connected via the chain and emit 4 wire buses, S, DI, CO and O for the entire chain length.
- [ ] (easy) Merge simple nets, leaving the intermediate wires, but avoiding the chain.  So
```
assign CLBLM_R_X11Y110_SLICE_X14Y110_CMUX = CLBLM_R_X11Y110_SLICE_X14Y110_C5Q;
assign CLBLL_R_X17Y120_SLICE_X26Y120_SR = CLBLM_R_X11Y110_SLICE_X14Y110_CMUX;

Becomes:

assign CLBLM_R_X11Y110_SLICE_X14Y110_CMUX = CLBLM_R_X11Y110_SLICE_X14Y110_C5Q;
assign CLBLL_R_X17Y120_SLICE_X26Y120_SR = CLBLM_R_X11Y110_SLICE_X14Y110_C5Q;

```
- [ ] (more difficult) Name nets with same name that VPR assigned them.  These names are found in the route output from VPR.  This involves consuming the VPR route output which looks like:

```
Net 0 (scalable_proc.uart.reg_dat_di[31])                                      
                                                                               
Node: 69480 SOURCE (29,26)  Class: 0  Switch: 0 
```

And then matching the VPR rr graph source node OPIN back into a 7-series source, and then renaming the net.

mkurc-ant:
@litghost Regarding "Add a comment above LUT6_2's with the equation form of the LUT from input to output. Ideally they will be reduced equations.":

You mean to write own implementation of an algorithm which converts a truth table with up to 6 inputs to a boolean equation ? Or use an already available solver for this task ? Pure python or an external library/tool ?

I would gladly go with this task, but the only solution I see right now is to use Karnaugh tables related stuff. Or do you have something else on your mind ?

litghost:
> @litghost Regarding "Add a comment above LUT6_2's with the equation form of the LUT from input to output. Ideally they will be reduced equations.":
> 
[...]

PR #585, opened 2019-04-13
Add utilities for parsing VPR route files.

litghost:
This begins to address #578 by providing a mapping from wire_pkey back
to VPR net name.

### DIFF
--- a/utils/lib/parse_route.py
+++ b/utils/lib/parse_route.py
@@ -1,0 +1,55 @@
+""" Library for parsing route output from VPR route files. """
+from collections import namedtuple
+
+Node = namedtuple('Node', 'inode x_low y_low x_high y_high ptc')
+
+
+def format_name(s):
+    """ Converts VPR parenthesized name to just name. """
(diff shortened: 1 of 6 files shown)


mithro:
Not blocking merging this, but FYI - [PRGA](https://prga.readthedocs.io) should also have a Python parsers for VPRs output files. We should collaborate with them on that....

litghost:
> Not blocking merging this, but FYI - [PRGA](https://prga.readthedocs.io) should also have a Python parsers for VPRs output files. We should collaborate with them on that....

Ah, sure.  This parse is kind of single purpose.  I think replacing this with duck2's summer work is probably the right long term plan, and PRGA can use that once it is ready.

mithro:
@litghost Yeah - @duck2's work should make this all much easier!

litghost:
People aren't reviewing this, so just addressing review in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/586

PR #586, opened 2019-04-13
Bind net names from VPR route to fasm2v output

litghost:
This addresses one of the output improvements suggested in #578

This PR should be review as after #585


xc7/fasm2bels/net_map.py
```diff
@@ -0,0 +1,72 @@
+""" Utilities for match VPR route names with xc7 site pin sources. """
+from collections import namedtuple
+from lib.parse_route import find_net_sources
+import re
+"""
```
mithro: This file has two docstrings?

xc7/fasm2bels/verilog_modeling.py
```diff
@@ -780,6 +788,38 @@ def set_iostandard(self, iostandards):
...
+        self.net_map = net_map
+
+    def check_for_net_name(self, wire_pkey):
+        if wire_pkey in self.net_map:
+            # Top-level port names supress net names.
+            name = self.net_map[wire_pkey]
+            if name in self.top_level_signal_nets:
+                return None
+
+            # Transform net names into escaped id and bus selection:
```
mithro: This comment should be a docstring / doctest?

xc7/fasm2bels/net_map.py
litghost: The first docstring is for module, the second docstring is for the Net namedtuple.

utils/lib/rr_graph/graph2.py
```diff
@@ -83,7 +83,9 @@ class Graph(object):
     """ Simple object for working with VPR RR graph. This class does not handle
     """
 
-    def __init__(self, switches, segments, block_types, grid, nodes):
+    def __init__(
+            self, switches, segments, block_types, grid, nodes, need_edges=True
```
elmsfu: nit: not sure `need_edges` is descriptive, is `build_pin_edges` accurate?

xc7/fasm2bels/net_map.py
elmsfu: For namedtuples how would you feel about this pattern:
```python
class Net(namedtuple('Net', 'name wire_pkey tile site_pin')):
    """
    Args:
        name (str): VPR net name
        wire_pkey (int): Wire table primary key.  This is unique in the part.
        tile (str): Name of tile this wire belongs too.  This is redundant
            information wire_pkey uniquely indentifies the tile.
        site_pin (str): Name of site pin this wire belongs. This is redundant
            information wire_pkey uniquely indentifies the site pin.
    """
```

utils/print_net_sources.py
```diff
@@ -0,0 +1,18 @@
+import argparse
+from lib.parse_route import find_net_sources
+
+
+def main():
+    parser = argparse.ArgumentParser(description="")
```
elmsfu: Add description please

xc7/fasm2bels/net_map.py
litghost: It appears to work the same, so I don't mind it.

xc7/fasm2bels/verilog_modeling.py
litghost: Refactored into a function and added some tests.

xc7/primitives/common_slice/Nlut/simtest/test_alut.py
```diff
@@ -1,51 +0,0 @@
-"""
```
mithro: Did you mean to delete this file?

mithro:
If @elmsfu reviews and is happy, feel free to merge.

xc7/primitives/common_slice/Nlut/simtest/test_alut.py
litghost: Yes, cocotb is no longer a dependency we support without significant effort.  We should find another approach to testing sim.v files.

elmsfu: I created #598 to come back on testing sim.v files. Is there a reason removing this should be in this PR? If we are going to delete them, I'd prefer an independent PR that removes all of them at once so we can revert to resurrect parts in the future.

litghost: > I created #598 to come back on testing sim.v files. Is there a reason removing this should be in this PR? If we are going to delete them, I'd prefer an independent PR that removes all of them at once so we can revert to resurrect parts in the future.

In order to enable pytest on the xc7 tree, all python file imports must work.  In this case, test_alut.py fails because cocotb is no longer supported.

elmsfu: ok makes sense.

PR #587, opened 2019-04-13
Make scalable tests debuggable with fasm2v

litghost:
- Remove clock divider from scalable tests.
- Use all IO pins in ROI.
- Re-enable round robin packing.

### DIFF
--- a/tests/9-scalable_proc/basys3_top.v
+++ b/tests/9-scalable_proc/basys3_top.v
@@ -1,60 +1,46 @@
 module top
 (
 input  wire         clk,
-input  wire         rst,
 
 output wire         ser_tx,
 input  wire         ser_rx,
 
-output wire [3:0]   leds
+input wire [15:0]   sw,
+output wire [15:0]  led
 );
 
 // ============================================================================
 
-// Clock divider k = 1/(1<<N_CLK_DIV)
-localparam integer N_CLK_DIV = 7;
+reg rst = 1;
+reg rst1 = 1;
+reg rst2 = 1;
+reg rst3 = 1;
+assign led[0] = rst;
+assign led[13:1] = sw[13:1];
+assign led[14] = ^sw;
+assign led[15] = ser_rx;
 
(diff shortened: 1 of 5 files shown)


xc7/make/arch_define.cmake
```diff
@@ -23,6 +23,7 @@ function(ADD_XC7_ARCH_DEFINE)
       --place_algorithm bounding_box
       --enable_timing_computations off
       --allow_unrelated_clustering on
+      --round_robin_prepacking on
```
mithro: Can you add a FIXME / TODO on the round-robin-packing to remove it after XYZ is done?

tests/9-scalable_proc/basys3_top.v
```diff
@@ -1,60 +1,46 @@
...
 input  wire         ser_rx,
 
-output wire [3:0]   leds
+input wire [15:0]   sw,
+output wire [15:0]  led
 );
 
 // ============================================================================
 
-// Clock divider k = 1/(1<<N_CLK_DIV)
```
mithro: Why are you removing the clock divider?

litghost: VPR implements the clock divider in logic, and Vivado will route the output of the logic clock back on the global network.  We can probably disable this behavior in Vivado, but "real" designs won't really have clock divider implemented this way.  Once we are out of the ROI, BUFR, PLL and MMCM.

xc7/make/arch_define.cmake
litghost: done

tests/9-scalable_proc/basys3_top.v
mkurc-ant: Lets hope that it will still work at 100MHz when implemented with the VPR

litghost: > Lets hope that it will still work at 100MHz when implemented with the VPR

BRAM n8 appears to be okay, ran for several thousand cycles without error. BRAM n9 has errors every 15k or so.

PR #591, opened 2019-04-15
Add usage comparision tools.

litghost:
These tools were created to provide a way to answer whether https://github.com/SymbiFlow/symbiflow-arch-defs/pull/576 and features like round robin packing were adversely affecting packing quality.

### DIFF
--- a/utils/compare_usage.py
+++ b/utils/compare_usage.py
@@ -1,0 +1,81 @@
+"""Tool for generating a comparision CSV between multiple usage.json files.
+
+utils/gather_usage.py should be used to generate usage.json in various
+configurations for comparision.
+
+"""
+import argparse
+import json
+import os.path
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+
+    parser.add_argument(
+        'usage_json',
+        nargs='+',
+        help="usage.json files generated from gather_usage.py"
+    )
+
+    args = parser.parse_args()
+
+    blocks = set()
+    bases = set()
+    targets = {}
+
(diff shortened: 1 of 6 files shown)


utils/usage.py
```diff
@@ -1,16 +1,6 @@
 import argparse
```
elmsfu: I was thinking we should rename this file. maybe `report_block_usage`

utils/lib/parse_usage.py
```diff
@@ -0,0 +1,20 @@
+import re
```
elmsfu: This is all based on a VPR specific format, right? If so maybe explain that in a docstring.

utils/compare_usage.py
```diff
@@ -0,0 +1,81 @@
...
+    row2 = ['']
+    for block in sorted(blocks):
+        row1.append(block)
+        for _ in range(len(bases) - 1):
+            row1.append('')
+
+        for base in sorted(bases):
+            row2.append(base)
+
+    print(','.join(row1))
```
elmsfu: Should we use the csv module? I'm not sure if would make the above loops simpler or just complicate the code.

litghost: I thought about it, and because of the double header row, and joined cells, this is easier to express without the CSV module.

utils/lib/parse_usage.py
litghost: I updated the docstring with more detail, take a look.

utils/usage.py
litghost: Sure

PR #594, opened 2019-04-16
gen.cmake: handle generated verilog and xml deps

### LINKED ISSUE #262: XML dependencies are calculated before ntemplate subsitution
The following `ntemplate.Nlut.pb_type.xml` which has an `xi:include` fails;
```xml
<pb_type name="{N}LUT" num_pb="1" xmlns:xi="http://www.w3.org/2001/XInclude">
 <!-- Normal LUT input/outputs -->
 <input  name="A1"   num_pins="1" />
 <input  name="A2"   num_pins="1" />
 <input  name="A3"   num_pins="1" />
 <input  name="A4"   num_pins="1" />
 <input  name="A5"   num_pins="1" />
 <input  name="A6"   num_pins="1" />
 <output name="O5"   num_pins="1" />
 <output name="O6"   num_pins="1" />

 <!-- Extended memory LUT pins -->
 <input  name="WA1"  num_pins="1" />
 <input  name="WA2"  num_pins="1" />
 <input  name="WA3"  num_pins="1" />
 <input  name="WA4"  num_pins="1" />
 <input  name="WA5"  num_pins="1" />
 <input  name="WA6"  num_pins="1" />
 <input  name="WA7"  num_pins="1" />
 <input  name="WA8"  num_pins="1" />
 <input  name="WE"   num_pins="1" />
 <input  name="DI1"  num_pins="1" />
 <input  name="DI2"  num_pins="1" />
 <output name="MC31" num_pins="1" />

 <xi:include href="../../slicel/{N}lut/{N}lut.pb_type.xml" xpointer="xpointer(pb_type/mode)" />
 <mode name="DISTRIBUTED_RAM">
  <pb_type name="RAM64X1D" num_pb="1" blif_model=".subckt RAM64X1D" class="memory">
   <clock  name="CLK1"          num_pins="1"  port_class="clock" />

   <!-- A1 - Read port -->
   <input  name="A1ADDR"        num_pins="6" port_class="address1" />
   <input  name="A1DATA"        num_pins="1" port_class="data_in1" />
   <!-- <output name="A1DATA"   num_pins="1"  port_class="data_out1" /> -->
   <!-- <input  name="A1WE"     num_pins="1"  port_class="write_en1" /> -->

   <!-- B1 - Write Port -->
   <input  name="B1ADDR"        num_pins="6" port_class="address2" />
   <input  name="B1DATA"        num_pins="1" port_class="data_in2" />
   <input  name="B1WEN"         num_pins="1" port_class="write_en2" />
   <!-- <output name="B1DATA"        num_pins="1" port_class="data_out2" /> -->
  </pb_type>
  <interconnect>
   <direct name="{N}LUT.A1" input="{N}LUT.A1" output="RAM64X1D.B1ADDR[0]" />
   <direct name="{N}LUT.A2" input="{N}LUT.A2" output="RAM64X1D.B1ADDR[1]" />
   <direct name="{N}LUT.A3" input="{N}LUT.A4" output="RAM64X1D.B1ADDR[2]" />
   <direct name="{N}LUT.A4" input="{N}LUT.A5" output="RAM64X1D.B1ADDR[3]" />
   <direct name="{N}LUT.A5" input="{N}LUT.A6" output="RAM64X1D.B1ADDR[4]" />
   <direct name="{N}LUT.A6" input="{N}LUT.A7" output="RAM64X1D.B1ADDR[5]" />
  </interconnect>
 </mode>
```

It causes the following cmake output;
```
tansell@tansell:~/github/SymbiFlow/symbiflow-arch-defs-a7/build$ cmake ..
CMake Error at make/file_targets.cmake:83 (get_target_property):
  get_target_property() called with non-existent target
  "file__home_tansell_github_SymbiFlow_symbiflow-arch-defs-a7_artix7_primitives_slicel_{N}lut_{N}lut.pb_type.xml".
  make/file_targets.cmake:243 (append_file_dependency)
  artix7/primitives/slicem/Nlut/CMakeLists.txt:2 (add_file_target)


CMake Error at make/file_targets.cmake:83 (get_target_property):
  get_target_property() called with non-existent target
  "file__home_tansell_github_SymbiFlow_symbiflow-arch-defs-a7_artix7_primitives_slicel_{N}lut_{N}lut.pb_type.xml".
  make/file_targets.cmake:96 (append_file_dependency)
  make/gen.cmake:321 (append_file_dependency)
  artix7/primitives/slicem/Nlut/CMakeLists.txt:9 (n_template)


CMake Error at make/file_targets.cmake:83 (get_target_property):
  get_target_property() called with non-existent target
  "file__home_tansell_github_SymbiFlow_symbiflow-arch-defs-a7_artix7_primitives_slicel_{N}lut_{N}lut.pb_type.xml".
  make/file_targets.cmake:96 (append_file_dependency)
  make/gen.cmake:321 (append_file_dependency)
  artix7/primitives/slicem/Nlut/CMakeLists.txt:9 (n_template)


CMake Error at make/file_targets.cmake:83 (get_target_property):
  get_target_property() called with non-existent target
  "file__home_tansell_github_SymbiFlow_symbiflow-arch-defs-a7_artix7_primitives_slicel_{N}lut_{N}lut.pb_type.xml".
  make/file_targets.cmake:96 (append_file_dependency)
  make/gen.cmake:321 (append_file_dependency)
  artix7/primitives/slicem/Nlut/CMakeLists.txt:9 (n_template)


CMake Error at make/file_targets.cmake:83 (get_target_property):
  get_target_property() called with non-existent target
  "file__home_tansell_github_SymbiFlow_symbiflow-arch-defs-a7_artix7_primitives_slicel_{N}lut_{N}lut.pb_type.xml".
  make/file_targets.cmake:96 (append_file_dependency)
  make/gen.cmake:321 (append_file_dependency)
  artix7/primitives/slicem/Nlut/CMakeLists.txt:9 (n_template)


-- Configuring incomplete, errors occurred!
See also "/home/tansell/github/SymbiFlow/symbiflow-arch-defs-a7/build/CMakeFiles/CMakeOutput.log".
```

kgugala:
V2X and N_TEMPLATE now can handle dependency on a generated XML or Verilog file. The list of the dependencies can be passed via GENERATEDXMLDEPS and GENERATEDVERILOGDEPS for both n_template and v2x functions.

This may fix #262

### DIFF
--- a/make/gen.cmake
+++ b/make/gen.cmake
@@ -300,6 +300,7 @@ function(N_TEMPLATE)
 
   foreach(PREFIX ${N_TEMPLATE_PREFIXES})
     foreach(SRC ${N_TEMPLATE_SRCS})
+      set(REAL_INCLUDE_FILES "")
       string(
         REPLACE
           "ntemplate."
@@ -314,6 +315,28 @@ function(N_TEMPLATE)
           SRC_WITH_PREFIX
           ${SRC_NO_NTEMPLATE}
       )
+      get_file_target(SRC_TARGET_NAME ${SRC})
+      get_target_property(SRC_INCLUDE_FILES ${SRC_TARGET_NAME} INCLUDE_FILES)
+      foreach(INC ${SRC_INCLUDE_FILES})
+        get_filename_component(INC_FILE ${INC} NAME)
+        get_filename_component(INC_DIR ${INC} DIRECTORY)
+        # template all the include files
+        string(
+          REPLACE
+            "ntemplate."
+            ""
+            INC_NO_NTEMPLATE
+            ${INC_FILE}
+        )
+        string(
+          REPLACE
+            "N"
+            ${PREFIX}
+            INC_WITH_PREFIX
+            ${INC_NO_NTEMPLATE}
+        )
+        list(APPEND REAL_INCLUDE_FILES ${INC_DIR}/${INC_WITH_PREFIX})
+      endforeach()
       get_file_location(SRC_LOCATION ${SRC})
       set(DEPS "")
       append_file_dependency(DEPS ${SRC})
@@ -329,7 +352,9 @@ function(N_TEMPLATE)
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
       )
 
-      add_file_target(FILE ${SRC_WITH_PREFIX} GENERATED DEPENDS ${SRC})
+      add_file_target(FILE ${SRC_WITH_PREFIX} GENERATED)
+      get_file_target(SRC_TARGET_NAME ${SRC_WITH_PREFIX})
+      set_target_properties(${SRC_TARGET_NAME} PROPERTIES INCLUDE_FILES "${REAL_INCLUDE_FILES}")
 
       list(APPEND OUTPUTS ${SRC_WITH_PREFIX})
 
--- a/make/gen.cmake
+++ b/make/gen.cmake
@@ -31,6 +31,8 @@ function(V2X)
   set(INCLUDES "")
 
   set(DEPENDS_LIST "")
+  set(MODEL_INCLUDE_FILES "")
+  set(PB_TYPE_INCLUDE_FILES "")
   get_target_property_required(PYTHON3 env PYTHON3)
   get_target_property(PYTHON3_TARGET env PYTHON3_TARGET)
   list(APPEND DEPENDS_LIST ${PYTHON3} ${PYTHON3_TARGET})
@@ -60,6 +62,8 @@ function(V2X)
 
       append_file_dependency(DEPENDS_LIST ${INCLUDE_SRC_DIR}/${INCLUDE_ROOT}.model.xml)
       append_file_dependency(DEPENDS_LIST ${INCLUDE_SRC_DIR}/${INCLUDE_ROOT}.pb_type.xml)
+      list(APPEND MODEL_INCLUDE_FILES ${INCLUDE_SRC_DIR}/${INCLUDE_ROOT}.model.xml)
+      list(APPEND PB_TYPE_INCLUDE_FILES ${INCLUDE_SRC_DIR}/${INCLUDE_ROOT}.pb_type.xml)
     endforeach()
   endforeach()
 
@@ -97,6 +101,8 @@ function(V2X)
   )
 
   add_file_target(FILE "${V2X_NAME}.pb_type.xml" GENERATED)
+  get_file_target(SRC_TARGET_NAME "${V2X_NAME}.pb_type.xml")
+  set_target_properties(${SRC_TARGET_NAME} PROPERTIES INCLUDE_FILES "${PB_TYPE_INCLUDE_FILES}")
 
   add_custom_command(
     OUTPUT "${V2X_NAME}.model.xml"
@@ -111,6 +117,8 @@ function(V2X)
   )
 
   add_file_target(FILE "${V2X_NAME}.model.xml" GENERATED)
+  get_file_target(SRC_TARGET_NAME "${V2X_NAME}.model.xml")
+  set_target_properties(${SRC_TARGET_NAME} PROPERTIES INCLUDE_FILES "${MODEL_INCLUDE_FILES}")
 
   add_custom_target(
     ${V2X_NAME}
--- a/make/gen.cmake
+++ b/make/gen.cmake
@@ -264,6 +264,38 @@ function(MUX_GEN)
   endif()
 endfunction(MUX_GEN)
 
+function(GET_TEMPLATED_FILENAME var SRC PREFIX)
+  # ~~~
+  # GET_TEMPLATED_FILENAME(
+  #   NAME <name>
+  #   var <calculated templated filename>
+  #   SRC <template file>
+  #   PREFIX <template prefixes>
+  #   )
+  # ~~~
+  #
+  # GET_TEMPLATED_FILENAME calculates file name that from given template and prefix.
+  # The template file should have a form of ntemplate.<rest>.
+  # The function removes the "ntemplate" prefix and converting all N's in file name
+  # to <prefix>.
+  #
+  string(
+    REPLACE
+      "ntemplate."
+      ""
+      SRC_NO_NTEMPLATE
+      ${SRC}
+  )
+  string(
+    REPLACE
+      "N"
+      ${PREFIX}
+      SRC_WITH_PREFIX
+      ${SRC_NO_NTEMPLATE}
+  )
+set(${var} ${SRC_WITH_PREFIX} PARENT_SCOPE)
+endfunction()
+
 function(N_TEMPLATE)
   # ~~~
   # N_TEMPLATE(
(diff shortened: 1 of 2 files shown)


make/gen.cmake
```diff
@@ -19,7 +19,7 @@ function(V2X)
   # files.  DO_NOT_APPLY_VERILOG_IMAGE_GEN suppress this default.
   set(options DO_NOT_APPLY_VERILOG_IMAGE_GEN)
   set(oneValueArgs NAME TOP_MODULE)
-  set(multiValueArgs SRCS)
+  set(multiValueArgs SRCS GENERATEDXMLDEPS GENERATEDVERILOGDEPS)
```
litghost: Why should the XML dependencies need to input?  And why not just have the verilog dependencies communicated via SRCS INCLUDE files?

make/gen.cmake
```diff
@@ -29,8 +29,22 @@ function(V2X)
   )
 
   set(INCLUDES "")
-
   set(DEPENDS_LIST "")
+  set(VERILOG_DEPS "")
+
+  if(V2X_GENERATEDXMLDEPS)
+    list(APPEND INCLUDES ${V2X_GENERATEDXMLDEPS})
```
litghost: Are these absolute file locations or file targets?

make/gen.cmake
```diff
@@ -283,9 +298,9 @@ function(N_TEMPLATE)
   #
   # If APPLY_VERILOG_IMAGE_GEN, ADD_VERILOG_IMAGE_GEN will be invoked with each
   # output file.
-  set(options APPLY_V2X APPLY_VERILOG_IMAGE_GEN)
+  set(options APPLY_V2X APPLY_VERILOG_IMAGE_GEN DEPS)
   set(oneValueArgs NAME)
-  set(multiValueArgs SRCS PREFIXES)
+  set(multiValueArgs SRCS PREFIXES GENERATEDVERILOGDEPS GENERATEDXMLDEPS)
```
litghost: Why should N_TEMPLATE need these arguments, it should all be computable from the SCANNER dependencies?

make/gen.cmake
```diff
@@ -332,15 +374,15 @@ function(N_TEMPLATE)
...
+        v2x(NAME ${V2X_NAME} SRCS ${SRC_WITH_PREFIX} GENERATEDXMLDEPS ${GEN_XML_DEPS} GENERATEDVERILOGDEPS ${GEN_VLOG_DEPS})
       endif()
+
       if(${N_TEMPLATE_APPLY_VERILOG_IMAGE_GEN})
       endif()
     endforeach(SRC)
   endforeach(PREFIX)
 
-  add_custom_target(${N_TEMPLATE_NAME} ALL DEPENDS ${OUTPUTS})
+  add_custom_target(${N_TEMPLATE_NAME} ALL DEPENDS ${OUTPUTS} ${GEN_VLOG_DEPS} ${GEN_XML_DEPS})
```
litghost: The OUTPUTS should already have transitive dependencies, so this shouldn't be required if things are working correctly.

make/gen.cmake
kgugala: the issue I had was that some xml files are included in other generated xml files, so SCANNER is able to determine the dependencies only for the first inclusion. The second (deeper) xml was newer added into dependencies.

litghost: Correct, hence why append_file_dependencies is recursive.  Each file is scanned and points to the next up the chain.  ntemplate should recursively walk the INCLUDE_FILES lists and apply the ntemplate transformation, and then assign INCLUDE_FILES to the relevant targest.  V2X should be transforming the verilog INCLUDE_FILES list to XML  etc.

make/gen.cmake
kgugala: as for Verilog:
SRCS are parsed as N template. The intention here is that the VERILOG deps were already generated and should not be processed furher.

XML deps:
some generated xmls are included by other generated xmls. This is to ensure that those will be included in the deps list.

make/gen.cmake
kgugala: file targets

make/gen.cmake
kgugala: Without this the generated files included by other generated files were not included in the dependencies list

litghost: I don't believe this statement does what you think it does.  If the "${GEN_VLOG_DEPS} ${GEN_XML_DEPS}" are required to build the file, this line does not help with that.  If these files are required when something else uses this file, this line does not help with that either.  What do you actually think this line does?

make/gen.cmake
litghost: Rather than pass the file targets as arguments, why not store the relevant absolute file locations in the SRCS target priority INCLUDE_FILES, and then append_file_dependencies will correctly generate the transitive dependency list.  The output XML will need to have their INCLUDE_FILES property set, but that is required not match whether that is done implicitly or explicitly.

make/gen.cmake
litghost: This should be handle via the INCLUDE_FILES property, otherwise the dependencies will not propagate to other tools.  I'm not clear what these arguments do that could not have been done with the INCLUDE_FILES property and append_file_dependency

make/gen.cmake
kgugala: hmm, the problem I had was that append_file_dependencies was not walking through the generated files included in other generated files (included in the first one)

litghost: > hmm, the problem I had was that append_file_dependencies was not walking through the generated files included in other generated files (included in the first one)

This would happen if INCLUDE_FILES wasn't set on the generated file targets.  This is something that v2x and n_template need to handle.

make/gen.cmake
```diff
@@ -314,6 +315,28 @@ function(N_TEMPLATE)
           SRC_WITH_PREFIX
           ${SRC_NO_NTEMPLATE}
       )
+      get_file_target(SRC_TARGET_NAME ${SRC})
+      get_target_property(SRC_INCLUDE_FILES ${SRC_TARGET_NAME} INCLUDE_FILES)
+      foreach(INC ${SRC_INCLUDE_FILES})
+        get_filename_component(INC_FILE ${INC} NAME)
+        get_filename_component(INC_DIR ${INC} DIRECTORY)
+        # template all the include files
+        string(
```
litghost: Please refactor the filename ntemplate logic into a CMake function

utils/n.py
```diff
@@ -24,7 +25,8 @@ def main(args):
     assert_eq(outname_value, outfile)
 
     template = open(templatepath, "r").read()
-    open(outpath, "w").write(template.format(N=replacement.upper()))
+    template = re.sub(r'(["\s])ntemplate\.N', r'\1{FN}', template)
```
litghost: This is a little restrictive, but lets start here.

make/gen.cmake
kgugala: done

kgugala:
> Need to update INCLUDE_FILES for V2X as well? Otherwise other tools will have problems?

done

make/gen.cmake
```diff
@@ -97,6 +97,8 @@ function(V2X)
   )
 
   add_file_target(FILE "${V2X_NAME}.pb_type.xml" GENERATED)
+  get_file_target(SRC_TARGET_NAME "${V2X_NAME}.pb_type.xml")
+  set_target_properties(${SRC_TARGET_NAME} PROPERTIES INCLUDE_FILES "${INCLUDE_FILES}")
```
litghost: This should be the pb_type XML include list, not the Verilog include list.

make/util.cmake
```diff
@@ -16,3 +16,22 @@ function(get_target_property_required var target property)
   endif()
   set(${var} ${PROP} PARENT_SCOPE)
 endfunction()
+
+function(GET_TEMPLATE_FILENAME var SRC PREFIX)
```
litghost: Please add comment describing this, also this problem belongs close to the ntemplate function

make/gen.cmake
```diff
@@ -111,6 +113,8 @@ function(V2X)
   )
 
   add_file_target(FILE "${V2X_NAME}.model.xml" GENERATED)
+  get_file_target(SRC_TARGET_NAME "${V2X_NAME}.model.xml")
+  set_target_properties(${SRC_TARGET_NAME} PROPERTIES INCLUDE_FILES "${INCLUDE_FILES}")
```
litghost: And this should be the model XML include list, etc

make/gen.cmake
kgugala: yeah, should now be fixed

make/util.cmake
kgugala: I added a description and moved the function to gen.cmake

make/gen.cmake
kgugala: should be fixed now

PR #604, opened 2019-04-17
Import of all pips (bidirectional and pseudo) to channels.db

mkurc-ant:
This PR adds additional columns to the `pip_in_tile` table namely: `can_invert`, `is_directional` and `is_pseudo`. All pseudo pips are now imported from the prjxray database instead of being rejected. All SQL queries regarding the `pip_in_tile` table table were modified to reject pseudo and bidirectional pips.

### DIFF
--- a/xc7/utils/prjxray_create_edges.py
+++ b/xc7/utils/prjxray_create_edges.py
@@ -146,7 +146,8 @@ WHERE
 
 
 def create_find_pip(conn):
-    """ Returns a function that takes (tile_type, pip) and returns pip_in_tile_pkey. """
+    """Returns a function that takes (tile_type, pip) and returns a tuple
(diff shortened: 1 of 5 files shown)


xc7/utils/prjxray_create_edges.py
```diff
@@ -158,7 +158,8 @@ def find_pip(tile_type, pip):
 FROM
   pip_in_tile
 WHERE
-  name = ?
+  is_directional = 1 AND is_pseudo = 0
```
litghost: Rather than filter here, please output is_directional and preserve the "# TODO: Handle bidirectional pips?"

mkurc-ant: I removed this condition from that query. It was not needed here because `make_connections` is always called for non-pseudo and directional pips. The comment about supporting bi-pips is a few lines above the call.

PR #610, opened 2019-04-19
WIP: CLB to SLICE split

mkurc-ant:
This PR is a continuation of work started in #537. It should be merged afterwards.

There are two new tables in the SQL database:
- "tile_former_site" Relates new SLICE tiles (instances) to concrete site instances of a CLB which were used to create them.
- "pip_in_tile_instance" Relates new SLICE tiles (instances) to pips from the pip_in_tile table.

The whole import flow works until the `handle_direction_connections` in `prjxray_assign_tile_pin_direction.py`. Right now it fails there as it cannot find SLICEL tile in the `edge_assignments` map.

### DIFF
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -262,3 +262,11 @@ CREATE TABLE constant_sources(
     FOREIGN KEY(vcc_track_pkey) REFERENCES track(pkey),
     FOREIGN KEY(gnd_track_pkey) REFERENCES track(pkey)
 );
+
+-- Grid location map.
+CREATE TABLE grid_loc_map(
+    phy_tile_pkey INT,
+    vpr_tile_pkey INT,
+    FOREIGN KEY(phy_tile_pkey) REFERENCES phy_tile(pkey),
+    FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey)
+);
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -270,3 +270,12 @@ CREATE TABLE grid_loc_map(
     FOREIGN KEY(phy_tile_pkey) REFERENCES phy_tile(pkey),
     FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey)
 );
+
+-- Tile type map.
+-- Maps physical tiles to VPR tiles. Used for CLB splitting
+CREATE TABLE tile_type_map(
+    phy_tile_type_pkey INT,
+    vpr_tile_type_pkey INT,
+    FOREIGN KEY(phy_tile_type_pkey) REFERENCES tile_type(pkey),
+    FOREIGN KEY(vpr_tile_type_pkey) REFERENCES tile_type(pkey)
+);
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -286,7 +286,7 @@ CREATE TABLE tile_type_map(
 -- Relates a pip_in_tile with a concrete tile instance in the VPR grid.
 -- This is necessary as in the VPR we have generic SLICE tiles but pips
 -- inside them are not generic.
-CREATE TABLE pip_in_tile_instance(
+CREATE TABLE pip(
     vpr_tile_pkey INT,
     pip_in_tile_pkey INT,
     FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey),
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -44,6 +44,16 @@ CREATE TABLE tile_type(
   name TEXT
 );
 
+-- VPR tile type table.
+-- Relates tile types from the tile_type table with their names as they
+-- appear in the VPR
+CREATE TABLE vpr_tile_type(
+  pkey INTEGER PRIMARY KEY,
+  tile_type_pkey INT,
+  name TEXT,
+  FOREIGN KEY(tile_type_pkey) REFERENCES tile_type(pkey)
+);
+
 -- Site type table, used to track site_type using a pkey, and provide
 -- the site_type_pkey <-> name mapping.
 CREATE TABLE site_type(
@@ -66,11 +76,11 @@ CREATE TABLE tile(
   pkey INTEGER PRIMARY KEY,
   name TEXT,
   tile_type_pkey INT,
-  site_remap_pkey INT,
+  vpr_tile_type_pkey INT,
   grid_x INT,
   grid_y INT,
   FOREIGN KEY(tile_type_pkey) REFERENCES tile_type(pkey)
-  FOREIGN KEY(site_remap_pkey) REFERENCES site(pkey)
+  FOREIGN KEY(vpr_tile_type_pkey) REFERENCES vpr_tile_type(pkey)
 );
 
 -- Site pin table, contains names of pins and their direction, along
@@ -280,15 +290,4 @@ CREATE TABLE tile_type_map(
     vpr_tile_type_pkey INT,
     FOREIGN KEY(phy_tile_type_pkey) REFERENCES tile_type(pkey),
     FOREIGN KEY(vpr_tile_type_pkey) REFERENCES tile_type(pkey)
-);
-
--- Pip in tile instance
--- Relates a pip_in_tile with a concrete tile instance in the VPR grid.
--- This is necessary as in the VPR we have generic SLICE tiles but pips
--- inside them are not generic.
-CREATE TABLE pip(
-    vpr_tile_pkey INT,
-    pip_in_tile_pkey INT,
-    FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey),
-    FOREIGN KEY(pip_in_tile_pkey) REFERENCES pip_in_tile(pkey)
 );
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -51,7 +51,17 @@ CREATE TABLE site_type(
   name TEXT
 );
 
--- Tile table, contains type and name of tile and location in grid.
+-- Tile table, contains type and name of tile and location in the physical grid.
+CREATE TABLE phy_tile(
+  pkey INTEGER PRIMARY KEY,
+  name TEXT,
+  tile_type_pkey INT,
+  grid_x INT,
+  grid_y INT,
+  FOREIGN KEY(tile_type_pkey) REFERENCES tile_type(pkey)
+);
+
+-- Tile table, contains type and name of tile and location in the VPR grid.
 CREATE TABLE tile(
   pkey INTEGER PRIMARY KEY,
   name TEXT,
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -44,10 +44,10 @@ CREATE TABLE tile_type(
   name TEXT
 );
 
--- VPR tile type table.
--- Relates tile types from the tile_type table with their names as they
--- appear in the VPR
-CREATE TABLE vpr_tile_type(
+-- Tile type alias.
+-- Used to make generic tile types from specific ones when generating
+-- architecture definition for the VPR
+CREATE TABLE tile_type_alias(
   pkey INTEGER PRIMARY KEY,
   tile_type_pkey INT,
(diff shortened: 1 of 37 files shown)


utils/lib/connection_database.sql
```diff
@@ -252,3 +262,42 @@ CREATE TABLE constant_sources(
...
+-- Grid location map.
+CREATE TABLE grid_loc_map(
+    phy_tile_pkey INT,
+    vpr_tile_pkey INT,
+    FOREIGN KEY(phy_tile_pkey) REFERENCES phy_tile(pkey),
+    FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey)
+);
+
+-- Tile type map.
+-- Maps physical tiles to VPR tiles. Used for CLB splitting
```
litghost: This map is redudaent with the grid_loc_map?

utils/lib/connection_database.sql
```diff
@@ -252,3 +262,42 @@ CREATE TABLE constant_sources(
...
+    vpr_tile_type_pkey INT,
+    FOREIGN KEY(phy_tile_type_pkey) REFERENCES tile_type(pkey),
+    FOREIGN KEY(vpr_tile_type_pkey) REFERENCES tile_type(pkey)
+);
+
+-- Former site table
+-- Contains entries only for CLB split products. Holds information
+-- which site instance correspond to the new tile. It is necessary to know
+-- that when remapping routing resources.
+CREATE TABLE tile_former_site(
```
litghost: Why not add a column into the the tile table? site_remap_pkey maybe.  Then site_remap_pkey == NULL means to use the phy_type connections, and site_remap_pkey != NULL means to place this site at this tile.  Clean and unambigous

utils/lib/connection_database.sql
```diff
@@ -252,3 +262,42 @@ CREATE TABLE constant_sources(
...
+    site_pkey INT,
+    FOREIGN KEY(vpr_tile_pkey) REFERENCES tile(pkey),
+    FOREIGN KEY(site_pkey) REFERENCES site(pkey)
+);
+
+-- Pip in tile instance
+-- Relates a pip_in_tile with a concrete tile instance in the VPR grid.
+-- This is necessary as in the VPR we have generic SLICE tiles but pips
+-- inside them are not generic.
+CREATE TABLE pip_in_tile_instance(
```
litghost: Just call this the "pip" table, just like there is a "wire" table and a "pip_in_tile" table.

xc7/utils/prjxray_form_channels.py
```diff
@@ -202,82 +209,134 @@ def import_grid(db, grid, conn):
...
         )
 
-    build_other_indicies(c)
     c.connection.commit()
 
+    build_phy_tile_grid_indicies(c)
+
 
-def import_nodes(db, grid, conn):
+def import_nodes(db, conn, tile_types_to_split, tile_wire_name_map):
```
litghost: This code doesn't need to change at all?  It should only operate on phy_tiles, and never care about VPR logical tiles.

xc7/utils/prjxray_form_channels.py
```diff
@@ -929,6 +986,160 @@ def create_constant_tracks(conn):
     return create_track(vcc_node, unique_pos), \
            create_track(gnd_node, unique_pos)
 
+vpr_pkey = 999999 # This is a hack which ensures that pkeys of tiles in VPR
```
litghost: This is super wrong

xc7/utils/prjxray_form_channels.py
```diff
@@ -929,6 +986,160 @@ def create_constant_tracks(conn):
...
+        vpr_tile_type_pkey: Tile type pkey
+        phy_tile_pkey: Corresponding tile pkey in the physical grid
+
+    Returns:
+        None
+    """
+
+    c = conn.cursor()
+
+    global vpr_pkey
```
litghost: If you are using the global keyword, you code is wrong, try again.

xc7/utils/prjxray_form_channels.py
```diff
@@ -929,6 +986,160 @@ def create_constant_tracks(conn):
...
+
+    # Insert the tile into the tile grid
+    c.execute(
+        "INSERT INTO tile(pkey, name, tile_type_pkey, grid_x, grid_y)"
+        "VALUES (?, ?, ?, ?, ?)",
+        (vpr_pkey, vpr_tile_name, vpr_tile_type_pkey, vpr_tile_loc[0], vpr_tile_loc[1])
```
litghost: This is super super wrong.

xc7/utils/prjxray_form_channels.py
```diff
@@ -929,6 +986,160 @@ def create_constant_tracks(conn):
...
+    )
+
+    # Insert location correspondence
+    new_pkey = c.lastrowid
+    c.execute(
+        "INSERT INTO grid_loc_map(phy_tile_pkey, vpr_tile_pkey)"
+        "VALUES (?, ?)", (phy_tile_pkey, new_pkey)
+    )
+
+    vpr_pkey += 1
```
litghost: What are you even doing here? This code is completely incorrect, scrap this entire attempt and try again.

xc7/utils/prjxray_form_channels.py
litghost: Why are you even assign a pkey?  Primary keys are generated automatically?

xc7/utils/prjxray_form_channels.py
```diff
@@ -929,6 +986,160 @@ def create_constant_tracks(conn):
...
+
+        # The tile is not being split
+        else:
+
+            # Insert the tile into grid
+            insert_vpr_tile(conn, tile[1], vpr_locs[0], tile_type_pkey, tile[0])
+
+            # If one physical location corresponds to more than one VPR location
+            # then fill the space with artificial EMPTY tiles.
+            for i in range(1, len(vpr_locs)):
```
litghost: In what case does this ever occur?  I think ``assert len(vpr_locs) == 1`` makes more sense here.

xc7/utils/prjxray_form_channels.py
```diff
@@ -929,6 +986,160 @@ def create_constant_tracks(conn):
...
+                # FIXME: the Vivado slice name, coordinates will differ.
+                vpr_tile_name = "%s_X%dY%d" % (vpr_tile_type, vpr_loc[0], vpr_loc[1])
+
+                # Insert the tile into grid
+                vpr_tile_pkey = insert_vpr_tile(conn, vpr_tile_name, vpr_loc, vpr_tile_type_pkey, tile[0])
+
+                # Get site instance basing on the new tile type and site
+                # location offset. Here we assume that new tile type is
+                # equal to its former site type.
+                site_name, site_pkey = c.execute("SELECT name, pkey FROM site WHERE x_coord = (?) AND y_coord = (?) AND site_type_pkey = (SELECT pkey FROM site_type WHERE name = (?))", (loc_ofs[0], loc_ofs[1], vpr_tile_type)).fetchone()
```
litghost: I don't understand why this is required.  Why doesn't the fwd_map provide the site_pkey?  This lookup seems error prone and not required.  Shouldn't this information already be encoded in a table?

utils/lib/connection_database.sql
mkurc-ant: It maps tile types instead of their instances in the grid.

utils/lib/connection_database.sql
mkurc-ant: Makes sense.

utils/lib/connection_database.sql
mkurc-ant: Ok. But mind that this table does not contain all pips. Just those for the tiles that are split.

utils/lib/connection_database.sql
litghost: But there is a never a case to use this map?  When would you use it?

utils/lib/connection_database.sql
litghost: It needs to contain all pips, otherwise the lookup becomes inconsistent.

xc7/utils/prjxray_form_channels.py
mkurc-ant: Yes... I know that it is super wrong. This was used to ensure that pkey values never overlap in phy_tile and tile tables. Needed it for debugging.

xc7/utils/prjxray_form_channels.py
mkurc-ant: Same as above

xc7/utils/prjxray_form_channels.py
mkurc-ant: Same as above

xc7/utils/prjxray_form_channels.py
mkurc-ant: Same as above

xc7/utils/prjxray_form_channels.py
litghost: But there is absolutely no reason to do that.  SQLite will assign unique pkeys, and if you need to know which pkeys are remaps, add a column

xc7/utils/prjxray_form_channels.py
mkurc-ant: While doing the CLB split the entire grid column is being divided into two. There are tiles which are not CLB in such a column and those are not split. So in the end we would have "holes" in the grid. And this piece of code inserts tiles there of NULL type.

Though I'm not sure whether this is actually needed. Are such "holes" in the grid ok? I mean for channel/track formation etc?

litghost: I understand what is happening, and I think this structure is confusing.  I think the tile splitter should generate a tile map AND a "filler" tile list.  Do not add filler tiles to the unmodified tiles, just add a "filler list".

So the logic looks like:
for live tiles:
  if map:
    output mapped tiles
  else:
    output moved tile (1 to 1)

for filler tiles:
  add_filler_tile

xc7/utils/prjxray_form_channels.py
mkurc-ant: But the information itself is required. I guess it can be encoded in the fwd_map. I'll look into it.

utils/lib/connection_database.sql
litghost: > Why not add a column into the the tile table? site_remap_pkey maybe. Then site_remap_pkey == NULL means to use the phy_type connections, and site_remap_pkey != NULL means to place this site at this tile. Clean and unambigous

I thought about this a little more, probably want this on the SLICEL tile type rather than directly on the SLICEL tile instance row.

So the remapped tile will point to the new SLICE tile type, which will have the site_remap_pkey pointing to the SLICEL site type

xc7/utils/prjxray_form_channels.py
litghost: > But the information itself is required. I guess it can be encoded in the fwd_map. I'll look into it.

What I'm saying is this lookup seems fragile and redundant.  The tile splitter code already understood the tile<-> relationship, and should've have encoded in an unambiguous way.

utils/lib/connection_database.sql
mkurc-ant: You are probably right. It is not used now but I can't be sure whether it is not going to be needed at some point of the flow.

litghost: I guess what I'm saying is that this information is better suited to a column in another table, for example the "site_remap_pkey" in the tile type table seems better than this table.

mkurc-ant: AFAIU "site_remap_pkey" is another thing. Here I want to store information about tile type split. For example that the CLBLM_L splits into SLICEM and SLICEL. But now as I think of it I am not sure whether the information is needed anywhere in the flow.

utils/lib/connection_database.sql
mkurc-ant: I think that your first suggestion is better. Because there will be multiple VPR tiles with tile type of SLICEL that will point to different site instances.

xc7/utils/prjxray_form_channels.py
mkurc-ant: I think it has to. Because the `wire` table is filled there. It contains wire to tile instance relation. It is later used when tracks are formed and tile_pkeys must point to VPR tiles.

litghost:
#665 is superceeding this PR

Issue #626, opened 2019-04-23
Clean up the top level directory


mithro:
We should try and keep the top level directory of the repository relatively clean.

I suggest we do the following;
 - [ ] Move `make` to `common/cmake`
 - [ ] Move `yosys` to `common/yosys`
 - [ ] Move `library` to `common/verilog`
 - [ ] Move `cmake.md` to `docs`

Maybe we should rename `common` to `support`?

This leaves us with the following top level things;
 - `build`
 - `CMakeLists.txt`
 - `common`
 - `COPYING`
 - `docs`
 - `ecp5`
 - `ice40`
 - `Makefile`
 - `README.md`
 - `testarch`
 - `tests`
 - `third_party`
 - `utils`
 - `vpr`
 - `xc7`

mithro:
@elmsfu - Thoughts?

mithro:
My general aim is to have as close as possible to one top level directory for each architecture.

mithro:
Issue #571 is part of this.

elmsfu:
I like the suggestions.

A few others to consider
 - [ ] put all archs in a subdirectory 
 - [ ] move `tests` to `common/test_designs`

mithro:
In some ways I think a wire isn't really a "design"....

mithro:
I kind of like having the architectures at the top level directory...

elmsfu:
I can see either way on archs, but I think `tests` is a poor name and is intended to be common

mithro:
Yeah, I agree to that...

PR #627, opened 2019-04-23
Move top-level make directory to common/cmake.

mithro:
* We now use cmake.
 * Makes the top level directory slightly cleaner.
 * Part of #626 


mithro:
Done!

PR #636, opened 2019-04-24
Add the RAM_SHIFTER module to make it easy to build simple RAM tests

HackerFoo:
This adds a test that runs input from switches and output to LEDs on the BASYS3 through a single bit output RAM.

I'm adding one target in this PR to get feedback, and then I'll duplicate it across all memory types.

### DIFF
--- a/xc7/tests/dram_shifter/dram_shifter_32x1s.v
+++ b/xc7/tests/dram_shifter/dram_shifter_32x1s.v
@@ -1,0 +1,46 @@
+// Double buffering with dual-port RAM
+// Uses single-port RAM to write switches to a section while reading from the same section to control LEDs,
+// so each switch acts as if connected directly to the corresponding LED.
+// Flip SW0 to change sections.
+module top (
+        input         clk,
+        input [15:0]  sw,
+        output [15:0] led,
+
+        // not used
+        input         rx,
+        output        tx
+);
+
+    assign tx = rx;  // TODO(#658): Remove this work-around
+
+    wire [3:0]        addr;
+    wire              ram_out;
+    wire              ram_in;
+
+    RAM_SHIFTER #(
+        .IO_WIDTH(16),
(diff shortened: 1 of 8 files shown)


xc7/tests/dram/dram_2_64x1d_shifter.v
```diff
@@ -0,0 +1,52 @@
+// Double buffering with dual-port RAM
```
acomodi: IMHO probably it is worth moving shifter tests in a different directory, called `dram_shifter` for example. And add all the dram types tests in there.

xc7/tests/dram/dram_2_64x1d_shifter.v
```diff
@@ -0,0 +1,52 @@
...
+        output        tx
+);
+
+   assign tx = rx; // makes VPR happy
+
+   wire [4:0]         addr;
+   wire               ram_out;
+   wire               ram_in;
+
+    RAM_SHIFTER #(
```
acomodi: wrong indentation?

HackerFoo: I haven't figured out how to get Emacs to do sane indentation for Verilog yet.

xc7/tests/dram/dram_2_64x1d_shifter.v
HackerFoo: That makes sense.

xc7/tests/dram_shifter/dram_shifter_32x1s.v
```diff
@@ -0,0 +1,46 @@
...
+        input         clk,
+        input [15:0]  sw,
+        output [15:0] led,
+
+        // not used
+        input         rx,
+        output        tx
+);
+
+    assign tx = rx; // makes VPR happy
```
litghost: Should add a TODO to remove this once ROI is removed.

PR #637, opened 2019-04-24
v2x: direct connections now are in the old format

acomodi:
Before this commit direct connections were in the following format:
```
<direct>
  <port name="out[0]" type="input"/>
  <port from="dsp_in_registered" name="out[0]" type="output"/>
</direct>
```
Direct connections's correct format is the following:
```
<direct input="<input>" name="<name>" output="<output>"/>
```

### DIFF
--- a/utils/vlog/tests/dsp_out_registered/golden.pb_type.xml
+++ b/utils/vlog/tests/dsp_out_registered/golden.pb_type.xml
@@ -1,3 +1,4 @@
+<?xml version="1.0"?>
 <pb_type xmlns:xi="http://www.w3.org/2001/XInclude" name="dsp_out_registered" num_pb="1">
   <input name="a" num_pins="32"/>
   <input name="b" num_pins="32"/>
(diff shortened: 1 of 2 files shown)


kgugala:
LGTM. Waiting for green CI before merge

mithro:
This is the **old** format currently used by upstream VtR
```
<direct input="<current_module>.<input>" name="<name>" output="<child_module>.<output>"/>
<direct input="<child_module>.<input>" name="<name>" output="<current_module>.<output>"/>
```

This is the **new** format;
```
<direct>
  <port name="<input>" type="input"/>
  <port from="<child_module>" name="<output>" type="output"/>
</direct>
<direct>
  <port from="<child_module>" name="<input>" type="input"/>
  <port name="<output>" type="output"/>
</direct>
```

Notice how in the current VtR format the **current** module name and the **child** module name need to specified. In the new VtR format only the **child** name needs to be specified.

The new format enables the following pattern;
```
 <pb_type name="module1" num_pb="1">
  <xi:include href="module.xml" xpointer="xpointer(pb_type/child::node())" />
 </pb_type>
 <pb_type name="module2" num_pb="1">
  <xi:include href="module.xml" xpointer="xpointer(pb_type/child::node())" />
 </pb_type>
```
IE You can use an `xi:include` to import the same module under two different names.

I go through some of my ideas around making the VtR XML format more composible in the VtR issue https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/272

As VtR doesn't actually support the new format, we use a xsltproc script to translate the XML file into the old format at architecture merging time. See this part of [xmlsort.xsl](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/xml/xmlsort.xsl#L39-L51);
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/2a41dcc11eee49c3315d811e7060409c4a9fac56/common/xml/xmlsort.xsl#L39-L51

Does that make sense?

Issue #599, opened 2019-04-16
v2x solve issues to have working tests


acomodi:
Currently v2x tests (utils/vlog/tests) are being disabled as the only way to make all of them pass was to have some workarounds in the `vlog_to_pbtype.py` scripts.

The workarounds should be avoided with cleaner solutions and then we can re-enstablish the v2x tests

mithro:
@acomodi Are you working on this?

mithro:
First issue is;
```
[  0%] Generating dsp_combinational_pb_type_diff.diff
--- /usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs/build/utils/vlog/tests/dsp_combinational/dsp_combinational.pb_type.golden.xml 2019-04-19 19:50:52.160854352 -0700
+++ /usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs/build/utils/vlog/tests/dsp_combinational/dsp_combinational.pb_type.actual.xml 2019-04-19 19:50:52.068855422 -0700
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<pb_type xmlns:xi="http://www.w3.org/2001/XInclude" blif_model=".subckt dsp_combinational" name="dsp_combinational" num_pb="1">
+<pb_type xmlns:xi="http://www.w3.org/2001/XInclude" blif_model=".subckt dsp_combinational" name="BLK_IG-dsp_combinational" num_pb="1">
   <input name="a" num_pins="32"/>
   <input name="b" num_pins="32"/>
   <input name="m" num_pins="1"/>
utils/vlog/tests/dsp_combinational/CMakeFiles/dsp_combinational_pb_type_diff.dir/build.make:63: recipe for target 'utils/vlog/tests/dsp_combinational/dsp_combinational_pb_type_diff.diff' failed   
```

mithro:
[Remove `BEL_FF` (etc) naming #561](https://github.com/SymbiFlow/symbiflow-arch-defs/issues/561)

mithro:
@acomodi The first issue listed above should be solved by the pull request at https://github.com/SymbiFlow/symbiflow-arch-defs/pull/613

acomodi:
@mithro Currently I am not actively dealing with re-establishing the v2x tests as I am more focused on the tile equivalence issue. As far as I know, though, it should not require too much effort to get tests working again, so I could deal with this as well.

mithro:
[...]

PR #639, opened 2019-04-24
Re-enable some v2x tests

kgugala:
This pull request fixed a number of the tests for v2x. This includes;
 - [x] `fig41-full-adder`
 - [x] `fig42-dff`
 - [x] `dsp_inout_registered`
 - [x] `dsp_inout_registered_dualclk`

All primitives are marked with `(* blackbox *)` and all flipflops are explicitly instantiated. The multiple includes also appear to be working correctly. Output interconnect has also been fixed.

There is still some issues with ports being incorrectly being marked as clocked which affect;
 - [ ] `dsp_in_registered`
 - [ ] `dsp_partial_registered`
 - [ ] `dsp_out_registered`
 - [x] (parts of) `dsp_modes`


utils/vlog/vlog_to_pbtype.py
```diff
@@ -179,26 +185,51 @@ def make_direct_conn(ic_xml, src, dst):
...
+        processed_cells = list()
         for cname, i_of in mod.cells:
             pb_name = strip_name(i_of)
+            do_not_include = False
+            if i_of in processed_cells:
+                # do not emit xml include for every instance of multi instace cell
+                do_not_include = True
+            else:
+                processed_cells.append(i_of)
+            instance = -1
```
acomodi: worth adding a comment about the `instance` meaning?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -151,18 +151,24 @@ def make_pb_content(yj, mod, xml_parent, mod_pname, is_submode=False):
     """Build the pb_type content - child pb_types, timing and direct interconnect,
     but not IO. This may be put directly inside <pb_type>, or inside <mode>."""
 
-    def get_full_pin_name(pin):
+    def get_full_pin_name(pin, instance=-1):
```
acomodi: Probably it is worth using a macro such as `INVALID_INSTANCE = -1` globally accessible by the script to make it more understandable.

utils/vlog/tests/dsp_in_registered/golden.model.xml
```diff
@@ -1,12 +1,12 @@
...
   <xi:include href="../dsp_combinational/dsp_combinational.model.xml" xpointer="xpointer(models/child::node())"/>
-  <!-- https://docs.verilogtorouting.org/en/latest/tutorials/arch/timing_modeling/#mixed-sequential-combinational-block -->
+  <xi:include href="../fig42-dff/dff.model.xml" xpointer="xpointer(models/child::node())"/>
   <model name="dsp_in_registered">
     <input_ports>
-      <port name="a" combinational_sink_ports="out" clock="clk"/>
-      <port name="b" combinational_sink_ports="out" clock="clk"/>
-      <port name="clk" is_clock="1"/>
-      <port name="m" combinational_sink_ports="out" clock="clk"/>
+      <port combinational_sink_ports="out" name="a"/>
```
elmsfu: Are the clock attributes no longer needed because the dff model captures that dependency now?

kgugala: yes, DFF defines this https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/fig42-dff/golden.model.xml

utils/vlog/vlog_to_pbtype.py
kgugala: makes sense

kgugala: done

utils/vlog/vlog_to_pbtype.py
kgugala: I added some comments

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
```diff
@@ -12,11 +13,15 @@ module dsp_in_registered (clk, a, b, m, out);
 	reg [DATA_WIDTH/2-1:0] q_a;
```
mithro: With this approach, the `reg` below should actually be `wire`.

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
```diff
@@ -12,11 +13,15 @@ module dsp_in_registered (clk, a, b, m, out);
...
 	reg [DATA_WIDTH/2-1:0] q_b;
 	reg q_m;
-	always @(posedge clk) begin
-		q_a <= a;
-		q_b <= b;
-		q_m <= m;
+
+	genvar i;
```
mithro: The verilog `always` shown below (which is called a "process")
```verilog
always @(posedge clk) begin
   ....
end
```
**should** cause Yosys to generate `$dff` objects once the `proc` pass is run (the `proc_dff` pass is the exact one).

This makes me think that we are missing `proc` somewhere as I described in the bug....

utils/vlog/tests/dsp_in_registered/golden.model.xml
mithro: FYI - A model never includes another model. Models are low level primitives.

**However**, we want to support instantiating a whole module or to instantiate just the parts. An example would be something like;
```verilog
module ABC(...);
    PART_A a(...);
    PART_B b(...);
endmodule;
```
Then people can either go;
```verilog
// Instantiate whole module in one go...
ABC abc(...);
/// ---- or ----
// Instantiate parts of module in a compatible configuration to be placed into the ABC module
PART_A a(...);
PART_B b(...);
```

Then the `pb_type.xml` ends up looking like this;
```xml
<pb_type name="module">
  <mode name="ALL_AT_ONCE">
    <pb_type name="whole_module" blif_model=".subckt module">
    </pb_type>
  </mode>
  <mode name="INDIVIDUAL_PARTS">
    <xi:include href="part_a.pb_type.xml" />
    <xi:include href="part_b.pb_type.xml" />
    <interconnect>
       ....
    </interconnect>
  </mode>
</pb_type>
```

And as the "whole" mode is a subckt, it needs a `<model>`.

Am I making any sense?

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
mithro: Just FYI - v2x should **support** the generate method you are specifying below. We should add it as a seperate test.

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
kgugala: yes, you're right I'll change that

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
kgugala: > should cause Yosys to generate $dff objects once the proc pass is run (the proc_dff pass is the exact one).

Yosys does generate the $dff objects. It is v2x who cannot handle that (I described it here https://github.com/SymbiFlow/symbiflow-arch-defs/issues/599#issuecomment-486136155)

I think it is possible to handle generated $dff in v2x. The question is what model shall we use for those dffs and later which timings should be assigned to them?

kgugala: > Just FYI - v2x should support the generate method you are specifying below. We should add it as a seperate test.

v2x does support that - I use the feature in this PR

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
kgugala: fixed

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
mithro: Okay, lets log a bug to support the implicit version and we can land this version.

utils/vlog/tests/dsp_in_registered/golden.model.xml
mithro: Poke.

utils/vlog/tests/dsp_in_registered/golden.pb_type.xml
```diff
@@ -5,262 +5,266 @@
...
-    </direct>
-    <direct>
-      <port name="out[62]" type="input"/>
-      <port from="dsp_in_registered" name="out[62]" type="output"/>
-    </direct>
-    <direct>
-      <port name="out[63]" type="input"/>
-      <port from="dsp_in_registered" name="out[63]" type="output"/>
-    </direct>
+    <direct input="dff[0].q" name="dsp_combinational_a_0" output="dsp_combinational.a[0]"/>
```
mithro: This should be reverted right?

utils/vlog/tests/dsp_in_registered/golden.model.xml
kgugala: I see your point here, but isn't it over complicating the whole flow?

I think we can simply disable emiting include directives in model generator and leave it with with only top level (primitive) definition. All the parts that are included, were actually generated as dependencies for the currently processing one, so we already have their models definition. So a user can either instantiate them, or the "whole" one without "includes".

Does that make sense?

mithro: @kgugala Ultimately the `model.xml` should be **only** includes of other models and top level models.

mithro: Just confirming,

### Valid

```xml
<models>
  <xi:include href="../common_slice/common_slice.model.xml" xpointer="xpointer(models/child::node())" />
  <model name="blah">
  </model>
</models>
```

### Never valid

```xml
<models>
  <model name="a">
    <xi:include href="../common_slice/common_slice.model.xml" xpointer="xpointer(models/child::node())" />
  </model>
  <model name="b">
  </model>
</models>
```

kgugala:
> This change need a rebase to be able to land.

done

utils/vlog/tests/dsp_in_registered/golden.pb_type.xml
kgugala: yep, I rebased the whole PR on top of current master. It now uses new direct connections style

utils/vlog/tests/dsp_in_registered/golden.model.xml
kgugala: so the model here has correct form:
```xml
<models xmlns:xi="http://www.w3.org/2001/XInclude">
  <xi:include href="../dsp_combinational/dsp_combinational.model.xml" xpointer="xpointer(models/child::node())"/>
  <xi:include href="../fig42-dff/dff.model.xml" xpointer="xpointer(models/child::node())"/>
  <model name="dsp_in_registered">
    <input_ports>
      <port combinational_sink_ports="out" name="a"/>
      <port combinational_sink_ports="out" name="b"/>
      <port is_clock="1" name="clk"/>
      <port clock="clk" combinational_sink_ports="out" name="m"/>
    </input_ports>
    <output_ports>
      <port name="out"/>
    </output_ports>
  </model>
</models>
```
It only misses the info about clock attributes

kgugala: I added ASSOC_CLOCK attribute and fixed vlog_to_model script so the attribute is actually used. Now the model is emitted with clock attributes

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
```diff
@@ -1,22 +1,31 @@
 `include "../dsp_combinational/dsp_combinational.sim.v"
+`include "../fig42-dff/dff.sim.v"
 
 module dsp_in_registered (clk, a, b, m, out);
 	localparam DATA_WIDTH = 64;
 
+	(* CLOCK *)
```
mithro: This should be auto-detected for two reasons;
 * the signal is named `clk`.
 * the signal is driving a clock signal of flip flops.

utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
```diff
@@ -1,22 +1,31 @@
...
 	input wire clk;
+	(* ASSOC_CLOCK="clk" *)
```
mithro: This should be auto-detected from the fact it goes into a flip flop?

utils/vlog/vlog_to_model.py
```diff
@@ -207,6 +207,9 @@ def is_registered_path(tmod, pin, pout):
                 for clk in clocks:
                     if is_registered(tmod, bits, iodir, clk):
                         attrs["clock"] = clk
+                assoc_clk = tmod.net_attr(name, "ASSOC_CLOCK")
+                if assoc_clk is not None:
+                    attrs["clock"] = assoc_clk
```
mithro: This part LGTM.

kgugala:
I have cherry picked https://github.com/SymbiFlow/symbiflow-arch-defs/pull/664/commits/0349a652be87a5b5b08c3ce2d38b545c7c94b9d2 from #664. This seems to solve a lot of issues here.

All the tests should now pass, but I'm unsure about two things:

1) clock association is now added to every signal which impacts a registered signal. Seehttps://github.com/SymbiFlow/symbiflow-arch-defs/blob/206a8a969b21910248cbba0f3130c224d6f17a70/utils/vlog/tests/dsp_partial_registered/golden.model.xml. The `m` signal has a `clk` clock associated because it impacts the registered output `out`. The `m` itself is not registered though. I'm unsure if this is correct. (CC @mithro)

2) Implicit cells are resolved correctly, but there is no pb_type, or model definition of them. We should somehow include tech mapping definitions here. (CC @mithro @litghost)

litghost:
> 2. Implicit cells are resolved correctly, but there is no pb_type, or model definition of them. We should somehow include tech mapping definitions here. (CC @mithro @litghost)

I haven't been following the v2x work, what is an implicit cell and what does it represent?  Under what circumstance would an implicit cell be acceptable over an explicit blackbox definition?

kgugala:
> > 1. Implicit cells are resolved correctly, but there is no pb_type, or model definition of them. We should somehow include tech mapping definitions here. (CC @mithro @litghost)
> 
> I haven't been following the v2x work, what is an implicit cell and what does it represent? Under what circumstance would an implicit cell be acceptable over an explicit blackbox definition?

Implicit cells are e.g. $dffs not instantiated directly, but inferred by Yosys from e.g.:

```Verilog
always @ (posedge clk) begin
    q <= d
end
```
This applies also to muxes, logic gates etc

litghost:
> > > 1. Implicit cells are resolved correctly, but there is no pb_type, or model definition of them. We should somehow include tech mapping definitions here. (CC @mithro @litghost)
> > 
> > 
> > I haven't been following the v2x work, what is an implicit cell and what does it represent? Under what circumstance would an implicit cell be acceptable over an explicit blackbox definition?
> 
> Implicit cells are e.g. $dffs not instantiated directly, but inferred by Yosys from e.g.:
> 
> ```verilog
> always @ (posedge clk) begin
>     q <= d
> end
> ```
> 
> This applies also to muxes, logic gates etc

There is never a reason to ever use process blocks outside of blackbox module definitions.    If a process appear in a module that is not a blackbox, that is an illegal v2x model.  Can you provide a valid example of a process block in a module that is not a blackbox?

kgugala:
> > > > 1. Implicit cells are resolved correctly, but there is no pb_type, or model definition of them. We should somehow include tech mapping definitions here. (CC @mithro @litghost)
> > > 
> > > 
> > > I haven't been following the v2x work, what is an implicit cell and what does it represent? Under what circumstance would an implicit cell be acceptable over an explicit blackbox definition?
> > 
> > 
> > Implicit cells are e.g. $dffs not instantiated directly, but inferred by Yosys from e.g.:
> > ```verilog
> > always @ (posedge clk) begin
> >     q <= d
> > end
> > ```
> > 
> > 
> > This applies also to muxes, logic gates etc
> 
> There is never a reason to use process blocks outside of blackbox module definitions. If a process appear in a module that is not a blackbox, that is an illegal v2x model. Can you provide a valid example of a process block in a module that is not a blackbox?

I haven't seen those in a regular flow, but a lot of tests added in #316 include such files e.g. https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v

litghost:
> > > > > 1. Implicit cells are resolved correctly, but there is no pb_type, or model definition of them. We should somehow include tech mapping definitions here. (CC @mithro @litghost)
> > > > 
> > > > 
> > > > I haven't been following the v2x work, what is an implicit cell and what does it represent? Under what circumstance would an implicit cell be acceptable over an explicit blackbox definition?
> > > 
> > > 
> > > Implicit cells are e.g. $dffs not instantiated directly, but inferred by Yosys from e.g.:
> > > ```verilog
> > > always @ (posedge clk) begin
> > >     q <= d
> > > end
> > > ```
> > > 
> > > 
> > > This applies also to muxes, logic gates etc
> > 
> > 
> > There is never a reason to use process blocks outside of blackbox module definitions. If a process appear in a module that is not a blackbox, that is an illegal v2x model. Can you provide a valid example of a process block in a module that is not a blackbox?
> 
> I haven't seen those in a regular flow, but a lot of tests added in #316 include such files e.g. https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v
> 
> >  GitHub**[SymbiFlow/symbiflow-arch-defs](https://github.com/SymbiFlow/symbiflow-arch-defs)**FOSS architecture definitions of FPGA hardware useful for doing PnR device generation. - SymbiFlow/symbiflow-arch-defs

IMO, that represents some form of blackbox.  In 7-series, output registration is part of the blackbox.  For example, the BRAM blackbox (e.g. RAMB18E1_VPR) supports optional registered output.  If the plan to is to fracture the original BRAM blackbox for any reason, these new implicit processes must become new blackboxs from VPR's perspective.

I recommend for now that we make it illegal to describe processes outside of blackbox modules, and so in that case, the "registered" part of the DSP should be pushed into a new module with the``(* blackbox *)`` comment, rather than doing it implicitly.  This enables names and timing annotations to be added in the method we already have, rather than performing a potentially complicited inference, and defining naming rules.  Not to mention the mess that would be dealing with implicit include statements.

I guess I'm suggesting that an explicit approach feels more consistent and less techinically challenging here.

@mithro, do you have an design to solve the implicit process problem?  Otherwise, I'm strongly inclined to stick with explicit blackbox only.

litghost:
To add color to my suggestion, in the example https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v, there are at least 2 possible interpretations, and it's not clear how v2x could possible distiguish.

1. Those $dff's should become some part common FF (e.g. FDRE or FDSE).  This actually happens in 7-series with the input or output registration in the ILOGIC or OLOGIC site.
2. Those $dff's are part of an existing blackbox that is being fractured, and so needs a specific blackbox that can only be instanced for that part (e.g. BRAM output registration FF's).  It is worth noting that Vivado will absorb output registration into the BRAM.  So this may be a special case of 1, but I think of it differently.

kgugala:
In this PR I suggested changing this particular test to https://github.com/SymbiFlow/symbiflow-arch-defs/blob/3c29bbe5cde36112a8ad47c79129e3df3056eb11/utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v (adding explicit dffs)

@litghost As I understood you correctly, this is what you'd like to do for all of them. Is that correct?

litghost:
> In this PR I suggested changing this particular test to https://github.com/SymbiFlow/symbiflow-arch-defs/blob/3c29bbe5cde36112a8ad47c79129e3df3056eb11/utils/vlog/tests/dsp_in_registered/dsp_in_registered.sim.v (adding explicit dffs)
> 
> @litghost As I understood you correctly, this is what you'd like to do for all of them. Is that correct?
> 
> >  GitHub**[SymbiFlow/symbiflow-arch-defs](https://github.com/SymbiFlow/symbiflow-arch-defs)**FOSS architecture definitions of FPGA hardware useful for doing PnR device generation. - SymbiFlow/symbiflow-arch-defs

I believe that is the superior technical solution because:
1. It correctly defines the include relationship, so dependencies are computed accurately
2. It allows fine grained control of the blackbox instancing.

I believe there are real unsolved technical issues with an approach that supports implicit processes, and I don't see a path forward to solve them without significant work.

mithro:
@kgugala I'm rebasing this change...

mithro:
I've updated the description of this pull request to be;
-----
> This pull request fixed a number of the tests for v2x. This includes;
> 
> * [x]  `fig41-full-adder`
> * [x]  `fig42-dff`
> * [x]  `dsp_inout_registered`
> * [x]  `dsp_inout_registered_dualclk`
> 
> All primitives are marked with `(* blackbox *)` and all flipflops are explicitly instantiated. The multiple includes also appear to be working correctly. Output interconnect has also been fixed.
> 
> There is still some issues with ports being incorrectly being marked as clocked which affect;
> 
> * [ ]  `dsp_in_registered`
> * [ ]  `dsp_partial_registered`
> * [ ]  `dsp_out_registered`
> * [x]  (parts of) `dsp_modes`
-----

If it goes green, then I'll merge it.

mithro:
The last failure seems to have been;
```
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/utils/report_block_usage.py", line 51, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/utils/report_block_usage.py", line 46, in main
    block, usage[block]
AssertionError: Expect usage of block BLK-TL-SLICEM = 0, found 1
```

PR #641, opened 2019-04-24
Move yosys scripts under common/yosys.

mithro:
Part of issue #626.


CMakeLists.txt
```diff
@@ -105,9 +105,9 @@ add_custom_target(all_rrgraph_xmls)
 # Target to lint all xml files.
 add_custom_target(all_xml_lint)
 
+add_subdirectory(common)
```
elmsfu:
```suggestion
add_subdirectory(common/yosys)
```

mithro: I think it is likely we will want other things under common, so I think it makes sense to not include just `common/yosys` directly.

PR #644, opened 2019-04-25
xc7: Adding notes for SLICEM.

mithro:
* LUT-RAM configurations + descriptions.
 * Shift register configurations.


mithro:
@HackerFoo I found this when doing some cleanup. I think it is probably useful enough we should land it.

PR #646, opened 2019-04-25
make: Use symlink instead of copy.

mithro:
This enables you to edit non-generated files in the `./build` directory
and have the changes appear in your git repository. You want to do this
so that tools which parse includes will find the generated files.


mithro:
@elmsfu I'm using vim with syntastic.vim - You can see my configuration @ https://github.com/mithro/rcfiles/tree/master/vim

Issue #642, opened 2019-04-24
Make v2x produce the new composable interconnect specification again (IE Revert #637).


mithro:
The pull request https://github.com/SymbiFlow/symbiflow-arch-defs/pull/637 changed v2x to output the **old** interconnect format. This we should probably revert #637.

> This is the **old** format currently used by upstream VtR
> ```xml
> <direct input="<current_module>.<input>" name="<name>" output="<child_module>.<output>"/>
> <direct input="<child_module>.<input>" name="<name>" output="<current_module>.<output>"/>
> ```
> 
> This is the **new** format;
> ```xml
> <direct>
>   <port name="<input>" type="input"/>
>   <port from="<child_module>" name="<output>" type="output"/>
> </direct>
> <direct>
>   <port from="<child_module>" name="<input>" type="input"/>
>   <port name="<output>" type="output"/>
> </direct>
> ```
> 
> Notice how in the current VtR format the **current** module name and the **child** module name need to specified. In the new VtR format only the **child** name needs to be specified.
> 
> The new format enables the following pattern;
> ```xml
>  <pb_type name="module1" num_pb="1">
>   <xi:include href="module.xml" xpointer="xpointer(pb_type/child::node())" />
>  </pb_type>
>  <pb_type name="module2" num_pb="1">
>   <xi:include href="module.xml" xpointer="xpointer(pb_type/child::node())" />
>  </pb_type>
> ```
> IE You can use an `xi:include` to import the same module under two different names.
> 
> I go through some of my ideas around making the VtR XML format more composible in the VtR issue https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/272
> 
> As VtR doesn't actually support the new format, we use a xsltproc script to translate the XML file into the old format at architecture merging time. See this part of [xmlsort.xsl](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/xml/xmlsort.xsl#L39-L51);
> https://github.com/SymbiFlow/symbiflow-arch-defs/blob/2a41dcc11eee49c3315d811e7060409c4a9fac56/common/xml/xmlsort.xsl#L39-L51
> 

I assume that #637 was merged because other pull requests depend on it. Hence I'm logging this issue to understand what needs to be done.

mithro:
@kgugala + @acomodi - FYI

mithro:
@acomodi This is done right?

acomodi:
[...]

PR #648, opened 2019-04-25
Revert "v2x: direct connections now are in the old format"

acomodi:
This reverts commit b1fca60f8c5e47fd54aeeb683f026109a72d9570.

This PR is WIP as by looking at the description of the **new** format in https://github.com/SymbiFlow/symbiflow-arch-defs/issues/642, `vlog_to_pbtype.py` incorrectly handles the direct connections. In particular all cell related signals are always treated as outputs.

I will remove WIP as soon as the tool creates direct connections consistent with https://github.com/SymbiFlow/symbiflow-arch-defs/issues/642.

### DIFF
--- a/utils/vlog/tests/dsp_out_registered/golden.pb_type.xml
+++ b/utils/vlog/tests/dsp_out_registered/golden.pb_type.xml
@@ -1,4 +1,3 @@
-<?xml version="1.0"?>
 <pb_type xmlns:xi="http://www.w3.org/2001/XInclude" name="dsp_out_registered" num_pb="1">
   <input name="a" num_pins="32"/>
   <input name="b" num_pins="32"/>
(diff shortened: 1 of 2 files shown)


kgugala:
@acomodi CI is red, please fix this before we can merge

utils/vlog/tests/dsp_out_registered/golden.pb_type.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0"?>
```
mithro: We might want to add the `<?xml version="1.0"?>` header back...

acomodi: Ok, I missed that

PR #653, opened 2019-04-26
Modified prjxray import to add tiles tags

acomodi:
This PR is still a WIP and addresses https://github.com/SymbiFlow/symbiflow-arch-defs/issues/582. It aims at creating the `tiles` tag into the arch xml description.

For now there is only the possibility to add the tiles, without any reference to the possible equivalent ones. Once also the equivalent tiles are added to the `tile` tag I will remove the WIP.

For now I left the top level-only tags (e.g. `fc`, `pinlocations`, etc.) also in the top level `pb_types` in the `complexblocklist` to let CI go green.


acomodi:
With the most recent updates `tiles` are added to the `arch.xml` file. I have added make targets to produce a `<tile_name>.tile.xml` file that will be expanded when merging the whole architecture.

The only problem is that the tile tag creation depends on the generated `pb_types.xml` files to extract the pins of the equivalent tiles and add the corresponding mapping (this applies only to the tiles that allow other equivalent tiles). This prevents to have two equivalent tiles of the same type (e.g. `CLBLM_L` cannot have `CLBLM_R` as equivalent tile and vice-versa) as CMake will cause an error because the `pb_type.xml` target of `CLBLM_R` is not yet created.

acomodi:
> Can you explain what the purpose of the new script xc7/utils/prjxray_tile_type_import.py does versus xc7/utils/prjxray_tile_import.py? I'm not sure how they work together. Either expand the relevant docstrings, or add a .md document explaining why both are needed.

Sure, I will add a generic `README.md` and fill it with the tile import. Maybe it could be filled and completed in the future with the other scripts.

Anyway, the main reason for having a separate `tile_import` script for the `tiles` tag is that all the `pb_types.xml` have to be created first as they contain the IO pin names that are needed to fill the `equivalent_tiles` pin mapping. Therefore the creation of the `tile.xml` files depends on (possibly) all the top level `pb_types.xml`.

xc7/make/project_xray.cmake
```diff
@@ -49,7 +49,7 @@ endfunction()
 function(PROJECT_XRAY_TILE)
   set(options FUSED_SITES)
   set(oneValueArgs PART TILE)
-  set(multiValueArgs SITE_TYPES)
+  set(multiValueArgs SITE_TYPES EQUIVALENT_TILES PIN_PREFIX)
```
litghost: Please add a comment explaining how to use "EQUIVALENT_TILES" and "PIN_PREFIX"

xc7/make/project_xray.cmake
```diff
@@ -63,7 +63,7 @@ function(PROJECT_XRAY_TILE)
   get_target_property_required(PYTHON3 env PYTHON3)
   get_target_property(PYTHON3_TARGET env PYTHON3_TARGET)
 
-  set(TILE_IMPORT ${symbiflow-arch-defs_SOURCE_DIR}/xc7/utils/prjxray_tile_import.py)
+  set(TILE_TYPE_IMPORT ${symbiflow-arch-defs_SOURCE_DIR}/xc7/utils/prjxray_tile_import.py)
```
litghost: Is this a typo, and should read prjxray_tile_type_import?

xc7/utils/README.md
```diff
@@ -0,0 +1,35 @@
...
+* model.xml
+* pb_type.xml
+* tile.xml
+
+`prjxray_tile_import.py` generates the `pb_type` and `model` definitions. It takes the name of the top level tile as an argument.
+This script needs to be called for each different tile that has to be generated.
+
+`prjxray_tile_type_import.py` generates the `tile` definition. It is located in a different script as it relies on the `pb_type` definitions.
+In fact, each `tile` can be associated to other tiles which are defined `equivalent`. In this case, the `tile.xml` definition must include the pin mapping between the two equivalent tiles.
+The pin mapping is created with the information included in the `pb_type.xml` of both the tiles, hence the `tile.xml` generation can be performed only when all the `pb_type.xml` definitions have been produced.
```
litghost: Can you expand how the information from the pb_type is used?

xc7/make/project_xray.cmake
acomodi: Done.

xc7/make/project_xray.cmake
acomodi: Done.

xc7/utils/README.md
acomodi: Done.

xc7/make/project_xray.cmake
```diff
@@ -47,9 +47,34 @@ function(PROJECT_XRAY_DUMMY_SITE)
...
 
 function(PROJECT_XRAY_TILE)
+  #
+  # This function is used to create targets to generate pb_type, model and tile XML definitions.
+  #
+  # PART name of the part that is considered (e.g. artix7, zynq7, etc.)
+  # TILE name of the tile that has to be generated (e.g. CLBLM_R, BRAM_L, etc.)
+  # SITE_TYPES list of sites contained in the considered tile (e.g. CLBLM_R contains a SLICEM and SLICEL sites)
+  # EQUIVALENT_TILES list of tiles equivalent to the considered one (e.g. CLBLL_R is equivalent to CLBLM_R and CLBLM_L)
+  # PIN_PREFIX translation pattern between prefixes of two equivalent tiles. This is needed as CLBLLs and CLBLMs have different
```
litghost: When the tile split is done, no prefix translation will be required.  How will that be expressed?

xc7/make/project_xray.cmake
litghost: Thanks,

xc7/make/project_xray.cmake
```diff
@@ -99,7 +124,7 @@ function(PROJECT_XRAY_TILE)
     --output-model ${CMAKE_CURRENT_BINARY_DIR}/${TILE}.model.xml
     ${FUSED_SITES_ARGS}
     DEPENDS
-    ${TILE_IMPORT}
+    ${TILE_TYPE_IMPORT}
```
litghost: Typo?  Should still be TILE_IMPORT?

xc7/utils/prjxray_tile_type_import.py
```diff
@@ -0,0 +1,268 @@
...
+    OUT_PORT_TAG = ['output']
+
+    ##########################################################################
+    # Utility functions to create tile tag.                                  #
+    ##########################################################################
+
+    def get_port_name(port):
+        """Used to split the port in port full name, port name and pin name"""
+        split = port.split('_')
+        pin = split[-1]
```
litghost: This will not work in the general case.  Let's just rely on exact pin matches for now?

acomodi: The problem with the exact pin matches is the there are two same pins in each CLB (e.g. `CLBLL_L_A` and `CLBLL_LL_A`) and each of them has to be related to the equivalent pin in the equivalent tile (e.g. `CLBLL_L_A` must be related to `CLBLM_M_A`).
Now, with the tile split being almost complete this will not cause an issue and `get_port_name` could be avoided.

Anyways I need to check if this works in the general case, as you said.

xc7/make/project_xray.cmake
acomodi: Yes indeed, I swapped again the two of them. I think I need to find a better name to distinguish between the tiles. What about:
- `prjxray_physical_tile_import.py` instead of `prjxray_tile_type_import.py` and - 
- `prjxray_pb_type_import.py` instead of `prjxray_tile_import.py`?

xc7/utils/prjxray_tile_type_import.py
acomodi: I have checked and this does not work in the general case. With the tile split this function can be avoided, as there is not anymore a situation as for the CLBs and we can rely on exact pin matches.

acomodi:
Equivalent tiles seem to work as expected. Better placement solutions are selected as shown in the following figures (the test is the `buttons` one):

Without equivalent tiles:
without_equivalent_tile

With equivalent tiles:
with_equivalent_tiles

The placer places the blocks closer to the synthetic IOs.

xc7/make/project_xray.cmake
acomodi: I have deleted the PIN_PREFIX, given that the prefix translation is not required anymore as you correctly said. I need to delete it from the

litghost:
Please merge with master, and I'll review again.

acomodi:
> Please merge with master, and I'll review again.

Done

xc7/make/project_xray.cmake
```diff
@@ -47,9 +47,31 @@ function(PROJECT_XRAY_DUMMY_SITE)
...
+  #   PART <part_name>
+  #   TILE <tile_name>
+  #   SITE_TYPES <site_name_1> <site_name_2> ...
+  #   EQUIVALENT_TILES <equivalent_tile_name_1> <equivalent_tile_name_2> ...
+  #   SITE_AS_TILE (option)
+  #   FUSED_SITES (option)
+  #   )
+  # ~~~
+
+  set(options FUSED_SITES)
```
litghost: Merge error here

xc7/utils/prjxray_physical_tile_import.py
```diff
@@ -0,0 +1,250 @@
...
+
+import lxml.etree as ET
+
+XI_URL = "http://www.w3.org/2001/XInclude"
+
+# Macros used to select relevant port names
+PORT_TAGS = ['input', 'output', 'clock']
+
+
+def prefix_name(tile):
```
litghost: Rather than c/p, can you import these functions from prjxray_tile_import?

xc7/utils/prjxray_physical_tile_import.py
```diff
@@ -0,0 +1,250 @@
...
+        """ Used to retrieve ports from a given XML root of a pb_type."""
+        ports = set()
+
+        for child in xml:
+            if child.tag in PORT_TAGS:
+                ports.add(child.attrib['name'])
+
+        return ports
+
+    def add_pinlocations(xml, fc_xml, pin_assignments, tile):
```
litghost: Can we de-dup these functions from prjxray_tile_import?

xc7/make/project_xray.cmake
acomodi: Done

xc7/utils/prjxray_physical_tile_import.py
acomodi: Done

xc7/utils/prjxray_physical_tile_import.py
acomodi: Done

PR #597, opened 2019-04-16
WIP: Generate flip flop models using v2x

mithro:

PR #657, opened 2019-04-29
utils: v2x: handle implicit cells

### LINKED ISSUE #652: Make v2x support implicit flipflops
As I mentioned in #639;

> The verilog `always` shown below (which is called a "process")
> ```verilog
> always @(posedge clk) begin
>    ....
> end
> ```
> **should** cause Yosys to generate `$dff` objects once the `proc` pass is run (the `proc_dff` pass is the exact one).

So the issue might be these lines;
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/2a41dcc11eee49c3315d811e7060409c4a9fac56/utils/vlog/yosys/json.py#L229-L236

The code seems to be striping out cells which start with `$` like the `$dff` cells generated by the `proc` pass...

We should fix this and add tests.

kgugala:
Fixes #652

### DIFF
--- a/utils/vlog/vlog_to_pbtype.py
+++ b/utils/vlog/vlog_to_pbtype.py
@@ -155,7 +155,9 @@ def make_pb_content(yj, mod, xml_parent, mod_pname, is_submode=False):
         """Returns the name of the module relative to the pin and a boolean that indicates whether
         the module is a cell (True) or the top one (False)"""
         cname, cellpin = pin
-        if cname != mod.name:
+        if cname.startswith("$"):
+            return cname, True
+        elif cname != mod.name:
             cname = mod.cell_type(cname)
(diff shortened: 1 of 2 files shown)


mithro:
@kgugala Is there some test that should be added here?

kgugala:
Actually it does not fix the issue completely. We're still missing a pb_type definition for implicit type

kgugala:
@mithro I think dsp_in_registered is a good test for that

mithro:
Generating the `pb_type.xml` and `model.xml` for the Yosys internal latches is what I'm trying to do in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/597

My previous attempt was at https://github.com/SymbiFlow/symbiflow-arch-defs/pull/194

Issue #409, opened 2019-02-28
RAM32X1D not working


litghost:
https://github.com/SymbiFlow/yosys/pull/8 enabled synthesis of RAM32X1D, which had not been tested.  It appears that murax when synth'd with RAM32X1D does not work, but works when the RAM32X1D synthesis is removed.  Unclear if this is a caused by a bad pb_type or bad FASM matching.

- [ ] Write tests for additional DRAM modes.  
- [ ] Confirm RAM32X1D issue replicates in smaller test case
- [ ] Identify and fix issue

litghost:
This issue was seen in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/402

litghost:
Fixed by #663

PR #663, opened 2019-04-30
Remodel RAM as dual-port fracturable

HackerFoo:
Replacing {S,D}PRAM with fewer primitives that have a read and write address, but only one output. Single port RAMs are built by tying the address lines, while dual port RAMs are built from pairs.


xc7/primitives/slicem/Ndram/d_dram.pb_type.xml
```diff
@@ -22,31 +22,31 @@
...
-     <direct name="O6" input="SPRAM64.O6" output="D_DRAM.O6" />
-     <direct name="SO6" input="SPRAM64.O6" output="D_DRAM.SO6" />
+     <direct name="O6" input="DPRAM64.O" output="D_DRAM.O6" />
+     <direct name="SO6" input="DPRAM64.O" output="D_DRAM.SO6" />
    </interconnect>
  </mode>
  <mode name="32_SINGLE_PORT">
-   <xi:include href="spram32.pb_type.xml" />
+   <xi:include href="dpram32.pb_type.xml" />
    <interconnect>
```
litghost: There should have two instances of the DPRAM32 pb_type?  Or are we starting with only the first one usable?

xc7/primitives/slicem/Ndram/dpram64.pb_type.xml
```diff
@@ -4,9 +4,9 @@
 
   <!-- A1 - Read port -->
   <input  name="A"   num_pins="6" />
-  <output name="O6"  num_pins="1" />
```
litghost: Why change this name?  O6 is correct?

xc7/primitives/slicem/Ndram/dpram64.pb_type.xml
```diff
@@ -19,24 +19,24 @@
   <input  name="WA"  num_pins="6" />
   <input  name="WA7" num_pins="1" />
   <input  name="WA8" num_pins="1" />
-  <input  name="DI1" num_pins="1" />
+  <input  name="DI" num_pins="1" />
```
litghost: Why change this?  The name is correct?

HackerFoo:
This is the schematic I'm using:

ram32_schematic

xc7/primitives/slicem/Ndram/d_dram.pb_type.xml
HackerFoo: A single port RAM is just a DPRAM where A = WA, right?

xc7/primitives/slicem/Ndram/dpram64.pb_type.xml
HackerFoo: There's only one output for DPRAM32/64, so I renamed it to O. The {A-D}_DRAMs still use the names from the CLB diagram.

xc7/primitives/slicem/Ndram/dpram64.pb_type.xml
HackerFoo: There's only one DI in a DPRAM, which usually maps to DI1, except it can be DI2 in 32 bit mode.

xc7/primitives/slicem/Ndram/dpram64.pb_type.xml
litghost: If the names in the next level up use the CLB diagram names, I'm okay.

HackerFoo:
RAM{32,64}x1d doesn't route yet, I think I need to fix some pack patterns. x1s produces something, but it doesn't work yet. It seems like no RAMs are being generated when using fasm2v and Vivado.

litghost:
> This is the schematic I'm using:
> 
> ram32_schematic

I think SMALL should also affect the top WA[5] & WE part of the schematic.

litghost:
> RAM{32,64}x1d doesn't route yet, I think I need to fix some pack patterns. x1s produces something, but it doesn't work yet. It seems like no RAMs are being generated when using fasm2v and Vivado.

That sounds like a FASM annotation problem.

HackerFoo:
RAM64X1{S,D} and RAM32X1S work! One to go...

xc7/primitives/slicem/Ndram/d_dram.pb_type.xml
litghost: I was saying that there are two DPRAM32 per LUT cell

xc7/primitives/slicem/Ndram/ntemplate.N_dram.pb_type.xml
```diff
@@ -70,22 +70,23 @@
...
     <interconnect>
-      <direct name="CLK" input="{N}_DRAM.CLK" output="SPRAM64.CLK" />
-      <direct name="A" input="{N}_DRAM.A" output="SPRAM64.A" />
-      <direct name="WA7" input="{N}_DRAM.WA[6]" output="SPRAM64.WA7" />
-      <direct name="WA8" input="{N}_DRAM.WA[7]" output="SPRAM64.WA8" />
-      <direct name="{N}I" input="{N}_DRAM.{N}I" output="SPRAM64.DI1" />
-      <direct name="WE" input="{N}_DRAM.WE" output="SPRAM64.WE" />
+      <direct name="CLK" input="{N}_DRAM.CLK" output="DPRAM64.CLK" />
+      <direct name="A" input="{N}_DRAM.A" output="DPRAM64.A" />
+      <direct name="WA" input="{N}_DRAM.A[5:0]" output="DPRAM64.WA[5:0]" />
```
litghost: This is not right, other than DLUT, all DRAM's get their address lines from the D1-D6 connections.

xc7/primitives/slicem/Ndram/ntemplate.N_dram.pb_type.xml
```diff
@@ -70,22 +70,23 @@
     </metadata>
   </mode>
   <mode name="64_SINGLE_PORT">
```
litghost: We can probably get rid of the 64_SINGLE_PORT and 32_SINGLE_PORT modes, if we add the correct support for the DI muxes.

xc7/primitives/slicem/Ndram/ntemplate.N_dram.pb_type.xml
HackerFoo: This is in single port mode, where the read and write ports are tied together, just like in DLUT.

litghost: > This is in single port mode, where the read and write ports are tied together, just like in DLUT.

Have we confirmed this is correct? Specifically have we tested whether Vivado will pack two RAM64X1D's into the same slice if their WA lines vary?  I believe we ran that test and it showed that Vivado would NOT pack those together, indicated that WA is always from the DLUT input pins.

HackerFoo: I think you're right from looking at the CLB User Guide. I'll check Vivado.

HackerFoo: WA7/WA8/WE are also slice-wide, so they should be moved out into slicem.pb_type.xml

HackerFoo: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/dc45861228a81ade14f22d828856a575f9b471d9/xc7/primitives/slicem/slicem.pb_type.xml#L267-L288

It's just handled in slicem.pb_type.xml. That's probably the easiest to understand way of doing it.

HackerFoo:
This fixes #409 and #549.

HackerFoo:
> What testing has been done on the RAM64X1D mode?

RAM64X1D works now. It has been tested on hardware, and should pass all tests on Kokoro.

xc7/primitives/slicem/Ndram/d_dram.pb_type.xml
HackerFoo: There are now two DPRAM32s per LUT.

xc7/primitives/slicem/Ndram/ntemplate.N_dram.pb_type.xml
HackerFoo: I've checked that Vivado doesn't pack RAMs with independent data inputs together.

xc7/primitives/slicem/Ndram/ntemplate.N_dram.pb_type.xml
HackerFoo: Done.

xc7/primitives/slicem/slicem.pb_type.xml
```diff
@@ -299,36 +299,34 @@
...
-             required.
-             -->
-        <mux name="P_AI" input="SLICEM_MODES.DI SLICEM_MODES.BI" output="A_DRAM.PARENT_DI" />
+        <!-- DI1 inputs -->
+        <direct name="D_DI1" input="SLICEM_MODES.DI" output="D_DRAM.DI1" />
+        <mux name="CDI1MUX" input="SLICEM_MODES.DI SLICEM_MODES.CI" output="C_DRAM.DI1">
+          <metadata>
+            <meta name="fasm_mux">
+              SLICEM_MODES.DI = NULL
+              SLICEM_MODES.CI = DI1MUX.CI
```
litghost: This should probably be CLUT.DI1MUX.CI?

HackerFoo: The prefix includes `CLUT` already.

litghost: Does it?  This is one level higher than before, this is the SLICEM level?

HackerFoo:
> Please test if top_dram_n3 works on basys3 hardware. Test tool is located in tests/9-scalable_proc/utils/receiver.py, otherwise LGTM

It works (with --baud 500000).

xc7/primitives/slicem/slicem.pb_type.xml
HackerFoo: Yes, I checked the FASM when I had an error due to calling it `CDI1MUX.CI` which matches the documentation. I looked at the database and changed it to `CLUT.DI1MUX.CI`, which resulted in `...CLUT.CLUT.DI1MUX.CI`, which of course failed.

So now it's just `DI1MUX.CI`, which results in `...CLUT.DI1MUX.CI`, which works.

HackerFoo: I'm not sure where it gets `CLUT` from, though. I just checked that it worked.

litghost: Probably worth fixing, probably a FASM bug.

PR #664, opened 2019-05-01
v2x: Adding tests for clock detection

mithro:


mithro:
@kgugala PTAL

mithro:
There seems to be an issue with the following code;
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/dc45861228a81ade14f22d828856a575f9b471d9/utils/vlog/yosys/run.py#L176-L187

mithro:
Need to add more tests around this function too;
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/dc45861228a81ade14f22d828856a575f9b471d9/utils/vlog/yosys/run.py#L190-L203

mithro:
Finally, need some tests for when these modules are included in other non-blackbox modules....

utils/vlog/tests/clocks/dff_one_clock/dff_one_clock.sim.v
```diff
@@ -0,0 +1,13 @@
+/*
+ * `input wire a` should be detected as a clock because it drives the flip
+ * flop.
+ */
+module block(a, b, c);
+	input wire a;
+	input wire b;
+	output wire c;
```
kgugala: shouldn't this be a reg?

utils/vlog/tests/clocks/dff_two_clocks/dff_two_clocks.sim.v
```diff
@@ -0,0 +1,16 @@
+module block(c1, c2, a, b, c, o1, o2);
+	input wire c1;
+	input wire c2;
+	input wire a;
+	input wire b;
+	input wire c;
+	output wire o1;
+	output wire o2;
```
kgugala: shouldn't those outputs be regs?

kgugala:
In the `dff_two_clocks` example the clock association is not inferred due to logic operation inside the always block. I described it in #667

utils/vlog/tests/clocks/dff_one_clock/dff_one_clock.sim.v
mithro: Yes probably.

utils/vlog/tests/clocks/dff_two_clocks/dff_two_clocks.sim.v
mithro: Yes probably.

mithro:
I made the following change;
```patch
diff --git a/utils/vlog/tests/clocks/dff_two_clocks/dff_two_clocks.sim.v b/utils/vlog/tests/clocks/dff_two_clocks/dff_two_clocks.sim.v
index 8cab5e8c..b16f3b24 100644
--- a/utils/vlog/tests/clocks/dff_two_clocks/dff_two_clocks.sim.v
+++ b/utils/vlog/tests/clocks/dff_two_clocks/dff_two_clocks.sim.v
@@ -7,10 +7,14 @@ module block(c1, c2, a, b, c, o1, o2);
        output wire o1;
        output wire o2;
 
+       reg r1;
+       reg r2;
         always @ ( posedge c1 ) begin
-                o1 <= a | b;
+                r1 <= a | b;
         end
         always @ ( posedge c2 ) begin
-                o2 <= b | c;
+                r2 <= b | c;
         end
+       assign o1 = r1;
+       assign o2 = r2;
 endmodule
```

kgugala:
I have pushed fixes for the rest of the issues here. All the v2x tests should now pass. Let's wait for CI and check if the changes here did not break anything in the normal flow.

kgugala:
All checks are green. I'll remove the WIP tag. @mithro can you review the latest changes?

mithro:
@kgugala Rebased onto master.

utils/vlog/tests/clocks/dff_one_clock/dff_one_clock.sim.v
mithro: Fixed.

utils/vlog/tests/clocks/dff_two_clocks/dff_two_clocks.sim.v
mithro: Fixed.

mithro:
@kgugala If the CI goes green. We should merge this change.

PR #613, opened 2019-04-20
Move the type information from the names into metadata.

mithro:
This metadata was originally encoded in the name as there was nowhere else to store it. Now that VPR has metadata support, we should store the information there.

Fixes #561.

PR #665, opened 2019-05-01
Implement CLB tile split

### LINKED ISSUE #415: Split CLB tile into 2 SLICE_L or SLICE_L/SLICE_M tiles

litghost:
Fixes #415

Remaining work:
 - [x] Add docstrings
 - [x] Add comments on changes to SQL schema
 - [x] Add tests for tile split class
 - [x] Verify CI passes
 - [x] Refactor prjxray_tile_import.py


litghost:
I've replicated the CI failure on my machine, looks like VPR can no longer pack some designs post split.  Possible we've lost some CLB resources, will have to compare before and after SLICE counts and check if it makes sense.

litghost:
> I've replicated the CI failure on my machine, looks like VPR can no longer pack some designs post split. Possible we've lost some CLB resources, will have to compare before and after SLICE counts and check if it makes sense.

The number of SLICE's is correct, but resource utilization is way way up.  Will have to dig deeper.

litghost:
Before:
```
        CLBLL_R: # blocks: 194, average # input + clock pins used: 14.3041, average # output pins used: 3.30412
        CLBLL_L: # blocks: 194, average # input + clock pins used: 45.0258, average # output pins used: 8.90206
        CLBLM_R: # blocks: 156, average # input + clock pins used: 44.2564, average # output pins used: 9.21154
        CLBLM_L: # blocks: 156, average # input + clock pins used: 41.4551, average # output pins used: 9.14103
```
After:
```
        BLK-TL-SLICEL: # blocks: 1626, average # input + clock pins used: 9.35424, average # output pins used: 1.99938
        BLK-TL-SLICEM: # blocks: 465, average # input + clock pins used: 21.2688, average # output pins used: 4.36129
```

litghost:
I've double checked pack pattern usage, they are the same (as expected) between pre and post split.

xc7/utils/prjxray_arch_import.py
```diff
@@ -176,6 +179,149 @@ def create_synth_constant_tiles(
...
+
+def get_tile_prefix(conn, g, tile_pkey, site_as_tile_pkey):
+    """ Returns tile prefix of specified tile. """
+    phy_locs = get_phy_tiles(conn, tile_pkey)
+
+    prefix_tile = None
+
+    # If this tile has multiples phy_tile's, make sure only one has bitstream
+    # data, otherwise the tile split was invalid.
+    for loc in phy_locs:
```
mkurc-ant: There should never be multiple physical locations for a VPR tile. After the split a SLICE may correspond to only one physical CLB.

mkurc-ant:
@litghost I see that you started your own implementation of CLB split. I'll keep my fingers crossed!
This week I am OOO so I won't be able to do much.

From the kokoro logs I see that there are some unconnected pins in the architecture def:

> Warning 14: CEUSEDMUX[0].CE[0] unconnected pin in architecture.
Warning 15: SRUSEDMUX[0].SR[0] unconnected pin in architecture.
Warning 16: CEUSEDMUX[0].CE[0] unconnected pin in architecture.
Warning 17: SRUSEDMUX[0].SR[0] unconnected pin in architecture.
Warning 18: WE_MUX[0].CE[0] unconnected pin in architecture.
Warning 19: WE_MUX[0].WE[0] unconnected pin in architecture.
Warning 20: WE_MUX[0].WE_OUT[0] unconnected pin in architecture.
Warning 21: C_DRAM128[0].DI_OUT[0] unconnected pin in architecture.
Warning 22: B_DRAM128[0].DI_OUT[0] unconnected pin in architecture.
Warning 23: A_DRAM128[0].DI_OUT[0] unconnected pin in architecture.
Warning 24: WE_MUX[0].CE[0] unconnected pin in architecture.
Warning 25: WE_MUX[0].WE[0] unconnected pin in architecture.
Warning 26: WE_MUX[0].WE_OUT[0] unconnected pin in architecture.

I would start by checking if the SLICE model is 100% ok.

I also have some tools which might be helpful that I can share:
- "grid visualizer" - Its a python script which can load the tile grid from various sources (arch.xml, rr_graph.xml etc.) and generate a SVG file with visualization. You can inspect whether tiles are where they should be.
- "rrGraphWalk" - Allows to check whether the routing is possible by walking through the rr graph from a start node (by id) to a target node. (Consumes 11+GB of RAM just to load the graph... Didn't have time to optimize for that).

xc7/utils/prjxray_arch_import.py
litghost: If we merge two tiles, then it can be true.  For example, INT_L merged on CLBLL_L.

litghost:
Problem appears to be something is triggering VPR's "clustering_pin_feasibility_filter" to filter routes.  This is likely a result of how the new pin assignments are being generated.  I'll do a check to see if this is just a bug in VPR, which I think is likely.

litghost:
Alright, CI is now passing, but with "--clustering_pin_feasibility_filter off".  The runtime of VPR didn't actually increase that much, so maybe we just disable it for now.

xc7/utils/prjxray_create_edges.py
```diff
@@ -89,57 +70,63 @@ def add_graph_nodes_for_pins(conn, tile_type, wire, pin_directions):
...
+            SELECT pkey, node_pkey, tile_pkey
+                FROM wire WHERE wire_in_tile_pkey = ?;""",
+            (wire_in_tile_pkey, )
         )
 
-        grid_x, grid_y = c3.fetchone()
-
-        updates = []
-        values = []
+        c3 = conn.cursor()
```
HackerFoo: Could you give these cursors better names? Maybe something like `read_cursor` and `write_cursor`. It wasn't immediately clear why `c2` commits at the end and `c3` doesn't.

litghost: Good idea, I've updated the write cursors.

utils/lib/connection_database.py
```diff
@@ -199,3 +216,179 @@ def node_to_site_pins(conn, node_pkey):
     for wire_pkey, tile_pkey, wire_in_tile_pkey in c.execute(
             FIND_WIRE_WITH_SITE_PIN, (node_pkey, )):
         yield wire_pkey, tile_pkey, wire_in_tile_pkey
+
+
+def get_pin_name_of_wire(conn, wire_pkey):
+    """ Returns pin name of wire.
+
+    For unsplit tiles, this is the name of the wire.
+    For split tiles with only 1 site, this is the name of the site pin o
```
acomodi: is the final `o` a typo?

xc7/utils/prjxray_routing_import.py
```diff
@@ -104,36 +104,33 @@ def import_graph_nodes(conn, graph, node_mapping):
...
+        tile_type = tile_type[7:]
+
         pin = m.group(2)
 
-        key = (tile_type, pin)
+        (wire_in_tile_pkeys, _) = get_wire_in_tile_from_pin_name(
+            conn=conn, tile_type_str=tile_type, wire_str=pin
+        )
+
+        c.execute(
```
acomodi: Probably worth to extend the `c` variable name to something like `cursor`

xc7/utils/prjxray_tile_import.py
```diff
@@ -22,6 +22,18 @@
...
+XI_INCLUDE = "{%s}include" % XI_URL
+
+
+def prefix_name(tile):
+    """ Add tile prefix.
+
+    This avoids namespace collision when embedding a site (e.g. SLICEL) as a
+    tile.
+    """
+    return 'BLK-TL-' + tile
```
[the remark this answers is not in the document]
litghost: Your understand of why this is done is precisely correctly.  I don't want the tile name to be toggled based on whether site_as_tile, and the readability gains from #613 aren't big enough to overcome the fact that prefixes are good for avoiding namespace collisions.  I believe the ``tile`` tag will resolve the namespace overlap anyways, so this is a short term change.

utils/lib/connection_database.py
litghost: Fixed

litghost: fixed

litghost:
> LGTM. Probably xc7 CI will take a considerable amount of time as picosoc is re-enabled

CI was 75 minutes before picosoc, and ~85 minutes after, so not a big increase.   A non-trival amount of time is the initial routing import (~25-30 minutes) and we are losing a good amount of time to https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/537

If we are concerned with CI runtime, https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/537 is probably a good first place to start, along with getting VPR to consume less memory when reading the XML.  The CI is currently limited to 10 parallel builds because of RAM pressure.

litghost:
I've filled https://github.com/SymbiFlow/symbiflow-arch-defs/issues/686 to collect and track ideas on improving CI runtime.  After some of the more pressing developments (e.g. timing and ROI breakout), doing an optimization pass over the CI is probably worthwhile as it will improve developer quality of life to have the CI complete in a reasonable amount of time, and the optimizations will also likely deliver improvements to VPR itself.

acomodi:
> and we are losing a good amount of time to verilog-to-routing/vtr-verilog-to-routing#537

I believe that for starters we could just add an option to disable `check_route` in VPR in order to avoid genfasm to check the route again. It should be pretty straightforward.

litghost:
> > and we are losing a good amount of time to [verilog-to-routing/vtr-verilog-to-routing#537](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/537)
> 
> I believe that for starters we could just add an option to disable `check_route` in VPR in order to avoid genfasm to check the route again. It should be pretty straightforward.

Seems like a good idea, maybe write up a short issue on the work around in the Symbiflow VTR project.

acomodi:
Done. Here it is https://github.com/SymbiFlow/vtr-verilog-to-routing/issues/47

PR #674, opened 2019-05-03
v2x: Add support for arrays of modules

kgugala:
This PR adds support for multiple instance of pb_types in v2x tool.

The problem here is that we don't know how many instances of a pb_type we have until we merge the xmls. The pb_number value is a part of included (child) xml. When the child xml is generated we don't know how many instances of this module are going to be instantiated by the parent. Moreover, the same module can be included by many different parents, each time with potentially different pb_num.

This PR modifies the `vlog_to_pbtype` script so it, beside correct handling of multiple instances of the same pb_type, emits an xml comments defining how many times a child pb_type is instantiated by the parent. Additional tool, `update_numpbs`, scans the xml files, updates the num_pb entries with data from comments emitted by `vlog_to_pbtype` and removes the comments. This tool is called in the flow after xml_sort, but before carry chains specialization.

`multiple_instance` test has been added for regression testing of the multiple instance handling feature.

This PR also removes duplicated `xml_sort` code from `devices.cmake`


litghost:
Can you explain why num_pb is not known during v2x?  Why isn't num_pb just the number of module instances referenced in the module?

kgugala:
> Can you explain why num_pb is not known during v2x? Why isn't num_pb just the number of module instances referenced in the module?

sure, I think the easiest way to explain this will be to use an example:

Let's look at the pb_type.xml of the `dsp_in registered` (at line 7 to be more precise): https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/dsp_in_registered/golden.pb_type.xml#L7

here we simply include pb_type xml of the `dsp_combinational` module. The included xml is generated independently in a different test, and at the moment of generation it does not now who (and if) will include it. The `dsp_conditional` pb_type has the pb_num defined as "1" https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/dsp_combinational/golden.pb_type.xml#L1. This is how all the pb_type xmls were generated before this PR.

The include is resolved by the `xml_sort` which simply copies the content of included file into the parent one.

The main problem here is that in the moment when child pb_type xml is generated we don't have the knowledge about it's usage. Actually, this would be really hard to determine, as single model can be potentially used in many different places, or even act as a top level pb_type in some cases (like in the above-mentioned test the `dsp_conditional` is a top level pb in it's own tests, and is child pb in `dsp_in_registered` test).

litghost:
> > Can you explain why num_pb is not known during v2x? Why isn't num_pb just the number of module instances referenced in the module?
> 
> sure, I think the easiest way to explain this will be to use an example:
> 
> Let's look at the pb_type.xml of the `dsp_in registered` (at line 7 to be more precise): https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/dsp_in_registered/golden.pb_type.xml#L7
> 
> here we simply include pb_type xml of the `dsp_combinational` module. The included xml is generated independently in a different test, and at the moment of generation it does not now who (and if) will include it. The `dsp_conditional` pb_type has the pb_num defined as "1" https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/vlog/tests/dsp_combinational/golden.pb_type.xml#L1. This is how all the pb_type xmls were generated before this PR.
> 
> The include is resolved by the `xml_sort` which simply copies the content of included file into the parent one.
> 
> The main problem here is that in the moment when child pb_type xml is generated we don't have the knowledge about it's usage. Actually, this would be really hard to determine, as single model can be potentially used in many different places, or even act as a top level pb_type in some cases (like in the above-mentioned test the `dsp_conditional` is a top level pb in it's own tests, and is child pb in `dsp_in_registered` test).
> 
> >  GitHub**[SymbiFlow/symbiflow-arch-defs](https://github.com/SymbiFlow/symbiflow-arch-defs)**FOSS architecture definitions of FPGA hardware useful for doing PnR device generation. - SymbiFlow/symbiflow-arch-defs
> 
> >  GitHub**[SymbiFlow/symbiflow-arch-defs](https://github.com/SymbiFlow/symbiflow-arch-defs)**FOSS architecture definitions of FPGA hardware useful for doing PnR device generation. - SymbiFlow/symbiflow-arch-defs

But the parent knows, and can replace the pb_type.  This is how tile import is done, see: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/utils/prjxray_tile_import.py#L347

At this point, num_pb can be changed as needed during import.

kgugala:
yes, parent knows. And the tool this PR adds does the sam thing.
In the regular flow the xmls are generated in different parts and then merged together into arch.xml using xml_sort function. After that I update the num_pb's

litghost:
> yes, parent knows. And the tool this PR adds does the sam thing.
> In the regular flow the xmls are generated in different parts and then merged together into arch.xml using xml_sort function. After that I update the num_pb's

But the parent can set the num_pb directly, so why delay until the arch flattening and recompute at that point?

mithro:
The parent should be setting the `num_pb` in the `<pb_type>` and then include the *contents* of the `<pb_type>` using xpointer as keith showed.

kgugala:
in current flow, arch flattening is the first place where all the xmls are merged together for the very first time. Before this step we have only include statements. Some xmls are included in a few places so we cannot update the numpb in includee.

I agree this could be done earlier in the flow, but it would require quite a big refactoring the way the arch.xml is assembled. I did not want to do such big change in the flow at this point.

I need the arrays of modules functionality for timings injection and I didn't want to delay the work on the timings

litghost:
> in current flow, arch flattening is the first place where all the xmls are merged together for the very first time. Before this step we have only include statements. Some xmls are included in a few places so we cannot update the numpb in includee.
> 
> I agree this could be done earlier in the flow, but it would require quite a big refactoring the way the arch.xml is assembled. I did not want to do such big change in the flow at this point.
> 
> I need the arrays of modules functionality for timings injection and I didn't want to delay the work on the timings

As part of the include statement you can change the num_pb, if you include the other attributes, similar to https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/utils/prjxray_tile_import.py#L347  This does introduce a dependency on the v2x process on the pb_type XML output of the includes, but that is okay, because it is probably worth sanity checking that anyways.

kgugala:
OK. I'll change the `vlog_to_pbtype` so it does the inclusion with updated numpb. Indeed this will be cleaner. Thanks

kgugala:
now the includes and num_pb updates are handled by `vlog_to_pbtype`. I left the commit which removes duplicated code from devices.cmake

@mithro @litghost please review this one again

kgugala:
One comment here: instead of emiting xpointer I simply emit the contents of the included XML. I have to read the file get it's pb_type attributes so I could resolve the include instead of delaying it until arch flattening.

mithro:
@kgugala I think we should still emit the xpointer include rather than including directly.

kgugala:
> @kgugala I think we should still emit the xpointer include rather than including directly.

sure, I just modified the script to emit include with pointer

mithro:
@kgugala I did a rebase and cleanup. Will merge if CI goes green.

PR #678, opened 2019-05-04
Add RAM32M to tech map and tests

HackerFoo:


mithro:
LGTM - if @litghost says it's good, we should merge.

HackerFoo:
@litghost That looks like a large enough task to warrant a new PR, so I created a new issue: #687

HackerFoo:
Note that `dram_test` currently only tests 1 of 9 distributed RAM types.

litghost:
> @litghost That looks like a large enough task to warrant a new PR, so I created a new issue: #687

I disagree.  If someone else needs to modify the RAM32M definition, how do they know that the new definition works on hardware?  

We are actively modifying the SLICEM definition for timing, and there needs to be a way to verify that things are working.  We have some in-hardware test converge for RAM64X1D and RAMB18E1 via ``xc7/tests/bram_test``, ``xc7/tests/dram_test``, and ``tests/9-scalable_proc``.  Besides ram shifter (which is a purely manually test I believe?), I do not believe we have a way to test these changes.

HackerFoo:
Is it okay to add a test as a separate PR and merge all three (RAM32M + RAM64M + tests) at once, or would you prefer me to cancel this PR and submit a combined PR?

litghost:
> Is it okay to add a test as a separate PR and merge all three (RAM32M + RAM64M + tests) at once, or would you prefer me to cancel this PR and submit a combined PR?

Can we have a focused test on RAM32M?  I think many narrower tests, and a combined test is warranted.

xc7/tests/dram_test/dram_test_32m.v
```diff
@@ -0,0 +1,122 @@
...
+    RAM32M #(
+        .INIT_D(INIT),
+        .INIT_C(INIT),
+        .INIT_B(INIT),
+        .INIT_A(INIT)
+    ) dram(
+        .WCLK(clk),
+        .ADDRD(write_address),
+        .ADDRC(read_address),
+        .DOC(read_data),
```
litghost: Can we test DOB and DOA too?

HackerFoo: Done. I'm using RAM32M as 32x6 dual port, because it's not clear how to test quad port memory with RAM_TEST.

PR #679, opened 2019-05-04
A tool for visualizing VPR tile grid

mkurc-ant:
I wrote a tool which allows to visualize the VPR grid thus allows to verify which tiles are there and where they are. Each tile type can be assigned a different color. The resulting drawing can be stored either as SVG or a PDF (uses cairosvg to convert SVG to PDF internally).

Possible grid definition sources are:
- prjxray database (reads directly from `tilegrid.json` and/or `tileconn.json`)
- SQL connection database (reads tile grid from a table)
- Architecture definition file (`arch.xml`)
- Routing graph file (`rr_graph.xml`)

It is possible also to visualize tile connections but only when using the `tileconn.json` file.


mithro:
This is pretty cool. Could you attach some pictures to the pull request?

utils/grid_visualizer/tile_colormap.json
```diff
@@ -0,0 +1,75 @@
+[
```
mithro: I think we should end up encoding this data in the `sim.v` and/or `pb_type.xml` as `<metadata>` tags?

mkurc-ant: That's a good idea. I remember that we had a conversation about that. Maybe an issue is already posted.

mkurc-ant:
This is an example render of the tile grid read from the prjxray database. It is a screenshot of a fragment. The SVG file is about 5MB

prjxray-tile-grid

mithro:
@mkurc-ant Need a better default color scheme :-P -- Can you use the ones from https://symbiflow.github.io/prjxray-db/artix7/ ?

mithro:
Or https://symbiflow.github.io/prjtrellis-db/ECP5/LFE5U-25F/index.html would be another option.

mkurc-ant:
@mithro Yes I can. I'll work on it.

mkurc-ant:
Like that? Took these colors directly from the web page.

prjxray-tile-grid2

mithro:
Much nicer looking!

utils/grid_visualizer/tile_colormap.json
mithro: Do you want to create / find such a issue?

mkurc-ant: Couldn't find it so I added a new one: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/741

utils/grid_visualizer/tile_colormap.json
```diff
@@ -0,0 +1,450 @@
...
```
mithro: This probably should be part of the xc7 directory?

utils/grid_visualizer/grid_visualizer.py
```diff
@@ -0,0 +1,723 @@
+#!/usr/bin/env python3
```
mithro: Can you check if this works for the ice40 architecture?

utils/grid_visualizer/README.md
```diff
@@ -0,0 +1,3 @@
+# Grid visualizer
+
+This is a python utility which can visualize tile grid. It can load the grid definition from various sources (prjxray database files, arch.xml, rr_graph.xml, SQL database) and render it to a SVG/PDF file.
```
mithro: Can you add the following to this README file?

 - [ ] A little screen showing the output
 - [ ] Some sample command lines which can be run

mithro:
@mkurc-ant Looks like this pull request has conflicts, can you rebase?

mkurc-ant:
@mithro I checked and the script works on iCE40 arch. I'll clean up the color maps but I need a script to generate them which I have somewhere on my home computer. Will do it soon.

utils/grid_visualizer/README.md
mkurc-ant: I updated the README file. How should I add screens ? Simply as files ?

mithro: Yeah, just a couple of png or jpg files.

PR #680, opened 2019-05-04
A tool for traversing VPR routing graph

mkurc-ant:
This tool allows to traverse the VPR routing graph. It allows to verify if there is a possible route between two VPR nodes. It also allows to list all possible routes when starting from a given node id. The search is exhaustive.

### DIFF
--- a/utils/rr_graph_walk/rr_graph_walk.py
+++ b/utils/rr_graph_walk/rr_graph_walk.py
@@ -26,62 +26,17 @@ class RoutingGraph(object):
 
     def __init__(self, xml_file):
 
-        # Load and parse the XML
-        print("Loading XML file...")
-        parser = ET.XMLParser(remove_comments=True)
-        xml_tree = objectify.parse(xml_file, parser=parser)
-        xml_root = xml_tree.getroot()
-
         self.nodes = {}
 
         self.edges_from = {}
         self.edges_to = {}
 
-        # Import nodes
-        self._import_nodes(xml_root)
-        # Import edges
-        self._import_edges(xml_root)
+        # Load and parse the routing graph
+        self._load_and_parse_rr_graph(xml_file)
 
-        del xml_root
-        del xml_tree
-        del parser
-        gc.collect()
-
-    def _import_nodes(self, xml_root):
+    def _load_and_parse_rr_graph(self, xml_file):
         """
-        Imports nodes from XML file
-        """
-
-        print("Importing nodes...")
-
-        # Find the "rr_nodes" section
-        xml_rr_nodes = xml_root.find("rr_nodes")
-        assert (xml_rr_nodes is not None)
-
-        # Parse nodes
-        for xml_node in progressbar(xml_rr_nodes):
-            assert xml_node.tag == "node"
-
-            # Find the "loc" tag
-            xml_loc = xml_node.find("loc")
-
-            node = RoutingGraph.Node(
-                id=int(xml_node.get("id")),
-                type=xml_node.get("type"),
-                xlow=int(xml_loc.get("xlow")),
-                ylow=int(xml_loc.get("ylow")),
-                xhigh=int(xml_loc.get("xhigh")),
-                yhigh=int(xml_loc.get("yhigh"))
-            )
-
-            self.nodes[node.id] = node
-
-        del xml_rr_nodes
-        gc.collect()
-
-    def _import_edges(self, xml_root):
-        """
-        Imports edges from XML file
+        Loads the routing graph, extracts nodes and edges from the XML.
         """
 
         def append_edge(edge_dict, key_node, target_node):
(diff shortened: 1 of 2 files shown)


mithro:
How does this compare to the existing [check_graph.py](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/check_graph.py) tool?

mkurc-ant:
It writes the route to a text file so it can be analyzed later. I also wanted this tool to use node locations in the grid to tell that eg. possible route target is close to the destination. That would allow to identify what is missing and where the unroutable situation actually is. But I haven't managed to finish that yet.

utils/rr_graph_walk/rr_graph_walk.py
```diff
@@ -0,0 +1,389 @@
...
+    )
+
+    parser.add_argument(
+        "--rr_graph", type=str, required=True, help="Routing graph XML file"
+    )
+    parser.add_argument(
+        "-s", type=int, required=True, help="Start node id (required)"
+    )
+    parser.add_argument(
+        "-e",
```
acomodi: Maybe it is worth to expand the name of this argument (e.g. `-end_inode`)? (also for `-s`)

mkurc-ant: Yep, good idea. Done it.

utils/rr_graph_walk/rr_graph_walk.py
```diff
@@ -0,0 +1,394 @@
...
+        self._import_nodes(xml_root)
+        # Import edges
+        self._import_edges(xml_root)
+
+        del xml_root
+        del xml_tree
+        del parser
+        gc.collect()
+
+    def _import_nodes(self, xml_root):
```
litghost: Consider using a streaming parser to avoid the large memory cost?  Might even be faster?

litghost: https://docs.python.org/2/library/xml.etree.elementtree.html#xml.etree.ElementTree.iterparse

utils/rr_graph_walk/rr_graph_walk.py
litghost: You need to update the ``args.s``  -> ``args.start_inode``, etc.  The script does not work right now.

utils/rr_graph_walk/rr_graph_walk.py
mkurc-ant: @litghost That was something I planned but didn't have time for. Now it is done. The runtime is increased by about 2min but the memory usage is dramatically reduced:

```
Non-iterative XML parsing:
	Elapsed (wall clock) time (h:mm:ss or m:ss): 13:30.93
	Maximum resident set size (kbytes): 17543404

Iterative XML parsing:
	Elapsed (wall clock) time (h:mm:ss or m:ss): 15:15.95
	Maximum resident set size (kbytes): 2244860
```

utils/rr_graph_walk/rr_graph_walk.py
mkurc-ant: Fixed.

utils/rr_graph_walk/rr_graph_walk.py
mithro: @mkurc-ant Is that 15 minutes runtime?

utils/rr_graph_walk/rr_graph_walk.py
```diff
@@ -0,0 +1,398 @@
...
+
+        node = self.nodes[node_id]
+
+        return "%s:%d [%d,%d,%d,%d]" % (
+            node.type, node.id, node.xlow, node.ylow, node.xhigh, node.yhigh
+        )
+
+    def verify_route(self, route, walk_direction):
+        """
+        Checks wheter a given route is valid
```
litghost: For the docstrings, document what the arguments are and their purpose.

utils/rr_graph_walk/rr_graph_walk.py
mkurc-ant: Yes it is. I implemented the depth first search algorithm which performs the full search over the rr graph. I also have a C++ implementation which I did expecting performance issues with Python. And the C++ version is about 2x faster.

utils/rr_graph_walk/rr_graph_walk.py
mkurc-ant: Done.

utils/rr_graph_walk/rr_graph_walk.py
litghost: Routers are typically breadth first search (https://en.wikipedia.org/wiki/Maze_solving_algorithm#Shortest_path_algorithm) not depth first.

mkurc-ant: But still I wanted this tool also to be able to find all possible routes from/to a node. And I found it easier to implement the tracking of the route using depth first search.

Breadth first search would explore broader area of the graph faster but I don't think it would hit a specific target node any faster.

PR #685, opened 2019-05-06
Support for FASM attributes for V2X

mkurc-ant:
This PR adds support for FASM features to the V2X. Every attribute of the top level module which name starts with `FASM_` will be converted to a corresponding entry in the metadata section. If no such features are present then the metadata tag is not generated at all.

Todo:
- [ ]  ~Add support for FASM prefix on arrays of modules~
- [x]  Add support for FASM LUT annotations
- [x]  Add support for FASM annotations on `<direct>` within `<interconnect>`
- [ ]  ~Add support for FASM mux annotations on `<mux>` within `<interconnect>`~
- [x]  Add support for FASM features on `<mode>`
- [x]  Add support for FASM parameters on primative `<pb_type>`'s


mkurc-ant:
@mithro @litghost My first question: What do you mean by "Add support for FASM prefix on arrays of modules" ? Could you give an example how this should work ?

litghost:
> @mithro @litghost My first question: What do you mean by "Add support for FASM prefix on arrays of modules" ? Could you give an example how this should work ?

https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/primitives/ff/ff.pb_type.xml#L112

In this example, num_pb = 8, and fasm_prefix has 8 entries.  We should probably support tagging each of the instances with a prefix, and then v2x gathers them as needed, and outputs fasm_prefix to accommodate the output.

mkurc-ant:
@litghost Ok, I got it now, thanks.

mkurc-ant:
@litghost I also have an issue with adding muxes to interconnects.

The first problem is how to define that mux in verilog. If it is 2:1 mux then the solution is straightforward - Yosys infers the mux and adds a `$mux` cell which can then be used by v2x. We can add attributes to the output net of that mux which will define its FASM features. For example:

```
module fasm_features_ic_mux_2to1(
  input I0,
  input I1,
  output O,
);

wire I0;
wire I1;

wire O_AND;
wire O_XOR;

// Child cells
and_gate gate1 (I0, I1, O_AND);
xor_gate gate2 (I0, I1, O_XOR);

wire S; // This is a dummy signal

// The mux
(* FASM_MUX = "insert features here" *)
wire O = S ? O_AND : O_XOR;

endmodule
```
But when we have eg 4:1 mux then, according to my experiments, Yosys doesn't detect that is a mux (it should at least infer the `$pmux` internal cell). Insteat, it adds a bunch of logic, comparators and `$mux`. This is hard to infer in v2x.

So instead I propose this: We define a dummy black box verilog module with a known name (eg. `v2x_mux`) which has parameters and attributes. Then we instantiate that mux in a place where needed. Then the v2x searches for that specific primitives and converts them to interconnect muxes with fasm attributes. For example:

```
module fasm_features_ic_mux_4to1(
  input I0,
  input I1,
  output O,
);

wire I0;
wire I1;

wire O_AND;
wire O_XOR;

// Child cells
and_gate gate1 (I0, I1, O_AND);
xor_gate gate2 (I0, I1, O_XOR);

// The 4:1 mux
v2x_mux #(.N(4)) any_name
(
.I({I1, I0, O_XOR, O_AND}),
.O(O)
);

endmodule
```

And the mux blackbox:

```
(* blackbox *)
module v2x_mux #
(
parameter N = 2
)
(
input [N-1:0] I,
output O
);
endmodule
```

Whad do you think ?

mithro:
The `tests/common` directory might be better as a `tests/basic` or a `tests/simple` ?

litghost:
> @litghost I also have an issue with adding muxes to interconnects.
>
> Whad do you think ?

This all sounds good.  I strongly recommend that you break up the different FASM features into small PR's so we can address them in smaller increments.  For example, the FASM MUX change should be seperate from the FASM prefix handling.

utils/vlog/tests/fasm_features/fasm_features.sim.v
```diff
@@ -0,0 +1,13 @@
+(* blackbox *)
+(* FASM_PREFIX = "MY_CELL" *)
+(* FASM_FEATURES = "USE_SOMETHING" *)
+(* FASM_LUT = "XLUT.INIT[7:0] = XLUT[0]; XLUT.INIT[15:8] = XLUT[1]" *)
```
mithro: `FASM_LUT` should only be placed on a `xxx.sim.v` that has `(* CLASS="lut" *)`?

https://github.com/SymbiFlow/symbiflow-arch-defs/blob/38d7f793ee0717dc068f35dff82583c67b51112a/utils/vlog/vlog_to_pbtype.py#L12-L13

utils/vlog/tests/fasm_features/fasm_features.sim.v
```diff
@@ -0,0 +1,13 @@
+(* blackbox *)
+(* FASM_PREFIX = "MY_CELL" *)
```
mithro: Do you think we could just use the module name for the `FASM_PREFIX`?

mkurc-ant:
@mithro The `tests/common` does not contain any actual tests but rather some verilog modules + XMLs neede for other tests. Hence the name. Maybe it is better as `tests/assets` ?

mkurc-ant:
@litghost Makes sense to break these things up. From my understanding adding v2x FASM support for arrays and muxes requires additional effort and can be done in separate PRs. So if it is ok then we can consider tasks for this PR as complete.

mithro:
> @mithro The `tests/common` does not contain any actual tests but rather some verilog modules + XMLs neede for other tests. Hence the name. Maybe it is better as `tests/assets` ?

But they have XML models committed?

utils/vlog/tests/fasm_features/fasm_features.sim.v
litghost: This will almost never work.  Blackbox names are usually generic (thinking FDRE or RAMB18E1) and prefixes are generally instance specific (think AFF, BFF).

utils/vlog/tests/common/and_gate.sim.v
```diff
@@ -0,0 +1,10 @@
+(* blackbox *)
+module and_gate(
+  input  wire I0,
+  input  wire I1,
+  output wire O
+);
```
mithro: We use the following verilog style;
```verilog
module and_gate(I0, I1, O);
  input wire I0;
  input wire I1;
  output wire O;

 assign O = I0 & I1;

endmodule
```

It makes it easier to attach the parameters to the ports which you should add;
```verilog
module and_gate(I0, I1, O);
  input wire I0;
  input wire I1;

 (* DELAY_CONST_I0="300e-12" *)
 (* DELAY_CONST_I1="300e-12" *)
  output wire O;

 assign O = I0 & I1;

endmodule
```

utils/vlog/vlog_to_pbtype.py
```diff
@@ -41,6 +41,14 @@
 
     - `(* PORT_CLASS="clock" *)` : specify the VPR "port_class"
 
+    - `(* FASM_xxxx *)` : All attributes with names starting from "FASM_" will
+        be converted do metadata entries with corresponding lowercase names.
+
+    - `(* FASM_xxxx_mmmm *)` : If there are modes and an attribute name begins
+       with "FASM" and ends with an upper case mode name, then it is applied
+       only for the matching mode. The base attribute name with mode name
+       removed is converted to lowercase "fasm_xxxx"
```
mithro: We should just apply the `(* FASM_xxx *)` only inside the given mode using a `generate` statement.

litghost:
> * For given `(* CLASS=XXX *)` I think we should add FASM support directly (rather than needing `FASM_xxx` attributes.

I don't believe this is not possible, at least not without additional explanation.

mithro:
> @litghost I also have an issue with adding muxes to interconnects.
> 
> The first problem is how to define that mux in verilog. If it is 2:1 mux then the solution is straightforward - Yosys infers the mux and adds a `$mux` cell which can then be used by v2x. We can add attributes to the output net of that mux which will define its FASM features. For example:
> 
> ```
> module fasm_features_ic_mux_2to1(
>   input I0,
>   input I1,
>   output O,
> );
> 
> wire I0;
> wire I1;
> 
> wire O_AND;
> wire O_XOR;
> 
> // Child cells
> and_gate gate1 (I0, I1, O_AND);
> xor_gate gate2 (I0, I1, O_XOR);
> 
> wire S; // This is a dummy signal
> 
> // The mux
> (* FASM_MUX = "insert features here" *)
> wire O = S ? O_AND : O_XOR;
> 
> endmodule
> ```
> 
> But when we have eg 4:1 mux then, according to my experiments, Yosys doesn't detect that is a mux (it should at least infer the `$pmux` internal cell). Insteat, it adds a bunch of logic, comparators and `$mux`. This is hard to infer in v2x.

FYI This could be fixed by a combination of some of the following yosys passes;
 * http://www.clifford.at/yosys/cmd_proc_mux.html
 * http://www.clifford.at/yosys/cmd_lut2mux.html
 * http://www.clifford.at/yosys/cmd_muxcover.html
 * http://www.clifford.at/yosys/cmd_opt_muxtree.html
 * http://www.clifford.at/yosys/cmd_pmuxtree.html

> So instead I propose this: We define a dummy black box verilog module with a known name (eg. `v2x_mux`) which has parameters and attributes. Then we instantiate that mux in a place where needed. Then the v2x searches for that specific primitives and converts them to interconnect muxes with fasm attributes. For example:
> 
> ```
> module fasm_features_ic_mux_4to1(
>   input I0,
>   input I1,
>   output O,
> );
> 
> wire I0;
> wire I1;
> 
> wire O_AND;
> wire O_XOR;
> 
> // Child cells
> and_gate gate1 (I0, I1, O_AND);
> xor_gate gate2 (I0, I1, O_XOR);
> 
> // The 4:1 mux
> v2x_mux #(.N(4)) any_name
> (
> .I({I1, I0, O_XOR, O_AND}),
> .O(O)
> );
> 
> endmodule
> ```
> 
> And the mux blackbox:
> 
> ```
> (* blackbox *)
> module v2x_mux #
> (
> parameter N = 2
> )
> (
> input [N-1:0] I,
> output O
> );
> endmodule
> ```
> 
> Whad do you think ?

You are correct we should create black boxes for the muxes. But we should use `mux_gen.py` to generate a `sim.v` file with the correct annotations. As these are black boxes, they will then be easy to see.

utils/vlog/tests/common/and_gate.sim.v
mkurc-ant: Ok, I fixed the style in all new tests

utils/vlog/vlog_to_pbtype.py
mkurc-ant: I am not sure if I understand your idea correctly. You mean apply `(* FASM_xxxx *)` only inside a generate statement? I don't think it is possible as attributes declared there apply not to the top level module but to child modules, nets etc.

As far as I understand we want to be able to specify different FASM features for the top level module for each mode, provided that it has modes. Right now it works just like that.

Could you see the `fasm_features_modes` test and tell if it is the correct behavior?

mkurc-ant:
@mithro Ok. If we are going to use the `mux_gen.py` then my concerns about how to define muxes are not relevant anymore.

mkurc-ant:
@mithro And regarding the `test/common` folder: Both verilog and XMLs placed there are used by some other tests related to FASM feature support. I needed a sub module that I could instantiate within modules that I test. The V2X automatically includes XMLs for child module instances so it requires those files. But they are not any test case themselves.

mkurc-ant:
@litghost @mithro I've been thinking more about muxes. The `mux_gen.py` is a solution to generate a mux blackbox. But in order to add support for `<mux>`es to the `<interconnect>` something else is required. I've created an issue https://github.com/SymbiFlow/symbiflow-arch-defs/issues/697

utils/vlog/tests/fasm_features/fasm_features.sim.v
mkurc-ant: Right now the v2x does not check that. It just passes all attributes starting with FASM_ to the `<metadata>` tag. Its up to verilog code author to use them properly.

Shall I add a kind of check so that nobody will be able to specify `FASM_LUT` on a module which do not have the `CLASS="lut"` set ?

litghost: It goes both ways actually.  CLASS="lut" ALWAYS requires FASM LUT annotations, or genfasm will fail.

mithro:
@mkurc all muxes should be generated with `mux_gen` and then included into the `sim.v` file. v2x should then never see any `$mux` or similar cells or be generating anything apart from direct interconnects.

mithro:
I opened https://github.com/SymbiFlow/symbiflow-arch-defs/issues/701 to track adding support of FASM output to muxgen.

mkurc-ant:
@mithro I agree that we should not infer muxes and generate them explicitly via the `mux_gen`. But generating only direct connections in V2X is not enough as they do not cover `<mux>` under the `<interconnect>`. I created https://github.com/SymbiFlow/symbiflow-arch-defs/issues/697 which describes my approach to the problem

mkurc-ant:
@litghost I added merging of metadata when including pb_types. However this works blindly by just concatenating strings and separating them with space.

litghost:
> @litghost I added merging of metadata when including pb_types. However this works blindly by just concatenating strings and separating them with space.

Just consider it an error to have duplicate metadata tags, rather than a potentially confusing behavior, that is almost certainly wrong.

utils/vlog/tests/fasm_features_child/fasm_features_child.sim.v
```diff
@@ -0,0 +1,24 @@
...
+  input  wire I0;
+  input  wire I1;
+  output wire O0;
+  output wire O1;
+  output wire O2;
+
+ (* FASM_PREFIX = "THIS_IS_THE_AND_GATE" *)
+ and_gate gate1(I0, I1, O0);
+
+ (* FASM_FEATURES = "FEATURE_A FEATURE_B" *)
```
litghost: FASM features applies to modes and blackbox's only, not to blackbox instances.

litghost:
@mkurc-ant:  Can I ask you to stop writing code and write out a design document for how you are planning to map https://docs.verilogtorouting.org/en/latest/utils/fasm/ to Verilog annotations?   The examples of FASM_FEATURES and FASM_LUT demonstrates that you don't understand what is being done here, and I think we need to take a step back and actually design these annotations.

I strongly recommend that you draw from the existing examples in xc7 and ice40, as those are the actual XML's

mithro:
@mkurc-ant There is a doc at https://docs.google.com/document/d/1ngck6LGVTXm6HDrVqWeEGdEELrKPWXVaeejkXurgfl4/edit#heading=h.edp8qrijjxwn which it attempting to do exactly what @litghost is looking for.... It is still a work in progress.

litghost:
> @mkurc-ant There is a doc at https://docs.google.com/document/d/1ngck6LGVTXm6HDrVqWeEGdEELrKPWXVaeejkXurgfl4/edit#heading=h.edp8qrijjxwn which it attempting to do exactly what @litghost is looking for.... It is still a work in progress.

I'd like a more focused document on the annotations themselves and edge cases.  I'd like to work through the various format issues independent other concerns.  Once we've iterated on the annotation design, and have confidence in an implementation, we can go back and annotate the Encyclopedia.

mkurc-ant:
@mithro @litghost Ok, I'll prepare such a document.

The test cases I provided test the functionality of v2x. Not whether such a functionality is correct in terms of legal FASM metadata specification.

utils/vlog/tests/fasm_features_child/fasm_features_child.sim.v
mkurc-ant: This is not true when specifying FASM features for arrays. Those are not handled in this PR but the entry work is done. I will remove it from the test case. But I'd rather not remove support for FASM features on blackbox instances from the code.

mkurc-ant:
@mithro @litghost I prepared the document you requested: https://docs.google.com/document/d/1J8lLT4SzlK_56-vBfz0f2DaNLHPn9rABUB0Fu1ljF4M

mkurc-ant:
The CI fails but I can't download the log. The only message I see when trying to do so is: 
`Test Fusion Url (This url will not work until build finishes): https://fusion.corp.google.com/runanalysis/info/prod%3Afoss-fpga-tools%2Fsymbiflow-arch-defs%2Fpresubmit/prod%3Afoss-fpga-tools%2Fsymbiflow-arch-defs%2Fpresubmit/KOKORO/2907333f-d2cf-4e0f-a352-d7d0c4f2db46/1557754581406/prod%3Afoss-fpga-tools%2Fsymbiflow-arch-defs%2Fpresubmit%20Build%20%23389/Targets`

Strange...

mkurc-ant:
I am closing the PR. Once we agree on the FASM in V2X functionality I will create a new one.

mithro:
Just FYI - I'll land an updated version of this after https://github.com/SymbiFlow/symbiflow-arch-defs/pull/703

PR #688, opened 2019-05-06
file_targets.cmake: correctly handle included files

kgugala:
If add_file_target was called with relative path, the includes were resolved from current source dir. The relative offset was not taken into account resulting in `get_target_property() called with non-existent target error`


kgugala:
> To have the verilog construction be parallel with the XML construction, does it make sense to modify deps_verilog? Just a thought.

good idea. I have modified this PR so it changes `deps_verilog` so it produces absolute paths

PR #693, opened 2019-05-07
iCE40: bel timing import

elmsfu:
* add `python-sdf-timing` to third_party
* remove values from pbtypes available from icestorm
* add tool to import values from sdf into merged xml


mithro:
```
make[3]: *** Deleting file 'ice40/devices/top-routing-virt/arch.timing.xml'
[ 50%] Generating arch.timing.xml
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-ice40/ice40/utils/ice40_import_bel_timing.py", line 6, in <module>
    from sdf_timing.sdfparse import parse as sdf_parse
ModuleNotFoundError: No module named 'sdf_timing'
ice40/devices/top-routing-virt/CMakeFiles/file_ice40_devices_top-routing-virt_arch.timing.xml.dir/build.make:62: recipe for target 'ice40/devices/top-routing-virt/arch.timing.xml' failed
make[3]: *** [ice40/devices/top-routing-virt/arch.timing.xml] Error 1
CMakeFiles/Makefile2:93205: recipe for target 'ice40/devices/top-routing-virt/CMakeFiles/file_ice40_devices_top-routing-virt_arch.timing.xml.dir/all' failed
make[2]: *** [ice40/devices/top-routing-virt/CMakeFiles/file_ice40_devices_top-routing-virt_arch.timing.xml.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....
[ 50%] Generating ../env/conda/bin/iceprog, ../env/conda/bin/icebox_hlc2asc, ../env/conda/bin/icebox_vlog, ../env/conda/bin/icepack, ../env/conda/bin/icetime, ../env/conda/bin/icebox
make[4]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
[ 50%] Built target thirdparty_icestorm
CMakeFiles/Makefile2:1523: recipe for target 'CMakeFiles/all_ice40.dir/rule' failed
make[1]: *** [CMakeFiles/all_ice40.dir/rule] Error 2
Makefile:680: recipe for target 'all_ice40' failed
make: *** [all_ice40] Error 2
```

common/cmake/devices.cmake
```diff
@@ -242,6 +242,8 @@ function(DEFINE_DEVICE_TYPE)
   #   DEVICE_TYPE <device_type>
   #   ARCH <arch>
   #   ARCH_XML <arch.xml>
+  #   [SCRIPT_OUTPUT_NAME]
+  #   [SCRIPTS]
```
mithro: This needs some documentation updating right?

ice40/utils/ice40_import_bel_timing.py
```diff
@@ -0,0 +1,316 @@
...
+import argparse
+from sdf_timing.sdfparse import parse as sdf_parse
+import lxml.etree as ET
+import re
+import logging
+from collections import namedtuple
+
+
+class pin_map(namedtuple('pin_map', 'sdf arch is_clk')):
+    """
```
mithro: Missing docstring?

ice40/utils/ice40_import_bel_timing.py
```diff
@@ -0,0 +1,316 @@
...
+
+import argparse
+from sdf_timing.sdfparse import parse as sdf_parse
+import lxml.etree as ET
+import re
+import logging
+from collections import namedtuple
+
+
+class pin_map(namedtuple('pin_map', 'sdf arch is_clk')):
```
mithro: `pin_map` isn't the normal formatting for a class? `PinMap` would be the normal name?

ice40/utils/ice40_import_bel_timing.py
```diff
@@ -0,0 +1,316 @@
...
+            ]
+        ),
+    'SB_LUT4':
+        (
+            'LogicCell40', [
+                pin_map('in0', 'I0', False),
+                pin_map('in1', 'I1', False),
+                pin_map('in2', 'I2', False),
+                pin_map('in3', 'I3', False),
+                pin_map('lcout', 'O', False)
```
mithro: Any reason not to modify the names to match in either the sdf file or in the sim/xml files?

ice40/utils/ice40_import_bel_timing.py
```diff
@@ -0,0 +1,316 @@
...
+                pin_map('MASK[13]', 'MASK[13]', False),
+                pin_map('MASK[14]', 'MASK[14]', False),
+                pin_map('MASK[15]', 'MASK[15]', False),
+            ]
+        ),
+    'SB_CARRY': ('ICE_CARRY_IN_MUX', [])
+}
+
+
+def get_scale(timescale):
```
mithro: Should this be part of the python-sdf-parser library? It seems like something every user of that library will need to do?

ice40/utils/ice40_import_bel_timing.py
```diff
@@ -0,0 +1,316 @@
...
+        return str(max_del * scale)
+
+    # TODO: need to take max across negedge and posedge
+
+    # remove all existing tags and warn on them
+
+    # iterate over existing arch and update/add delay tags
+    #root = tree.getroot()
+    for el in tree.iter():
+        if el.tag == 'pb_type':
```
mithro: Save indentation level using;
```
    if el.tag != 'pb_type':
         continue
```

common/cmake/devices.cmake
elmsfu: Added some

ice40/utils/ice40_import_bel_timing.py
elmsfu: Added

ice40/utils/ice40_import_bel_timing.py
elmsfu: Done

ice40/utils/ice40_import_bel_timing.py
elmsfu: That had crossed my mind, but felt like there are likely cases that we still need this and need to denote if it's a clock.

ice40/utils/ice40_import_bel_timing.py
elmsfu: Agree. I'll make a PR to do that.

ice40/utils/ice40_import_bel_timing.py
elmsfu: Realized I can use `tree.iter('pb_type')`

elmsfu:
> How is this being tested?
I've tested that a few designs still work on hardware. I've been looking at running the fpga-perf-tool and seeing if this has any impact on the results.

> Do you have an example of the output and how it is patched?

Here's the diff between before (arch.unique.xml) and after (arch.timing.xml) https://gist.github.com/elmsfu/7d49d6758c7be4961a31af01be0408d5

mithro:
A diff with context would be more useful :-)

common/cmake/devices.cmake
```diff
@@ -242,17 +242,23 @@ function(DEFINE_DEVICE_TYPE)
...
   #   )
   # ~~~
   #
   # Defines a device type with the specified architecture.  ARCH_XML argument
   # must be a file target (see ADD_FILE_TARGET).
   #
+  # optional SCTIPTs can be run after the standard flow to augment the
```
litghost: Spelling of scripts

common/cmake/devices.cmake
```diff
@@ -242,17 +242,23 @@ function(DEFINE_DEVICE_TYPE)
...
+  # final arch xml. The name and script must be provided and each
+  # script will be run as `cmd input > output`
```
litghost: Maybe consider running it as:

```
cmd < input > output
```

So that scripts don't need to be away of files at all, and just use stdin/stdout?

ice40/devices/top-routing-virt/CMakeLists.txt
```diff
@@ -1,4 +1,44 @@
...
 
 add_file_target(FILE arch.xml SCANNER_TYPE xml)
-define_device_type(DEVICE_TYPE top-routing-virt ARCH ice40 ARCH_XML arch.xml)
+
+get_target_property_required(CONDA_DIR env CONDA_DIR)
+get_target_property_required(PYTHON3 env PYTHON3)
+
+set(ICEBOX_TIMING ${symbiflow-arch-defs_SOURCE_DIR}/third_party/icestorm/icefuzz/timings.py)
+set(CLEAN_ICEBOX_TIMING cleaned_timing.txt)
+set(TIMING_TXT_FILE ${CONDA_DIR}/share/icebox/timings_hx1k.txt)
```
litghost: Rather than using CONDA_DIR directly, I recommend adding an icebox directory CACHE variable that defaults to ${CONDA_DIR}/share/icebox/timings_hx1k.txt, etc.

ice40/icestorm.cmake
```diff
@@ -0,0 +1,117 @@
...
+  define_arch(
+    ARCH ice40
+    YOSYS_SCRIPT ${symbiflow-arch-defs_SOURCE_DIR}/ice40/yosys/synth.tcl
+    DEVICE_FULL_TEMPLATE \${DEVICE}-\${PACKAGE}
+    VPR_ARCH_ARGS
+      --clock_modeling route
+      --allow_unrelated_clustering off
+      --target_ext_pin_util 0.7
+      --router_init_wirelength_abort_threshold 2
+      --congested_routing_iteration_threshold 0.8
```
elmsfu: @litghost I ended up adding this to get more consistent routing completion.

litghost: Seems reasonable to me.

CMakeLists.txt
```diff
@@ -115,7 +115,7 @@ add_thirdparty_package(
   BUILD_INSTALL_COMMAND "cd ${symbiflow-arch-defs_SOURCE_DIR}/third_party/python-sdf-timing && ${PYTHON3} setup.py develop"
   NO_EXE
   DEPENDS ${PYTHON3_TARGET}
-)
+  )
```
litghost: Stray format change?

common/cmake/devices.cmake
```diff
@@ -346,7 +355,7 @@ function(DEFINE_DEVICE_TYPE)
   set(ARCH_SCHEMA ${symbiflow-arch-defs_SOURCE_DIR}/common/xml/fpga_architecture.xsd)
   xml_lint(
     NAME ${DEFINE_DEVICE_TYPE_ARCH}_${DEFINE_DEVICE_TYPE_DEVICE_TYPE}_arch_lint
-    FILE ${FINAL_OUTPUT}
+    FILE ${FINAL_FILE}
```
litghost: This should be FINAL_OUTPUT?  And xml_lint should invoke get_file_location + append_file_dependency?

common/cmake/devices.cmake
```diff
@@ -355,7 +364,8 @@ function(DEFINE_DEVICE_TYPE)
     ${DEFINE_DEVICE_TYPE_DEVICE_TYPE}
     PROPERTIES
     DEVICE_MERGED_FILE ${CMAKE_CURRENT_SOURCE_DIR}/${FINAL_OUTPUT}
-  )
+    )
+  add_dependencies(${DEFINE_DEVICE_TYPE_DEVICE_TYPE} ${FINAL_TARGET})
```
litghost: Use "append_file_dependencies(${FINAL_OUTPUT})" over add_dependencies.

common/cmake/devices.cmake
```diff
@@ -884,6 +894,8 @@ function(ADD_FPGA_TARGET)
   set(OUT_ROUTE ${OUT_LOCAL}/${TOP}.route)
 
   set(VPR_DEPS "")
+  list(APPEND VPR_DEPS ${OUT_EBLIF})
```
litghost: Use "append_file_dependencies(${OUT_EBLIF_REL})" over this

ice40/devices/top-routing-virt/CMakeLists.txt
```diff
@@ -1,4 +1,45 @@
...
+get_target_property_required(PYTHON3_TARGET env PYTHON3_TARGET)
+get_target_property_required(ICESTORM_TARGET env ICESTORM_TARGET)
+
+set(ICEBOX_TIMING ${symbiflow-arch-defs_SOURCE_DIR}/third_party/icestorm/icefuzz/timings.py)
+set(CLEAN_ICEBOX_TIMING cleaned_timing.txt)
+set(TIMING_TXT_FILE ${ICEBOX_SHARE}/timings_hx1k.txt)
+set(SDF_FILE timings_hx1k.sdf)
+
+add_custom_command(
+  OUTPUT ${CLEAN_ICEBOX_TIMING}
```
litghost: Rather than two add_custom_command, might be better to just have one?

ice40/devices/top-routing-virt/CMakeLists.txt
```diff
@@ -1,4 +1,45 @@
...
+  )
+
+add_custom_command(
+  OUTPUT ${SDF_FILE}
+  COMMAND ${PYTHON3} ${ICEBOX_TIMING} -t ${CLEAN_ICEBOX_TIMING} -s > ${SDF_FILE}
+  DEPENDS ${ICEBOX_TIMING} ${CLEAN_ICEBOX_TIMING} ${PYTHON3} ${PYTHON3_TARGET}
+  )
+
+add_custom_target(
+  sdf_target
```
litghost: Maybe ice40_sdf_target?

ice40/devices/top-routing-virt/CMakeLists.txt
elmsfu: You mean combine these? I think that just makes it more opaque.

common/cmake/devices.cmake
elmsfu: That appends to a list, but I'd still need to add the dependency right?

common/cmake/devices.cmake
elmsfu: Are you suggesting that I modify `xml_lint()` to change how the FILE parameter is expected?

ice40/devices/top-routing-virt/CMakeLists.txt
elmsfu: good clarification.

CMakeLists.txt
elmsfu: yep, snuck in rebasing and merging

ice40/devices/top-routing-virt/CMakeLists.txt
litghost: If you don't combine these, then you have to an introduce a target between these two add_custom_command.  Parallel CMake will attempt to do the grep step more than once if multiple targets need the sdf_target.  If you want to keep them as seperate add_custom_command, there needs to be an intermediate target between the two.  See https://github.com/SymbiFlow/symbiflow-arch-defs/pull/773/ and https://github.com/SymbiFlow/symbiflow-arch-defs/pull/825

CMake is pretty dumb in this regard.

common/cmake/devices.cmake
litghost: So if you only depend on FINAL_TARGET, you have the target dependency but not the file dependency.  As a result, if the FINAL_TARGET was already built, but FINAL_OUTPUT changed, the device type won't be see as "old".

CMake's depedency system for output is crazy, which is why I recommend using "append_file_dependencies".

litghost: For further reading: https://samthursfield.wordpress.com/2015/11/21/cmake-dependencies-between-targets-and-files-and-custom-commands/

common/cmake/devices.cmake
litghost: ``xml_lint()`` is currently broken if you attempt to invoke it from a directory other than the directory in which the add_custom_command was defined.  If ``xml_lint()`` is invoked after the file is already generated, everything works out, but it doesn't model the target dependency correctly.

elmsfu: This seems out of scope of this change. FINAL_FILE is the full path so bypasses the problem you mentioned, but I agree this should be addressed. Creating a separate issue.

common/cmake/devices.cmake
elmsfu: Done

ice40/devices/top-routing-virt/CMakeLists.txt
elmsfu: Done

common/cmake/devices.cmake
elmsfu: Done

ice40/icestorm.cmake
acomodi: I have tried to use this fix so that we can avoid using https://github.com/SymbiFlow/vtr-verilog-to-routing/issues/108. `picosoc` was able to route, but it actually took a significant amount of iterations (480 to be precise).
Would it be better to reduce the threshold even more (say 0.4/0.5)?
By looking at all the routing logs the iteration would never be more than 50 (this means that if routing is taking more than 50 iterations it's highly probable it encountered some high congestion).

As far as I understand the `congested_routing` starts at iteration number `max_iterations * congested_routing_iteration_threshold`, with max_iteration set to 500, a  threshold of 0.4 seems reasonable IMO.

litghost: I actually think we should consider a new option.  When I've seen the router struggling, it is usually on the last couple overused nodes (overused nodes 1-3).  @elmsfu / @acomodi Can you go back and check what the router iteration log looked like?  I'm thinking the new option is something akin to "when overused nodes < 5 and does not converge in X iterations, use congestion algo".

elmsfu: I like this idea, definitely a better heuristic. I suggest a new issue be opened for that (especially as it is mostly a VPR feature) and we merge this PR unless there are other open comments.

common/cmake/devices.cmake
```diff
@@ -324,19 +319,22 @@ function(DEFINE_DEVICE_TYPE)
       set(TEMP_TARGET arch.${OUTPUT_NAME}.xml)
       add_custom_command(
 	OUTPUT ${TEMP_TARGET}
-	COMMAND ${PYTHON3} ${CMD_W_ARGS} ${FINAL_OUTPUT} > ${TEMP_TARGET}
+	COMMAND ${CMD_W_ARGS} < ${FINAL_FILE} > ${TEMP_TARGET}
 	DEPENDS
         ${PYTHON3} ${PYTHON3_TARGET}
-	${CMD} ${FINAL_OUTPUT}
+	${CMD} ${FINAL_TARGET}
```
litghost: This needs to depend on both ``${FINAL_FILE}`` and ``${FINAL_TARGET}``.  I recommend using ``append_file_dependencies`` over direct inclusion to avoid forgetting one over the other (and because of include file dependencies, if any).

PR #708, opened 2019-05-09
Initial xc7 route timing import.

litghost:
This is the initial 7-series routing import, which currently just writes 0 delays for all routing, but will use timing information from prjxray-db once present.

### DIFF
--- a/utils/check_ptc.py
+++ b/utils/check_ptc.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 """ Tool for sanity checking rrgraph CHAN PTC's.
 """
 import lxml.etree as ET
@@ -47,32 +48,35 @@ def check_ptc(xml):
         if not all((starts_at_zero, ends_at_max_val, all_values_present)):
             sorted_nodes = sorted(zip(ptcs, nodes), key=lambda x: x[0])
             for idx, (ptc, node) in enumerate(sorted_nodes):
-                if idx != ptc:
-                    if idx > 1:
-                        raise ValueError(
-                            """Gap in ptc value for type = {node_type} @ ({x}, {y})
+                if idx == ptc:
+                    continue
+
+                if idx > 1:
+                    raise ValueError(
+                        """\
+Gap in ptc value for type = {node_type} @ ({x}, {y})
 Expect PTC = {idx}, found {ptc}
 Current node is id = {cur_node}
 Previous node is id = {prev_node}""".format(
-                                x=x,
-                                y=y,
-                                node_type=node_type,
-                                idx=idx,
-                                ptc=ptc,
-                                cur_node=node,
-                                prev_node=sorted_nodes[idx - 1][1],
-                            )
+                            x=x,
+                            y=y,
+                            node_type=node_type,
+                            idx=idx,
+                            ptc=ptc,
+                            cur_node=node,
+                            prev_node=sorted_nodes[idx - 1][1],
                         )
-                    else:
-                        raise ValueError(
-                            "Lowest ptc is {ptc} for type = {node_type} @ ({x}, {y})"
-                            .format(
-                                x=x,
-                                y=y,
-                                node_type=node_type,
-                                ptc=ptc,
-                            )
+                    )
+                else:
+                    raise ValueError(
+                        "Lowest ptc is {ptc} for type = {node_type} @ ({x}, {y})"
+                        .format(
+                            x=x,
+                            y=y,
+                            node_type=node_type,
+                            ptc=ptc,
                         )
+                    )
 
 
 def main():
--- a/utils/check_ptc.py
+++ b/utils/check_ptc.py
@@ -3,6 +3,7 @@
 import lxml.etree as ET
 import argparse
 
+
 def check_ptc(xml):
     """ Checks ptc values used on CHANX and CHANY rr graph nodes are valid.
 
@@ -24,19 +25,23 @@ def check_ptc(xml):
             loc_xml = node.find('loc')
             assert loc_xml is not None
 
-            for x in range(int(loc_xml.attrib['xlow']), int(loc_xml.attrib['xhigh'])+1):
-                for y in range(int(loc_xml.attrib['ylow']), int(loc_xml.attrib['yhigh'])+1):
+            for x in range(int(loc_xml.attrib['xlow']),
+                           int(loc_xml.attrib['xhigh']) + 1):
+                for y in range(int(loc_xml.attrib['ylow']),
+                               int(loc_xml.attrib['yhigh']) + 1):
                     key = (node_type, x, y)
 
                     if key not in chan_ptcs:
                         chan_ptcs[key] = []
 
-                    chan_ptcs[key].append((node.attrib['id'], int(loc_xml.attrib['ptc'])))
+                    chan_ptcs[key].append(
+                        (node.attrib['id'], int(loc_xml.attrib['ptc']))
+                    )
 
     for (node_type, x, y), node_ptcs in chan_ptcs.items():
         nodes, ptcs = zip(*node_ptcs)
         starts_at_zero = min(ptcs) == 0
-        ends_at_max_val = max(ptcs) == len(ptcs)-1
+        ends_at_max_val = max(ptcs) == len(ptcs) - 1
         all_values_present = len(ptcs) == len(set(ptcs))
 
         if not all((starts_at_zero, ends_at_max_val, all_values_present)):
@@ -44,25 +49,30 @@ def check_ptc(xml):
             for idx, (ptc, node) in enumerate(sorted_nodes):
                 if idx != ptc:
                     if idx > 1:
-                        raise ValueError("""Gap in ptc value for type = {node_type} @ ({x}, {y})
+                        raise ValueError(
+                            """Gap in ptc value for type = {node_type} @ ({x}, {y})
 Expect PTC = {idx}, found {ptc}
 Current node is id = {cur_node}
 Previous node is id = {prev_node}""".format(
-                            x=x,
-                            y=y,
-                            node_type=node_type,
-                            idx=idx,
-                            ptc=ptc,
-                            cur_node=node,
-                            prev_node=sorted_nodes[idx-1][1],
-                            ))
+                                x=x,
+                                y=y,
+                                node_type=node_type,
+                                idx=idx,
+                                ptc=ptc,
+                                cur_node=node,
+                                prev_node=sorted_nodes[idx - 1][1],
+                            )
+                        )
                     else:
-                        raise ValueError("Lowest ptc is {ptc} for type = {node_type} @ ({x}, {y})".format(
-                            x=x,
-                            y=y,
-                            node_type=node_type,
-                            ptc=ptc,
-                            ))
+                        raise ValueError(
+                            "Lowest ptc is {ptc} for type = {node_type} @ ({x}, {y})"
+                            .format(
+                                x=x,
+                                y=y,
+                                node_type=node_type,
+                                ptc=ptc,
+                            )
+                        )
 
 
 def main():
@@ -73,6 +83,7 @@ def main():
 
     xml = ET.parse(args.input_xml, ET.XMLParser(remove_blank_text=True))
     check_ptc(xml)
+
 
 if __name__ == "__main__":
     main()
--- a/utils/check_ptc.py
+++ b/utils/check_ptc.py
@@ -1,0 +1,78 @@
+""" Tool for sanity checking rrgraph CHAN PTC's.
+"""
+import lxml.etree as ET
+import argparse
+
+def check_ptc(xml):
+    """ Checks ptc values used on CHANX and CHANY rr graph nodes are valid.
+
+    CHAN ptc numbers are an index per x/y coordinate and channel type (CHANX or
+    CHANY).  ptc's at a particular coordinate/type must start at 0, and fill
+    to the max value.
+
+    """
+    chan_ptcs = {}
+    nodes = {}
+
+    for node in xml.find('rr_nodes').iter('node'):
+        assert node.attrib['id'] not in nodes
+        nodes[node.attrib['id']] = node
+
+        node_type = node.attrib['type']
+
+        if node_type in ['CHANX', 'CHANY']:
+            loc_xml = node.find('loc')
+            assert loc_xml is not None
+
+            for x in range(int(loc_xml.attrib['xlow']), int(loc_xml.attrib['xhigh'])+1):
+                for y in range(int(loc_xml.attrib['ylow']), int(loc_xml.attrib['yhigh'])+1):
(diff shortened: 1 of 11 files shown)


xc7/utils/prjxray_form_channels.py
```diff
@@ -82,28 +205,42 @@ def import_tile_type(db, write_cur, tile_types, site_types, tile_type_name):
     tile_types[tile_type_name] = write_cur.lastrowid
 
     wires = {}
-    for wire in tile_type.get_wires():
+    for wire, wire_rc_element in tile_type.get_wires().items():
+        capacitance = 0.0
```
mkurc-ant: Shouldn't we insert here some reasonable dummy variables other than zero ? Zeros might trigger some assertions in the VPR or lead to some math exceptions / corner cases.

xc7/utils/prjxray_form_channels.py
```diff
@@ -137,21 +275,54 @@ def add_wire_to_site_relation(
...
+                    drive_resistance = site_pin.timing.drive_resistance / 1e3
+                elif isinstance(site_pin.timing, prjxray.tile.InpinTiming):
+                    # microFarads -> Farads
+                    capacitance = site_pin.timing.capacitance / 1e6
+                else:
+                    assert False, site_pin
+            else:
+                # Use min value instead of 0 to prevent
+                # VPR from freaking out over a zero net delay.
+                intrinsic_delay = 2**-126
```
mkurc-ant: I guess it would be good to set driver_resistance and capacitance also to 2**-126 (or other)

xc7/utils/prjxray_form_channels.py
```diff
@@ -653,24 +973,31 @@ def insert_tracks(conn, tracks_to_insert):
...
                 node_type = graph2.NodeType.CHANX
             elif track.direction == 'Y':
                 node_type = graph2.NodeType.CHANY
             else:
                 assert False, track.direction
 
+            if idx == 0:
+                capacitance, resistance = get_node_rc(conn, node)
+            else:
+                capacitance = 0
```
mkurc-ant: I'd also avoid zeros here.

xc7/utils/prjxray_form_channels.py
```diff
@@ -634,6 +929,31 @@ def classify_nodes(conn):
...
+    """
+    capacitance = 0
+    resistance = 0
+
+    cur = conn.cursor()
+    for wire_cap, wire_res in cur.execute("""
+SELECT capacitance, resistance FROM wire_in_tile WHERE pkey IN (
+    SELECT wire_in_tile_pkey FROM wire WHERE node_pkey = ?
+    );""", (node_pkey, )):
+        capacitance += wire_cap
```
mkurc-ant: This calculation is wrong from the electrical point of view. The actual R and C of a node will depend on its physical position and length and also which "to" and "from" edges are considered.

xc7/utils/prjxray_form_channels.py
litghost: No, that is not what is important.

xc7/utils/prjxray_form_channels.py
litghost: 0 is not only legal, but most wires in the actual timing model use zeros.

xc7/utils/prjxray_form_channels.py
litghost: You are wrong, 0 is the most correct.

xc7/utils/prjxray_form_channels.py
litghost: This is not an electrical model, but a timing model.  This is in fact the correct timing model.  I've created an Excel based timing model for approximating the Vivado model ( https://github.com/SymbiFlow/prjxray/blob/master/prjxray/timing.py ), and it matches Vivado's timing model +/- 1.25 %

HackerFoo:
I looked at the changes, but I'm not familiar yet with any of this, so I don't think I can contribute a useful review.

xc7/utils/prjxray_create_edges.py
```diff
@@ -285,11 +285,273 @@ def __init__(self, pins=None, tracks=None):
...
+            loc,
+            other_connector,
+            src_wire_pkey=None,
+            dest_wire_pkey=None,
+            pip_pkey=None,
+            switch_pkey=None
+    ):
         """ Connect two Connector objects at a location within the grid.
 
         Args:
```
elmsfu: Add docstring for new arguments.

xc7/utils/prjxray_form_channels.py
```diff
@@ -72,7 +75,127 @@ def import_site_type(db, write_cur, site_types, site_type_name):
...
+        -------
+        switch_pkey : int
+            Switch primary key that represents provided arguments.
+
+        """
+        delay = 0.0
+        drive_resistance = 0.0
+        internal_capacitance = 0.0
+
+        if pip_timing is not None:
```
elmsfu: Would prefer if we didn't have to convert here. Could add conversion to the PipTiming class.
Higher level question, why are the values not in base SI?

xc7/utils/prjxray_form_channels.py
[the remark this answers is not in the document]
litghost: Agreed, will do

xc7/utils/prjxray_create_edges.py
```diff
@@ -285,11 +285,273 @@ def __init__(self, pins=None, tracks=None):
...
+            Switch row primary key, can be used if switch_pkey is already
+            known (e.g. synthetic edge).  If switch_pkey is not None, other
+            arguments should be None to avoid ambiguity.
+
+        Returns
+        -------
+        Switch row primary key to connect through specified pip.
+
+        """
+        if switch_pkey is not None:
```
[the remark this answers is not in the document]
litghost: In the case of synthetic edges (e.g. constant network, or bypasses), a lookup cannot succeed.  In these cases a switch_pkey is provided, rather than attempting a lookup that will fail.

utils/lib/connection_database.py
```diff
@@ -25,11 +25,10 @@ def create_tables(conn):
     c.execute(
         """
 INSERT INTO
-    switch(name)
+    switch(name, internal_capacitance, drive_resistance, intrinsic_delay, switch_type)
 VALUES
-    ("__vpr_delayless_switch__"),
-    ("routing"),
```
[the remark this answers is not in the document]
litghost: "routing" was the catch all routing switch, but with this PR it is replaced with usage of actually switches.

litghost:
> Some questions. A big PR like this is difficult to take in all at once.

I do apologize for the size, but most of this PR is unfortunately tied together.  I can isolated the check_ptc utility, but the major other changes in form_channels and create_edges are fairly coupled.  If needed, I can attempt to isolate some of the smaller changes, but I don't think I can isolate the ~600 lines in form_channels and create_edges to small PRs.

xc7/utils/prjxray_create_edges.py
litghost: Yep, missed that.  Done

xc7/utils/prjxray_form_channels.py
litghost: The reason I chose these units is that timing values are done in non-SI units (e.g. ns or ps) by default.  The choose units are internally consistent (milliOhms * microFarads = nanoseconds), and 21 ns is much easier to read than 0.000000021 seconds.  For model development it is significantly easier to work with units that are close to the decimal point.

It is worth noting that VPR cannot accept anything besides SI units, hence why the conversion is being done here.

elmsfu: Makes sense. I think a comment in `PipTiming` describing the internal consistency and methods in the class to return in base SI would be cleaner.

litghost: There is a comment to that affect here: https://github.com/SymbiFlow/prjxray/blob/master/prjxray/timing.py#L75 but I can add more

litghost: How do you feel about using http://docs.enthought.com/scimath/units/unit_numpy.html ?

litghost: I'm unimpressed with the documentation on scimath, I withdraw the suggestion.

elmsfu: Ahh I was looking at [PipTiming](https://github.com/SymbiFlow/prjxray/blob/02026c2d8bffd5b0d050b0d0fa242590ba7e6aa8/prjxray/tile.py#L39)

elmsfu:
I think it looks good once it's green. I'd like another person to look over and give approval because I am probably missing some details.

utils/check_ptc.py
```diff
@@ -0,0 +1,89 @@
+""" Tool for sanity checking rrgraph CHAN PTC's.
```
mithro:
```suggestion
#!/usr/bin/env python3

""" Tool for sanity checking rrgraph CHAN PTC's.
```

utils/check_ptc.py
```diff
@@ -0,0 +1,89 @@
...
+
+def check_ptc(xml):
+    """ Checks ptc values used on CHANX and CHANY rr graph nodes are valid.
+
+    CHAN ptc numbers are an index per x/y coordinate and channel type (CHANX or
+    CHANY).  ptc's at a particular coordinate/type must start at 0, and fill
+    to the max value.
+
+    """
+    chan_ptcs = {}
```
mithro: Maybe `chan_ptcs = defaultdict(list)` (unclear if it's better than just the if test....)

utils/check_ptc.py
```diff
@@ -0,0 +1,89 @@
...
+        ends_at_max_val = max(ptcs) == len(ptcs) - 1
+        all_values_present = len(ptcs) == len(set(ptcs))
+
+        if not all((starts_at_zero, ends_at_max_val, all_values_present)):
+            sorted_nodes = sorted(zip(ptcs, nodes), key=lambda x: x[0])
+            for idx, (ptc, node) in enumerate(sorted_nodes):
+                if idx != ptc:
+                    if idx > 1:
+                        raise ValueError(
+                            """Gap in ptc value for type = {node_type} @ ({x}, {y})
```
mithro:
```suggestion
                            """\
Gap in ptc value for type = {node_type} @ ({x}, {y})
```

utils/check_ptc.py
```diff
@@ -0,0 +1,89 @@
...
+    for (node_type, x, y), node_ptcs in chan_ptcs.items():
+        nodes, ptcs = zip(*node_ptcs)
+        starts_at_zero = min(ptcs) == 0
+        ends_at_max_val = max(ptcs) == len(ptcs) - 1
+        all_values_present = len(ptcs) == len(set(ptcs))
+
+        if not all((starts_at_zero, ends_at_max_val, all_values_present)):
+            sorted_nodes = sorted(zip(ptcs, nodes), key=lambda x: x[0])
+            for idx, (ptc, node) in enumerate(sorted_nodes):
+                if idx != ptc:
```
mithro: Can you save a level of indenting with;
```suggestion
                if idx == ptc:
                    continue
```

xc7/utils/prjxray_create_edges.py
```diff
@@ -285,22 +285,301 @@ def __init__(self, pins=None, tracks=None):
...
+        src_wire_pkey : int
+            Source wire row primary key.  May be None if switch_pkey is not
+            None.
+        pip_pkey : int
+            Pip connecting source to destination wire.  May be None if
+            switch_pkey is not None.
+        dest_wire_pkey : int
+            Destination wire row primary key.  May be None if switch_pkey
+            is not None.
+        switch_pkey : Int
```
mithro:
```suggestion
        switch_pkey : int
```

xc7/utils/prjxray_create_edges.py
```diff
@@ -285,22 +285,301 @@ def __init__(self, pins=None, tracks=None):
...
+            assert pip_pkey is None
+            return switch_pkey
+        else:
+            assert switch_pkey is None
+
+        cur = self.conn.cursor()
+
+        cur.execute(
+            """
+        SELECT src_wire_in_tile_pkey, dest_wire_in_tile_pkey, switch_pkey, backward_switch_pkey
```
mithro:
```suggestion
SELECT src_wire_in_tile_pkey, dest_wire_in_tile_pkey, switch_pkey, backward_switch_pkey
FROM pip_in_tile WHERE pkey = ?
```

xc7/utils/prjxray_create_edges.py
```diff
@@ -285,22 +285,301 @@ def __init__(self, pins=None, tracks=None):
...
+            See diagram for details.
+
+        Diagram:
+
+         Site pin
+            +
+            |  tile wire #1  +---+ tile wire #2
+            +>-------------->+pip+--------------->
+            |                +---+
+            +
```
mithro: Maybe improve the diagram as;
```
           --+
￼             |    tile wire #1  +-----+ tile wire #2
￼             +==>-------------->+ pip +--------------->
￼             | ^-Site pin       +-----+
￼           --+
```

xc7/utils/prjxray_create_edges.py
```diff
@@ -285,22 +285,301 @@ def __init__(self, pins=None, tracks=None):
...
+FROM wire WHERE pkey = ?""", (wire_pkey, )
+        )
+        values = cur.fetchone()
+        node_pkey = values[0]
+        edge_nodes = values[1:5]
+        site_pin_graph_node_pkey = values[5]
+
+        cur.execute(
+            """
+SELECT site_pin_switch_pkey FROM wire_in_tile WHERE pkey = (
```
mithro: The wrapping of these `cur.execute` SQL strings are not very consistent....

xc7/utils/prjxray_create_edges.py
```diff
@@ -326,7 +610,21 @@ def connect_at(self, loc, other_connector):
             tracks_model, graph_nodes = other_connector.tracks
             for idx, pin_dir in tracks_model.get_tracks_for_wire_at_coord(loc):
                 if pin_dir in self.pins.edge_map:
-                    return self.pins.edge_map[pin_dir], graph_nodes[idx]
+                    # Site pin -> Interconnect is modelled as:
+                    #
```
mithro: Extra line?

xc7/utils/prjxray_create_edges.py
```diff
@@ -335,27 +633,47 @@ def connect_at(self, loc, other_connector):
             tracks_model, graph_nodes = self.tracks
             for idx, pin_dir in tracks_model.get_tracks_for_wire_at_coord(loc):
                 if pin_dir in other_connector.pins.edge_map:
-                    return graph_nodes[idx], other_connector.pins.edge_map[
-                        pin_dir]
+                    # Interconnect -> Site pin is modelled as:
+                    #
```
mithro: Extra line?

PR #711, opened 2019-05-10
Add Vivado targets

litghost:
Initial use of Vivado.  This supports is generation of design checkpoint, bitstream and design project from fasm2v output.

Targets:
 - Load checkpoint in Vivado GUI
 - Load project in Vivado GUI
 - Setup basic simulation using clock forces
 - Output Vivado timing information to file
 - Compare Vivado bitstream -> FASM to VPR bitstream -> FASM


xc7/utils/vivado_output_timing.py
```diff
@@ -0,0 +1,32 @@
...
+""".format(name=args.name, util_tcl=args.util_tcl),
+        file=f_out
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+
+    parser.add_argument('--name', required=True)
+    parser.add_argument('--util_tcl', required=True)
```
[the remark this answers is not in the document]
litghost: So I don't understand the question?  Are you ask what the argument is for?

xc7/utils/timing_utils.tcl
```diff
@@ -0,0 +1,101 @@
+proc write_timing_info {filename} {
```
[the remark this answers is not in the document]
litghost: Sure, fixed.

xc7/utils/vivado_output_timing.py
```diff
@@ -0,0 +1,32 @@
+""" Utility for generating TCL script to output timing information from a
+design checkpoint.
+"""
+import argparse
+
+
+def create_runme(f_out, args):
```
elmsfu: nit: name doesn't reflect function

```suggestion
def create_output_timing(f_out, args):
```

xc7/utils/vivado_output_timing.py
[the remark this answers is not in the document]
elmsfu: Sorry for the ramble, but trying to think of a simpler approach.

I see that timing_utils.tcl script is being passed, but wonder if we would ever have a reason to pass a different script as the argument. If not, may be they shouldn't be separated into different files. The flow is fragmented across tcl, python, and the cmake invocation.

Maybe there isn't a way around it for generating/running tcl wrapped in python, but this python generates a new tcl script that runs 3 commands and I don't see  using it with different parameters.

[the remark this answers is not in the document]
litghost: The purpose of the argument is *not* to enable switching of scripts.  The reason for the argument is two fold.  The first is to avoid knowledge of paths in the templating script.  The second is more fundamental, which is that non-trival TCL should not be located within a python template.

As for whether we can avoid the python script altogether, I think this approach (writing the tcl script and then invoking vivado to source it) is less bad than something that involves passing arguments are environment variables, or having CMake cat the script to a file, etc.

xc7/utils/vivado_output_timing.py
litghost: Fixed

xc7/utils/vivado_output_timing.py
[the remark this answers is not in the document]
elmsfu: ok, sounds good. I figured it might be a matter of some trade-off, I think the path makes sense. I could go either way on if non-trivial TCL is better as separate file or as a string in a template. First option is easier to develop but a little more difficult for a new reader to track and adds cmake complexity in my opinion.

Happy with this solution though, so marking as resolved.

xc7/utils/vivado_create_runme.py
```diff
@@ -0,0 +1,76 @@
+""" Utility for generating TCL script to generate bitstream, project and
```
mithro: Add 
```suggestion
#!/usr/bin/env python3

""" Utility for generating TCL script to generate bitstream, project and 
```

xc7/utils/vivado_create_runme.py
```diff
@@ -0,0 +1,76 @@
...
+report_timing_summary
+""".format(name=args.name, ),
+        file=f_out
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+
+    parser.add_argument('--name', required=True)
```
mithro: Help for these arguments would be useful....

xc7/utils/vivado_create_sim.py
```diff
@@ -0,0 +1,48 @@
...
+restart
+run 1us
+run 1us
+""", file=f_out)
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+
+    parser.add_argument('--top', required=True)
```
mithro: Help for these arguments?

xc7/utils/timing_utils.tcl
```diff
@@ -0,0 +1,109 @@
...
```
mithro: Maybe this is better in `xc7/utils/vivado/timing_utils.tcl`

litghost: I think we already an excessively deep folder hierarchy, I see no reason to make it even deeper, especially considering it would have at most 4 items, and is joining 14 other items.

xc7/utils/vivado_create_runme.py
litghost: No

litghost:
> When generating code for a reasonably powerful language, I prefer to use the language rather than generate it, so that variables are either concatenated to the top or bottom, preferably as a function call. Then it's easy to store the logic of the script as a separate file, which allows reuse and proper syntax highlighting & indentation editor support.
> 
> For example, rather than generating a list of function calls, pass a list in a variable and use "foreach".

I think I'm following your recommendation for the create_timing TCL, and I think runme is a good example where passing many unnamed arguments would be more confusing than the current approach.  This is because this is no significant logic (e.g. no nesting, no flow control, etc) in the runme and simulation cases, so syntax highlighting is less important.  If TCL had a good structured format I could pass (e.g. JSON), I would consider re-writing the runme TCL to consume a JSON file describing the parts.

Thoughts?

litghost:
@HackerFoo https://github.com/SymbiFlow/symbiflow-arch-defs/pull/711/commits/ec71e7808e2d77ea5a127fc3c9785e14a4c016f4 I think is a clear win, but it is not clear to me that that same follows for the runme or simulation TCL cases.

HackerFoo:
@litghost You can use something like this for structured data:

```tcl
set table {fooA barA
           fooB barB}

foreach {foo bar} $table {
  # use $foo and $bar
}
```

Or for something more complex, TCL has [dicts](https://wiki.tcl-lang.org/page/dict#2da322eb1ce304a9e300ec933491fb7bc0a8b54530c56cb21789b57dc9941204).

Ideally, arguments would be passed into Vivado as a dict with [-tclargs](https://www.xilinx.com/support/answers/56501.html) or loaded as a separate TCL script, so there would be no need to generate TCL code, and it would be much easier to manually modify.

litghost:
> @litghost You can use something like this for structured data:
> 
> ```tcl
> set table {fooA barA
>            fooB barB}
> 
> foreach {foo bar} $table {
>   # use $foo and $bar
> }
> ```
> 
> Or for something more complex, TCL has [dicts](https://wiki.tcl-lang.org/page/dict#2da322eb1ce304a9e300ec933491fb7bc0a8b54530c56cb21789b57dc9941204).
> 
> Ideally, arguments would be passed into Vivado as a dict with [-tclargs](https://www.xilinx.com/support/answers/56501.html) or loaded as a separate TCL script, so there would be no need to generate TCL code, and it would be much easier to manually modify.

Neither of those suggestions are actually helpful in improving the quality of this design.  If we are going to generate TCL data (e.g. a TCL dict or otherwise), the current solution is superior.  If we are going to pass arguments via -tclargs, a fully unstructured approach is unsatisfactory.   Seven unnamed arguments is a lot.  Inventing a structured format passed in via ``-tclargs`` seems way worse than what we already have.

litghost:
> @litghost You can use something like this for structured data:
> 
> ```tcl
> set table {fooA barA
>            fooB barB}
> 
> foreach {foo bar} $table {
>   # use $foo and $bar
> }
> ```
> 
> Or for something more complex, TCL has [dicts](https://wiki.tcl-lang.org/page/dict#2da322eb1ce304a9e300ec933491fb7bc0a8b54530c56cb21789b57dc9941204).
> 
> Ideally, arguments would be passed into Vivado as a dict with [-tclargs](https://www.xilinx.com/support/answers/56501.html) or loaded as a separate TCL script, so there would be no need to generate TCL code, and it would be much easier to manually modify.

What would you change about this:
https://gist.github.com/litghost/6d0c1059a0cb86802fe1dabdfb55c718

Keep in mind that pretty much every line of that script has an argument in it.  Template is here: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/711/files#diff-efb3719673ba5ad619c2c63ac9ee5067R10

Basically I'm challenging your assertion this can be improved using more TCL or tclargs.

HackerFoo:
How about this?

```tcl
# untested
proc runme {name part bit_v top bit_tcl clocks} { 
  create_project -force -part $part design design
  read_verilog $bit_v
  synth_design -top $top
  
  source $bit_tcl
  
  foreach {pin, period} $clocks {
      set half_period [expr {double($period) / 2}]
      create_clock -period $period -name $pin -waveform {{0.000 $half_period}} [get_ports $pin]
  }
  
  set_property CFGBVS VCCO [current_design]
  set_property CONFIG_VOLTAGE 3.3 [current_design]
  set_property BITSTREAM.GENERAL.PERFRAMECRC YES [current_design]
  set_property IS_ENABLED 0 [get_drc_checks {{LUTLP-1}}]
  place_design
  route_design
  write_checkpoint -force "design_$name.dcp"
  write_bitstream -force "design_$name.bit"
  save_project_as -force "design_$name.xpr"
  report_timing_summary
}

# below is generated

runme "name" "part" "bit_v" "top" {
    pin_0 period_0
    pin_1 period_1
    ...
}
```

Now there is no confusion about what curly brackets mean. Notice how much easier it would be to modify or add a clock pin by hand, avoiding a simple mistake like setting the clock waveform to `{{0 $period}}`.

HackerFoo:
Even better, though: 

```tcl
source "symbiflow.tcl"

runme "name" "part" "bit_v" "top" {
    pin_0 period_0
    pin_1 period_1
    ...
}
```

litghost:
> Now there is no confusion about what curly brackets mean. Notice how much easier it would be to modify or add a clock pin by hand, avoiding a simple mistake like setting the clock waveform to `{{0 $period}}`.

Except that isn't true?  How would you even add a clock pin by hand?  You'd have to modify the tcl script, then re-invoke the make script.  Versus just modifying the generated output, and re-invoking Vivado.

HackerFoo:
> > Now there is no confusion about what curly brackets mean. Notice how much easier it would be to modify or add a clock pin by hand, avoiding a simple mistake like setting the clock waveform to `{{0 $period}}`.
> 
> Except that isn't true? How would you even add a clock pin by hand? You'd have to modify the tcl script, then re-invoke the make script. Versus just modifying the generated output, and re-invoking Vivado.

What I mean by "by hand" is "without using the build system." You wouldn't need to run any make script, just `vivado -source runme.tcl`.

Of course it probably doesn't make sense to add a clock pin without making other changes, and since `top_bit.v.tcl` is generated it probably doesn't make much sense in this case. It's still more readable, though. You could easily change the duty cycle for all clocks or invert them, but I doubt that's useful?

litghost:
> What I mean by "by hand" is "without using the build system." You wouldn't need to run any make script, just `vivado -source runme.tcl`.

Both designs have this property, unless I'm misunderstanding you.  Both designs require a generation step to write out the argument list.  In your case, the body of the code is sourced.

> It's still more readable, though.

I really don't think I agree with this, unless you are referring to the templating script vs the template output.  I think the flat template output (e.g. no source) is easier to comprehend (no sourcing), and the runme function is not something that is useful as function without the template driver anyways.  This is distinct from the timing output function.

HackerFoo:
A larger interface between two components makes them harder to read, (re)use (e.g. generating scripts for other vendors' tools), document, etc. This is compounded when those components are in different languages, and even more so when they are executed in different phases.

In the case of editing the generated script directly, it's easier when the parameters are in one place, and computed values, such as `period / 2` don't need to be manually computed.

This also makes it easier to read because you only need to read the whole source at most once (or until it works), afterwards you can just look at the parameters. The generator should be easier to read as well, since it just serializes the parameters.

litghost:
> e.g. generating scripts for other vendors' tools

What vendor has a TCL environment that shares any functions with Vivado?  Maybe ISE, but I don't think that's particularly relevant.

> In the case of editing the generated script directly, it's easier when the parameters are in one place, and computed values, such as `period / 2` don't need to be manually computed.

You are assuming that adjusting parameters is going to be the common reason to manually modify the TCL, and I don't think that will be borne out.

> This also makes it easier to read because you only need to read the whole source at most once (or until it works), afterwards you can just look at the parameters. The generator should be easier to read as well, since it just serializes the parameters.

But the runme function isn't an easily consumable thing, its not really a good function (unlike say output timing).  It's just a bunch of operations strung together that work in a narrow space.  Anyone changing parameters is going to have to have both files open to understand the impact of changing a parameter.

HackerFoo:
I don't want to hold this PR up any further, because the generated scripts are fairly small, and I don't think we're making progress in this discussion.

PR #714, opened 2019-05-10
WIP: Generate arch.xml from verilog

kgugala:
This is an initial PR for generating the arch.xml file from verilog sim models. For now it will fail as the models don't have any fasm tags. Also, muxes are generated incorrectly.

I don't want to stall the PR until all the v2x fixes are merged, I'll continue work on arch generation and rebase it when v2x features are merged.


litghost:
For clarity, does this generate an arch that correctly place and routes, but fails in genfasm?

kgugala:
not yet. Right now it fails on routing muxes v2x conversion. To be more precise `vlog_to_pbtype` fails on https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/primitives/common_slice/routing/srusedmux/srusedmux.sim.v#L10 saying that `I0` has no driver - the tool cannot handle constant drivers. I'm not sure how those constant drives should be actually handled - what pb_type's interconnect should be generated from those.

Since @mithro / @mkurc-ant are working on v2x features I think this will not be a case soon.

litghost:
> `I0` has no driver - the tool cannot handle constant drivers. I'm not sure how those constant drives should be actually handled - what pb_type's interconnect should be generated from those.

Constant drives?  E.g. things connected to a site internal constant source (e.g. 0 or 1)?  They should come from special black boxes, see https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/primitives/common_slice/common_slice.pb_type.xml#L68 .  Then during synthesis, if a real blackbox (e.g. FDRE) has a connection that supports one of these dedicated constant blackboxes, it will emit that special blackbox.

kgugala:
OK, so it is a bug in vlog_to_pbtype. I'll open an issue for that

xc7/primitives/common_slice/routing/Nused/CMakeLists.txt
```diff
@@ -1,16 +1,7 @@
 add_file_target(FILE ntemplate.Nused.sim.v SCANNER_TYPE verilog)
```
mithro: As far as I can tell, the `Nused` elements don't have any features / functionality? They don't seem to appear in any of the XML files?

PR #730, opened 2019-05-15
Change CARRY0 + 3x CARRY to CARRY4_VPR.

litghost:
Requires https://github.com/SymbiFlow/yosys/pull/23

Fixes #729.

- [x] Need to update Yosys once https://github.com/SymbiFlow/yosys/pull/23 is merged and build
- [x] Update timing with BEL approximations

PR #737, opened 2019-05-20
Change from blackbox to whitebox annotation.

litghost:
Yosys's treatment of blackbox attributes's now includes stripping all contents
within the blackbox.  The whitebox attribute is what should be used to
explicitly define contents of a blackbox.

This PR also updates Yosys (as of 4 days ago at b6345b1).


litghost:
Superceeded by #730

PR #738, opened 2019-05-20
Arch XML timings

kgugala:
This PR adds one additional step in arch.xml generation - it uses sdf timing info to update the BEL timings entries in the xml.

For now it updates only LUT5 timings, but can be easily adopted to update all the required delays.

The same mechanism could be used after we completely switch to arch.xml generation from verilog files using V2X.


utils/update_arch_timings.py
```diff
@@ -0,0 +1,109 @@
+import lxml.etree as ET
```
litghost: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/577

kgugala:
@mithro do we have to add travis deploy keys to the `python-sdf-timing` repo?

kgugala:
> @mithro do we have to add travis deploy keys to the `python-sdf-timing` repo?

nevermind, I changed the submodule to https

utils/update_arch_timings.py
```diff
@@ -0,0 +1,118 @@
...
+import argparse
+from sdf_timing import sdfparse
+from sdf_timing.utils import get_scale_seconds
+import re
+import os
+
+
+def remove_site_number(site):
```
mithro: Can you add a docstring?

utils/update_arch_timings.py
```diff
@@ -0,0 +1,118 @@
...
+    number = re.search(r'\d+$', site)
+    if number is not None:
+        site = site[:-len(str(number.group()))]
+    return site
+
+
+def get_cell_type(bel, site):
```
mithro: Can you add a docstring?

utils/update_arch_timings.py
```diff
@@ -0,0 +1,118 @@
...
+    if bel.endswith("5LUT") and site == "SLICEL":
+        return "LUT5"
+    if bel.endswith("5LUT") and site == "SLICEM":
+        return "LUT_OR_MEM5LRAM"
+
+    return None
+
+
+def find_timings(timings, bel, site):
```
mithro: Add docstring?

utils/update_arch_timings.py
```diff
@@ -0,0 +1,118 @@
...
+    for delay in cells[cellsite]:
+        if not cells[cellsite][delay]['is_absolute']:
+            continue
+        entry = cells[cellsite][delay]['delay_paths']['slow']['max']
+        bel_timings[delay] = float(entry) * get_scale_seconds('1 ns')
+
+    return bel_timings
+
+
+def get_pb_type_chain(node):
```
mithro: Should this be part of `lib/pb_type.py`?

utils/update_arch_timings.py
```diff
@@ -0,0 +1,118 @@
...
+            return list(reversed(pb_types))
+
+        if parent.tag == 'pb_type':
+            pb_types.append(parent.attrib['name'])
+
+        node = parent
+
+
+def main():
+
```
mithro: Extra new line.

mithro:
@elmsfu How does this compare to your import for the iCE40?

utils/update_arch_timings.py
kgugala: done. I've moved the function to lib/pb_type

utils/update_arch_timings.py
kgugala: fixed

utils/bels.json
```diff
@@ -0,0 +1,172 @@
+{
```
litghost: Why is this in utils?  It belongs to xc7?

kgugala: yes it should be in xc7. I'll move it there

kgugala:
> @kgugala Did you look at @elmsfu's pull request which seems to do something similar for ice40?

you mean #693 ? 
the scripts passed to `DEFINE_DEVICE_TYPE` looks interesting. I'll cherry pick it here - this way we will not conflict with each other,

mithro:
@kgugala Yes, #693 - it imports an SDF into an XML file.

kgugala:
@mithro I think the sdf import itself has to be arch specific, but I'll reuse the generic script approach

utils/CMakeLists.txt
```diff
@@ -1,9 +1,12 @@
 get_target_property_required(PYTEST env PYTEST)
 
+set(PYTHON_SDF_TIMING_DIR ${symbiflow-arch-defs_SOURCE_DIR}/third_party/python-sdf-timing)
+
 add_custom_target(
   test_python_utils
   DEPENDS conda_pytest intervaltree
-  COMMAND ${PYTEST} --doctest-modules -vv
+  COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${PYTHON_SDF_TIMING_DIR}
```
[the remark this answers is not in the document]
kgugala: yep, I'll do that

utils/update_arch_timings.py
```diff
@@ -0,0 +1,121 @@
...
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        '--input_arch', required=True, help="Input arch.xml file"
+    )
+    parser.add_argument('--sdf_dir', required=True, help="SDF files directory")
+    parser.add_argument(
+        '--out_arch', required=True, help="Output arch.xml file"
+    )
+    parser.add_argument(
```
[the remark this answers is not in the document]
kgugala: I'm adding tags to the base xmls in this PR. That solves the issue of which delay from sdf should be used where. see: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/8f605b137f32785ff3543d549a19df5bb786379a/xc7/primitives/ff/ff.pb_type.xml#L85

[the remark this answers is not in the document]
kgugala: json is used to translate the vpr naming to Vivado naming (exported to sdfs)

utils/update_arch_timings.py
kgugala: done

utils/update_arch_timings.py
kgugala: done

utils/update_arch_timings.py
kgugala: done

kgugala:
@mithro this PR needs the latest timing DB to pass the CI.

utils/bels.json
kgugala: done. The file is now in xc7/bels.json

utils/CMakeLists.txt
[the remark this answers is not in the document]
kgugala: done

xc7/primitives/slicem/Ndram/ntemplate.N_dram.pb_type.xml
```diff
@@ -18,11 +18,11 @@
       <input  name="in"  num_pins="5" port_class="lut_in" />
       <output name="out" num_pins="1" port_class="lut_out" />
       <delay_matrix type="max" in_port="in" out_port="out">
-        0.068e-9
-        0.068e-9
-        0.068e-9
-        0.068e-9
-        0.068e-9
+        {{iopath_A1_O5}}
```
mithro: Why is this double escaped?

kgugala: because it first goes through templating (`n.py`) which also uses python's `.format`. The same happens with LUTs in Nlut.pb_type.xml

xc7/primitives/ff/ff.pb_type.xml
```diff
@@ -45,13 +45,13 @@
           <clock  name="C"  num_pins="1"/>
           <input  name="S"  num_pins="1"/>
           <output name="Q"  num_pins="1"/>
```
mithro: What is preventing us either;
 (a) Renaming the port names in the XML to match the SDF file? IE
```xml
        <T_setup    value="{setup_CLK_DIN}"   clock="CLK" port="DIN" />
	<T_setup    value="{setup_CLK_CE}"    clock="CLK" port="CE"  />
	<T_setup    value="{recovery_CLK_SR}" clock="CLK" port="SR"  />
	<T_hold     value="{hold_CLK_DIN}"    clock="CLK" port="DIN" />
	<T_hold     value="{hold_CLK_CE}"     clock="CLK" port="CE"  />
	<T_hold     value="{removal_CLK_SR}"  clock="CLK" port="SR"  />
	<T_clock_to_Q max="{iopath_CLK_Q}"    clock="CLK" port="Q"   />
```
 (b) Rewriting the SDF file to match these names?

kgugala: the problem is that the bel timing fuzzer was bugged and did not emit those timings correctly. This merge https://github.com/SymbiFlow/prjxray/commit/8cb7366fa2c361fd5c8eec0aa0cb976808a39752 fixes that. Current DB was generated before the fix has been applied so it's missing e.g `setup_CLK_DIN`.

mithro: @kgugala I don't think that answers my question? I was assuming you had the values like `setup_CLK_DIN` available to substitute?

kgugala: not in the current database. Because of the bug some timingchecks were emitted as e.g. `setup_CLK_CLK` or not emitted at all. This is now fixed, but not present in DB yet. Locally I run it with sdfs generated with the latest fuzzer and I can substitute all the timings tags in this PR. 

Answering your question: we need to update sdfs in the DB to have those values.

mithro: My question is **not** about the values.

My question is about the names. Is there any reason we can't use the same names in both the XML and SDF files?

litghost: The names have to be different in at least one case, the SR timing is shared between the FDRE and FDSE, and the pin names must match the blackbox.  In theory we could handle this at the techmap, but I don't think that is a good idea.

kgugala: ah. 
No, I don't think there is a reason preventing us from using the same names. I think rewriting xmls will be easier. The names in sdfs are in fact the Vivado names. Unfortunately those are not always consistent - e.g. the `D` input of a dff is named `DIN` in timingchecks and `D` in iopaths (in Latch configuration). BEL timing fuzzer could be updated to make the pins consistent.

I'd do the xml pin renaming in a separate PR so it is reviewable

litghost: > I'd do the xml pin renaming in a separate PR so it is reviewable

Renaming the XML is not good enough, because the blackbox pin names must match the eblif.  I don't think I recommend changing the XML at all.

mithro: Where do the blackbox pin names come from?

litghost: The eblif and the XML must match.  The eblif is generated from Yosys (generally speaking).

PR #743, opened 2019-05-21
Clean JSON5 output to JSON output for Vivado timing output target.

litghost:


.travis.yml
```diff
@@ -4,7 +4,11 @@ python:
 
 addons:
   apt:
+    sources:
+      - ubuntu-toolchain-r-test
+      - llvm-toolchain-trusty-5.0
```
mithro: `llvm` **and** `g++-6`?

litghost: I copied from prjxray, I'll see if "ubuntu-toolchain-r-test" works.

PR #747, opened 2019-05-21
Enable Cinternal output.

litghost:
This should fix master on CI, currently xc7 is broken at HEAD on master.


mithro:
LGTM.

litghost:
Waiting for green CI, then will merge

Issue #755, opened 2019-05-22
v2x VPR tests fail during packing stage


acomodi:
### Description
There is an issue with the v2x tests to check if the XML that is produced is VPR-compatible.
The VPR packing stage is not able to pack all the nets into one `TILE` only, but uses a `TILE` for each net, exceeding the available architecture resources. 

#### How does it work
v2x generated the `pb_type` and `model` XMLs of the architecture. `vpr_pbtype_arch_wrapper.py` encapsulates the `pb_type` in a complete architecture file.

At this point an `.eblif` file is produced to test VPR against the architecture file.

#### Setup
I have been testing this issue with the `multiple_instance` test. The architecture files and .eblif are the following:

Eblif
```
.model multiple_instance
.inputs a0 a1 b0 b1 cin0 cin1
.outputs cout0 cout1 sum0 sum1
.subckt adder a=a0 b=b0 cin=cin0 cout=cout0 sum=sum0
.subckt adder a=a1 b=b1 cin=cin1 cout=cout1 sum=sum1
.end
```

XML
```xml
<?xml version="1.0"?>
<!-- set: ai sw=1 ts=1 sta et -->
<architecture xmlns:xi="http://www.w3.org/2001/XInclude">
  <models>
    <model name="adder">
      <input_ports>
        <port combinational_sink_ports="cout sum" name="a"/>
        <port combinational_sink_ports="cout sum" name="b"/>
        <port combinational_sink_ports="cout sum" name="cin"/>
      </input_ports>
      <output_ports>
        <port name="cout"/>
        <port name="sum"/>
      </output_ports>
    </model>
  </models>
  <complexblocklist>
    <!-- input buffer -->
    <pb_type name="IBUF">
      <output name="I" num_pins="1"/>
      <pb_type blif_model=".input" name="PIN" num_pb="1">
        <output name="inpad" num_pins="1"/>
      </pb_type>
      <interconnect>
        <direct input="PIN.inpad" name="P" output="IBUF.I"/>
      </interconnect>
      <pinlocations pattern="custom">
        <loc side="right" xoffset="0" yoffset="0">IBUF.I</loc>
      </pinlocations>
    </pb_type>
    <!-- output buffer -->
    <pb_type name="OBUF">
      <input name="O" num_pins="1"/>
      <pb_type blif_model=".output" name="PIN" num_pb="1">
        <input name="outpad" num_pins="1"/>
      </pb_type>
      <interconnect>
[...]

PR #645, opened 2019-04-25
WIP - utils/vlog: More tests from timing tutorial.

mithro:
* [Adding "Figure 43 - XXX Single Port RAM", fixes #XX.](./utils/vlog/tests/fig43-single_port_ram_mixed)

 * [Adding "Figure 44" - Synchronous Single Port RAM, fixes #XX.](./utils/vlog/tests/fig44-single_port_ram_seq)

 * [Adding "Figure 45" - Asynchronous Single Port RAM, fixes #XX.](./utils/vlog/tests/fig45-single_port_ram_seq_comb)

 * [Adding "Figure 46" - Synchronous Dual port RAM with independent clocks, fixes #XX.](./utils/vlog/tests/fig46-multiclock_dual_port_ram)

 * [Adding Single Port ROM.](./utils/vlog/tests/single_port_rom_mixed)

Signed-off-by: Tim 'mithro' Ansell <me@mith.ro>

PR #703, opened 2019-05-09
mux_gen + v2x: Support generating FASM annotations for muxes.

mithro:
* Cleanup of duplicate argument parsing + main code.
 * Separate out building interconnect information from generating the interconnect XML.
 * Support for treating routing muxes specially.
 * New form for muxes similar to new form for the direct XML.
 * Working on the tests.

This currently enables;
 - [x] Automatic generation of `fasm_mux` using mux_gen and v2x.
 - [x] Automatic generation of pack patterns.

This fixes numerous issues;
* Fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/701
* Fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/169 - vlog_to_pbtype.xml tool using object *type* rather than name in output.
* Fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/697 - v2x: Add support for "fasm_mux" in "interconnect"

PR #756, opened 2019-05-22
[DNM][WIP] Use Yosys to generate blif in v2x tests

kgugala:
This is PR is to discuss the VPR issues discovered when attempting to pack the test circuit to the arch generated basing on the pb_type xmls (see  #755 ) 

This PR is based on #645, but only one test is enabled. Also the flow has been modified to use Yosys for blif generation and the ` update_test_eblif.py`  script to transofr the yosys generated blif to a form expected by the test arch.

Once the issues in this PR are resolved, we can close it and cherry-pick the solution to #645

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,8 +110,9 @@ add_custom_target(all_xml_lint)
 add_custom_target(all_vpr_test_pbtype)
 
 add_subdirectory(common)
-add_subdirectory(utils)
 add_subdirectory(vpr)
+# FIXME: utils tests depend on targets from vpr
+add_subdirectory(utils)
 add_subdirectory(library)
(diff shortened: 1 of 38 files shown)


kgugala:
@litghost @mithro this is the code to trigger #755. To reproduce:

clone
```
make env -j`nproc`
cd build/utils/vlog/tests/multiple_instance
make test_multiple_instance
```

mithro:
@kgugala What is the status of this pull request? I think using Yosys to generate the eblif is a pretty good idea (verses using the current method).

kgugala:
@mithro I think #755 is still not solved. @acomodi can you take a look on that?

acomodi:
@kgugala @mithro Yes, I'll look into it

kgugala:
work done here is included in #703. We can close this one

PR #758, opened 2019-05-22
Map wires to nets and flatten wires for easier tracing.

litghost:
Prior to this PR, wires were always named programatically.
This PR uses net map to rename wires, and then propigates those names
back to the module instances.

### DIFF
--- a/xc7/fasm2bels/verilog_modeling.py
+++ b/xc7/fasm2bels/verilog_modeling.py
@@ -30,6 +30,7 @@ import functools
 from .make_routes import make_routes, ONE_NET, ZERO_NET, prune_antennas
 from .connection_db_utils import get_wire_pkey
 
+
 def pin_to_wire_and_idx(pin):
     """ Break pin name into wire name and vector index.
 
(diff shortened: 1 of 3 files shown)


xc7/fasm2bels/verilog_modeling.py
```diff
@@ -31,6 +31,37 @@
...
+
+    >>> pin_to_wire_and_idx('A')
+    ('A', None)
+    >>> pin_to_wire_and_idx('A[0]')
+    ('A', 0)
+    >>> pin_to_wire_and_idx('A[1]')
+    ('A', 1)
+
+    """
+    idx = pin.find('[')
```
mithro: I think we already have this function somewhere.
I also think a regex might have been better here.

The function is so small I don't really care however.

xc7/fasm2bels/verilog_modeling.py
```diff
@@ -114,10 +145,166 @@ def escape_verilog_name(name):
...
+
+
+class Constant(ConnectionModel):
+    """ Represents a boolean constant, e.g. 1'b0 or 1'b1. """
+
+    def __init__(self, value):
+        assert value in [0, 1]
+        self.value = value
+
+    def to_string(self, net_map=None):
```
mithro: Why `to_string` rather than `__str__`?

litghost: Because of the net_map argument.  I don't believe arguments can be passed to __str__?

PR #760, opened 2019-05-23
Increase default buffer delay to be estimate inter-cluster delay.

litghost:


xc7/utils/prjxray_arch_import.py
```diff
@@ -557,7 +557,10 @@ def main():
             "R": "551",
             "Cin": ".77e-15",
             "Cout": "4e-15",
-            "Tdel": "6.8e-12",
+            # TODO: This value should be the "typical" pip switch delay from
```
mithro: Can you a log an issue about this?

mithro:
Also, CI failed?

xc7/utils/prjxray_arch_import.py
litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/762

PR #761, opened 2019-05-23
Output pre-symbiflow techmap'd Verilog.

litghost:
This pre-symbiflow verilog can be used with Vivado synthesis if xc7/techmap/unmap.v is
applied.  This allows for post-Yosys timing comparisions between VPR and Vivado.


mithro:
`unmap.v` doesn't seem to be used anywhere?

litghost:
> `unmap.v` doesn't seem to be used anywhere?

I believe I mentioned that the process of getting Vivado timing was manual, using ``unmap.v`` is part of it.

mithro:
Do you want to add a script or some type of document which explains how to use `unmap.v` and Vivado?

Issue #749, opened 2019-05-21
Add support for loading timing information in v2x from SDF 


mithro:
We currently have a [Python parser for loading SDF files](https://github.com/SymbiFlow/python-sdf-timing). v2x is written in Python.

It would be good if for v2x we could use SDF files rather than timing attributes like found in the [CARRY4_VPR module](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/primitives/common_slice/carry/carry4_vpr.sim.v), see below;
```verilog
(* whitebox *)
module CARRY4_VPR(O0, O1, O2, O3, CO_CHAIN, CO_FABRIC0, CO_FABRIC1, CO_FABRIC2, CO_FABRIC3, CYINIT, CIN, DI0, DI1, DI2, DI3, S0, S1, S2, S3);
  parameter CYINIT_AX = 1'b0;
  parameter CYINIT_C0 = 1'b0;
  parameter CYINIT_C1 = 1'b0;


  (* DELAY_CONST_CYINIT="0.491e-9" *)
  (* DELAY_CONST_CIN="0.235e-9" *)
  (* DELAY_CONST_S0="0.223e-9" *)
  output wire O0;


  (* DELAY_CONST_CYINIT="0.613e-9" *)
  (* DELAY_CONST_CIN="0.348e-9" *)
  (* DELAY_CONST_S0="0.400e-9" *)
  (* DELAY_CONST_S1="0.205e-9" *)
  (* DELAY_CONST_DI0="0.337e-9" *)
  output wire O1;


  (* DELAY_CONST_CYINIT="0.600e-9" *)
  (* DELAY_CONST_CIN="0.256e-9" *)
  (* DELAY_CONST_S0="0.523e-9" *)
  (* DELAY_CONST_S1="0.558e-9" *)
  (* DELAY_CONST_S2="0.226e-9" *)
  (* DELAY_CONST_DI0="0.486e-9" *)
  (* DELAY_CONST_DI1="0.471e-9" *)
  output wire O2;
```

mithro:
This should probably be done before https://github.com/SymbiFlow/symbiflow-arch-defs/issues/748

mkurc-ant:
There is a PR: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/765

litghost:
This has been resolved by patching timing after arch is complete.

PR #765, opened 2019-05-23
Support for SDF timing import for V2X

mkurc-ant:
This PR adds a support for importing timings from SDF file to XML through V2X.

The SDF file path and SDF cell instance(s) should be explicitly given. If not then the script looks for a cell with the same name as the pb_type. The script can merge timings from multiple SDF cells.

Verilog and SDF port names must match one-to-one. And also all timings must be available in SDF otherwise the VPR throws an error during importing such incomplete arch def.

The V2X test suite was modified so now tests which use SDF timings can be defined. For now there is one test for the timing import which imports them for the CARRY4_VPR module. However the Verilog module definition had to be modified to match the SDF.


utils/vlog/tests/timing_logic/carry4_vpr.sim.v
```diff
@@ -0,0 +1,45 @@
...
+  assign CO1 = CI2;
+  assign CO2 = CI3;
+  assign CO3 = CI4;
+
+  assign O0 = CI0 ^ S0;
+  assign O1 = CI1 ^ S1;
+  assign O2 = CI2 ^ S2;
+  assign O3 = CI3 ^ S3;
+
+  //assign CO_CHAIN = CO3;
```
litghost: You'll need alias support for timing of this combintorial path.

utils/vlog/tests/timing_logic/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_file_target(FILE carry4_vpr.sim.v SCANNER_TYPE verilog)
+
+v2x_test_both(
+  NAME carry4_vpr
+  TOP_MODULE carry4_vpr
+  SDF_FILE CLBLL_L.sdf
+  SDF_CELLS SLICEL/CARRY4 SLICEL/CARRY4_AX SLICEL/CARRY4_BX SLICEL/CARRY4_CX SLICEL/CARRY4_DX
```
litghost: The way I modelled DIn lines was via the O5 cell, not the AX cells.

This strategy doesn't handled SLICEL vs SLICEM like @kgugala's patch does.  Are you coordinating with @kgugala to avoid duplication of work?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -54,68 +54,225 @@
...
-if args.includes:
-    for include in args.includes.split(','):
-        yosys.run.add_include(include)
+    If sdf_inst_name is None and there is only one instance of that cell in SDF
+    then that one is taken.
+    """
 
-vjson = yosys.run.vlog_to_json(args.infiles, flatten=False, aig=False)
-yj = YosysJSON(vjson)
+    def parse_timescale(timescale_str):
```
kgugala: this function is a part of python-sdf-timing lib. Why do you duplicate the code here instead of using it directly from the lib?

utils/vlog/vlog_to_pbtype.py
```diff
@@ -54,68 +54,225 @@
...
+        list.
+        """
+
+        collected_timings = {}
+
+        # For each path
+        for sdf_cell_path in sdf_cell_paths:
+
+            # Only cell name given
+            if "/" not in sdf_cell_path:
```
kgugala: in general, you should get the separator from sdf header

utils/vlog/vlog_to_pbtype.py
```diff
@@ -54,68 +54,225 @@
...
-        sys.exit(1)
+            # Get the SDF timing info for a particular CELL/INSTANCE
+            try:
+                sdf_cell = sdf["cells"][sdf_cell_name]
+            except KeyError:
+                print(
+                    "ERROR, the SDF file does not contain data for cell '{}'".
+                    format(sdf_cell_path)
+                )
+                continue
```
kgugala: why continue? If you want to skip the unexisting cells you can check for it with:
```
if sdf_cell_name in sdf["cells"]:
```

utils/vlog/tests/timing_logic/CMakeLists.txt
mkurc-ant: You are referring to a test example. It is not generic. In order for the whole SDF timing import to work we need to do significant changes to the architecture and how it is built by CMake.

The approach with timing import through V2X requires us to have a different model for each timing variant of a pb_type.

And yes, I am coordinating with @kgugala

utils/vlog/vlog_to_pbtype.py
```diff
@@ -54,68 +54,225 @@
...
+        exit(-1)
+
+    # Process SDF timing entires
+    for key, sdf_timing in sdf_timings.items():
+
+        sdf_inp = sdf_timing["from_pin"]
+        sdf_out = sdf_timing["to_pin"]
+
+        assert sdf_timing["is_incremental"] == False
+        assert sdf_timing["is_cond"] == False
```
kgugala: It'd be nice if you'd print some message here e.g. "Incremental delays are not supported"

utils/vlog/tests/timing_logic/CMakeLists.txt
litghost: Can @kgugala or you explain the relationship between this PR and https://github.com/SymbiFlow/symbiflow-arch-defs/pull/738 ?  If this PR was simply adding annotations that #738 then made concrete from SDF in a later step, I'd understand.  But this is directly emitting the values.

utils/vlog/vlog_to_pbtype.py
```diff
@@ -54,68 +54,225 @@
...
+                            sdf_timing["delay_paths"][sdf_variant][var] * scale
+                        )
+                    )
+
+            # "T_clock_to_Q"
+            else:
+                xml_tag = ET.SubElement(pb_type_xml, "T_clock_to_Q")
+                xml_tag.set("clock", "{}.{}".format(pb_name, pb_clk))
+                xml_tag.set("port", "{}.{}".format(pb_name, pb_out))
+                for var in ("min", "max"):
```
kgugala: this loop is identical for both the cases. Please make it common for both of them

utils/vlog/tests/timing_logic/carry4_vpr.sim.v
mkurc-ant: What do you mean by "alias support"? I was having a thought of giving such ports attributes like this: `(* SDF_ALIAS="C0" *)`

litghost: Ya, something like that.  Basically CO_CHAIN should use the same SDF timing path as CO3.

utils/vlog/tests/timing_logic/CMakeLists.txt
mkurc-ant: @litghost You are right, those two approaches are rather contradictory. This PR relates to the issue https://github.com/SymbiFlow/symbiflow-arch-defs/issues/749 by @mithro. However, the code can be easily modified to inject annotations instead of concrete values like in #738

litghost: I think #738 has the right idea to allow one pb_type (or v2x) to describe both the SLICEM and SLICEL variants (or ALUT, BLUT, etc) without the need to repeat the sim verilog.  I like the idea of auto-annotationing v2x, but let's emit the SDF tag for #738 to convert to a concrete value.  This will have additional value when do corner analysis or handling speed grades.

mkurc-ant: Ok, So maybe I'll make the v2x output annotations by default and add a switch which will make it output values. Concrete timing values in a Verilog model might be useful eg. for simulation.

mithro:
@mkurc-ant This was not quite what I was expected... This approach might work but I need to dig in deeper.

I was /actually/ suggesting that we add support for loading SDF files in Yosys. IE Using [a Python script pass](https://github.com/YosysHQ/yosys/blob/master/examples/python-api/pass.py) we load the SDF information into the same types of cells that the `specify` blocks in Verilog load into.

common/cmake/v2x.cmake
```diff
@@ -94,7 +98,7 @@ function(V2X)
       ${DEPENDS_LIST}
       ${symbiflow-arch-defs_SOURCE_DIR}/utils/vlog/vlog_to_pbtype.py
     COMMAND
-      ${CMAKE_COMMAND} -E env YOSYS=${YOSYS}  ${PYTHON3} ${symbiflow-arch-defs_SOURCE_DIR}/utils/vlog/vlog_to_pbtype.py ${TOP_ARG}
+      ${CMAKE_COMMAND} -E env YOSYS=${YOSYS} PYTHONPATH=${PYTHON_SDF_TIMING_DIR} ${PYTHON3} ${symbiflow-arch-defs_SOURCE_DIR}/utils/vlog/vlog_to_pbtype.py ${TOP_ARG} ${V2X_PB_TYPE_EXTRA_ARGS}
```
mithro: Please install python-sdf-timing into the conda environment so this isn't needed.

utils/vlog/tests/timing_logic/carry4_vpr.sim.v
mithro: Can you see what SDF supports for doing this? 

I think we probably should look at modifying the SDF file rather than inventing more new attributes?

litghost: @mithro : I don't think I agree.  We added the CO_CHAIN output due to a limitation in VPR around carry chain modelling.  The SDF file should be truth, and it makes little sense to modify the truth values to accommodate modelling issues, especially ones that limitations in VPR.

common/cmake/v2x.cmake
mkurc-ant: Done.

utils/vlog/vlog_to_pbtype.py
mkurc-ant: Done.

utils/vlog/vlog_to_pbtype.py
mkurc-ant: Done.

utils/vlog/vlog_to_pbtype.py
mkurc-ant: Done.

utils/vlog/vlog_to_pbtype.py
mkurc-ant: Done.

utils/vlog/vlog_to_pbtype.py
mkurc-ant: Done.

mkurc-ant:
@litghost I corrected the test so now it uses timing from CARRY4 and CARRY4_O5 as you requested.

mkurc-ant:
@litghost And I added the functionality to inject tags instead of numerical timing values.

litghost:
What is the state of this PR?

mkurc-ant:
Right now we are not importing timings through V2X and I don't know if we are going to do so. Either way this PR is based on an "old" version of V2X and I doubt if it can be merged directly with what @mithro and @kgugala are working on now.

I think we should close it. And if ever SDF import is needed we can reuse some of the code.

mithro:
I have added a GitHub issue for the new v2x repository here -> https://github.com/SymbiFlow/python-symbiflow-v2x/issues/9

mithro:
@kgugala / @mkurc-ant - I think a bunch of the stuff you did in this pull request could actually be merged into the https://github.com/SymbiFlow/python-symbiflow-v2x (at least the test could probably be reused?).

Issue #767, opened 2019-05-23
make test_python_utils fails on fresh checkout


mithro:
`make test_python_utils` fails if it is the only target you run from a fresh checkout. Looks like it is missing some dependency information?

See output below;
```
Downloading and Extracting Packages
py-1.8.0             | 65 KB     | ################################################################################################################################################### | 100%
attrs-19.1.0         | 32 KB     | ################################################################################################################################################### | 100%
more-itertools-4.3.0 | 83 KB     | ################################################################################################################################################### | 100%
pytest-4.5.0         | 354 KB    | ################################################################################################################################################### | 100%
pluggy-0.11.0        | 17 KB     | ################################################################################################################################################### | 100%
atomicwrites-1.3.0   | 9 KB      | ################################################################################################################################################### | 100%
wcwidth-0.1.7        | 17 KB     | ################################################################################################################################################### | 100%
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
make[4]: Leaving directory '/usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs-v2x/build'                                                                                   
Built target conda_pytest
make[4]: Entering directory '/usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs-v2x/build'                                                                                  
Scanning dependencies of target test_python_utils
make[4]: Leaving directory '/usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs-v2x/build'                                                                                   
make[4]: Entering directory '/usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs-v2x/build'                                                                                  
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.7.1, pytest-4.5.0, py-1.8.0, pluggy-0.11.0 -- /usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs-v2x/build/env/conda/bin/python                  
cachedir: .pytest_cache
rootdir: /usr/local/google/home/tansell/github/SymbiFlow/symbiflow-arch-defs-v2x/utils
collected 115 items / 6 errors / 109 selected                                                                                                                                                

=========================================================================================== ERRORS ===========================================================================================
______________________________________________________________________________ ERROR collecting clean_json5.py _______________________________________________________________________________
clean_json5.py:1: in <module>
    import pyjson5
E   ModuleNotFoundError: No module named 'pyjson5'
[...]

PR #769, opened 2019-05-23
cmake: Fix dependencies for Python targets.

mithro:
Fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/767

### DIFF
--- a/utils/CMakeLists.txt
+++ b/utils/CMakeLists.txt
@@ -2,7 +2,8 @@ get_target_property_required(PYTEST env PYTEST)
 
 add_custom_target(
   test_python_utils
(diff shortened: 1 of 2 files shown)


utils/CMakeLists.txt
```diff
@@ -2,7 +2,12 @@ get_target_property_required(PYTEST env PYTEST)
 
 add_custom_target(
   test_python_utils
-  DEPENDS conda_pytest intervaltree
+  DEPENDS
+    conda_pytest
```
litghost: Just use ``all_pip``

mithro: Done.

PR #775, opened 2019-05-25
xml: Adding tiles schema.

### LINKED ISSUE #774: tiles info needs to be added to XML schema
```
[100%] Generating arch.merged.lint.html
...
github/SymbiFlow/symbiflow-arch-defs-v2x/build/xc7/archs/artix7/devices/xc7a50t-basys3-roi-virt/arch.unique_pack.xml:331: element tiles: Schemas validity error : Element 'tiles': This element is not expected.                                                         
```

mithro:
This change needs validation, I've never done xsd before.

Fixes #774

### DIFF
--- a/common/xml/fpga_architecture.xsd
+++ b/common/xml/fpga_architecture.xsd
@@ -584,6 +584,41 @@
     <xs:attribute name="name" type="xs:string" use="required"/>
   </xs:complexType>
 
+  <xs:complexType name="map">
+    <xs:attribute name="from" type="xs:string" use="required"/>
+    <xs:attribute name="to" type="xs:string" use="required"/>
+    <xs:attribute name="num_pins" type="xs:string"/>
+  </xs:complexType>
+
+  <xs:complexType name="tile_mode">
+    <xs:choice maxOccurs="unbounded" minOccurs="0">
+      <xs:element name="map" type="map"/>
+    </xs:choice>
+    <xs:attribute name="name" type="xs:string" use="required"/>
+  </xs:complexType>
+
+  <xs:complexType name="equivalent_tiles">
+    <xs:choice maxOccurs="unbounded" minOccurs="0">
+      <xs:element name="mode" type="tile_mode"/>
+    </xs:choice>
+  </xs:complexType>
+
+  <xs:complexType name="tile">
+    <xs:choice maxOccurs="unbounded" minOccurs="0">
+      <xs:element name="equivalent_tiles" type="equivalent_tiles"/>
+      <xs:element name="pinlocations" type="pinlocations"/>
+      <xs:element name="fc" type="fc"/>
+      <xs:element name="switchblock_locations" type="switchblock_locations"/>
+    </xs:choice>
+    <xs:attribute name="name" type="xs:string" use="required"/>
+  </xs:complexType>
+
+  <xs:complexType name="tiles">
+    <xs:sequence>
+      <xs:element name="tile" type="tile" maxOccurs="unbounded"/>
+    </xs:sequence>
+  </xs:complexType>
+
   <xs:complexType name="pb_type">
     <xs:choice maxOccurs="unbounded" minOccurs="0">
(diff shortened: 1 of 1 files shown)


mithro:
Hi @Laksen, 

Could you take a look over these changes to the XML schema definition you wrote? While the new XML now passes the validation I'm unsure if I've done the right thing.

The schema is being changed as shown in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/559 and [this Google Doc](https://docs.google.com/document/d/1LQ0H2C9IFnCVtSuaniCs1dDmBoMBWwBQLKAGuo_kSc0/edit#heading=h.uo6gk3q6tsrk). 

Thanks!

Tim '@mithro' Ansell

mithro:
@acomodi - After https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/559 should should `<pb_type>` still have `<fc>` or `<pinlocations>` tags?

acomodi:
@mithro There is also an open PR in SYmbiFlow VTR fork (https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/36) that moves the top level-related pb_type tags into the tiles one and deals with equivalent tiles. Until it gets merged, the tiles tags are not parsed in VtR, hence we cannot delete the `<fc>`, `<pinlocations>` and `<switchblock>` from pb_types.
I could split the PR in two:
- Equivalent tiles
- Move top-level pb_type tags to the `<tiles>` tag.

This way we could merge the second one and start adopting the `<tiles>` tags in SymbiFlow architectures.

This would also require to add a parsing step on the ice40 architectures. I've created a script that should be slightly modified in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/583. I could integrate that in the cmake steps to modify all the architectures which do not make use of [prjxray_physical_tile_import.py](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/utils/prjxray_physical_tile_import.py), which defines the `<tiles>` tag for xc7 architectures.

common/xml/fpga_architecture.xsd
```diff
@@ -584,6 +584,40 @@
     <xs:attribute name="name" type="xs:string" use="required"/>
   </xs:complexType>
 
+  <xs:complexType name="map">
+    <xs:attribute name="from" type="xs:string" use="required"/>
+    <xs:attribute name="to" type="xs:string" use="required"/>
+  </xs:complexType>
```
acomodi: For now, here there should be also a `num_pin` attribute which is optional. I still have to modify VtR XML parser to allow the definition of the port range (e.g. DIN1[3:0] maps to DIN[7:4] and DIN2[3:0] maps to DIN[3:0]).

common/xml/fpga_architecture.xsd
```diff
@@ -584,6 +584,40 @@
...
+    <xs:choice maxOccurs="unbounded" minOccurs="0">
+      <xs:element name="mode" type="tile_mode"/>
+    </xs:choice>
+  </xs:complexType>
+
+  <xs:complexType name="tile">
+    <xs:choice maxOccurs="unbounded" minOccurs="0">
+      <xs:element name="equivalent_tiles" type="equivalent_tiles"/>
+      <xs:element name="pinlocations" type="pinlocations"/>
+      <xs:element name="fc" type="fc"/>
```
acomodi: I am unsure if it is possible, but it would be nice to set `fc` as required if there is no `default_fc` defined in the architecture.

common/xml/fpga_architecture.xsd
mithro: `num_pin` or `num_pins`?

common/xml/fpga_architecture.xsd
mithro: I think that didn't happen before, so lets leave that for another time.

common/xml/fpga_architecture.xsd
acomodi: Right, `num_pins` is correct

PR #783, opened 2019-05-27
Small improvements to EBLIF generation

mithro:
* Add support for luts.
 * Ignore ports marked as carries.
 * Correctly generate connections for the clock pins. 
 * Improve composability around `<pb_type>` attributes.


utils/vpr_pbtype_to_eblif.py
```diff
@@ -85,15 +123,26 @@ def main(args):
         args.pb_type
     )
 
-    pbtype_name, clocks, inputs, outputs = ports(pbtype_leaf)
+    blif_model, clocks, inputs, outputs, carry = ports(pbtype_leaf)
+    blif_model = pbtype_leaf.attrib['blif_model']
+
+    if 'subckt' in blif_model:
+        name = blif_model[len('.subckt '):]
```
mkurc-ant: I'd rather use a regexp match. This will crash eg. if there is no explicit ".subkt " in the string.

utils/vpr_pbtype_to_eblif.py
```diff
@@ -85,15 +123,26 @@ def main(args):
         args.pb_type
     )
 
-    pbtype_name, clocks, inputs, outputs = ports(pbtype_leaf)
+    blif_model, clocks, inputs, outputs, carry = ports(pbtype_leaf)
```
mkurc-ant: Since `blif_model` is overwritten in the next line, I'd use here `_, clocks, inputs, outputs, carry = ...`. Or rename it to `pbtype_name` as the function returns it.

utils/vpr_pbtype_to_eblif.py
mithro: Fixed this and added some tests for this type of thing.

utils/vpr_pbtype_to_eblif.py
mithro: Done.

mkurc-ant:
Ok, LGTM

Issue #564, opened 2019-04-09
cmake -DUSE_CONDA=FALSE is broken


HackerFoo:
```
sf-shell-chrootenv:dusty@hank:~/src/symbiflow-arch-defs/build$ cmake -DUSE_CONDA=FALSE ..
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.7.2", minimum required is "3") 
CMake Error at make/env.cmake:314 (message):
  Could not find program progressbar2.
Call Stack (most recent call first):
  CMakeLists.txt:60 (add_conda_pip)
```

[make/env.cmake:314](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/make/env.cmake#L314)

I've tried commenting out those errors, but then I run into trouble with lines like:

```
get_target_property_required(BLAH_TARGET env BLAH_TARGET)
```

and removing `_required` leads to a dead end where those things really are required.

elmsfu:
I think the logic [here](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/make/env.cmake#L308-L318) is not quite right. If doesn't look at the `NO_EXE` flag. It should probably be checking if python can import those packages instead of `find_program` if `NO_EXE` is provided.

You can probably work around by setting environment variables before running `cmake`

mithro:
I think when using conda, the system is checking for progressbar2 being installed in the conda environment?

mithro:
`python -c "import progressbar2"; echo $?` could be the correct way to check the module is available?

litghost:
I believe this can be closed?

HackerFoo:
Yes, fixed with https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1521.

PR #787, opened 2019-05-28
Add Python lint target

elmsfu:
This adds a target that will display all linting errors. Once errors are addressed, this should be added to CI and the `--zero-exit` flag should be removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,21 +135,23 @@ add_subdirectory(testarch)
 add_subdirectory(tests)
 
 get_target_property_required(YAPF env YAPF)
+get_target_property_required(YAPF_TARGET env YAPF_TARGET)
 add_custom_target(
   check_python
-  DEPENDS conda_yapf
+  DEPENDS ${YAPF_TARGET}
   COMMAND ${symbiflow-arch-defs_SOURCE_DIR}/common/python_health_helper.sh -c -y ${YAPF}
   )
 add_custom_target(
   format_python
-  DEPENDS conda_yapf
+  DEPENDS ${YAPF_TARGET}
   COMMAND ${symbiflow-arch-defs_SOURCE_DIR}/common/python_health_helper.sh -f -y ${YAPF}
   )
 
 get_target_property_required(FLAKE8 env FLAKE8)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -134,20 +138,19 @@ get_target_property_required(YAPF env YAPF)
...
-  COMMAND ${symbiflow-arch-defs_SOURCE_DIR}/common/yapf_helper.sh -c -y ${YAPF}
-  )
-
-add_custom_command(
-  TARGET format_python
-  COMMAND ${symbiflow-arch-defs_SOURCE_DIR}/common/yapf_helper.sh -f -y ${YAPF}
+get_target_property_required(FLAKE8 env FLAKE8)
+add_custom_target(
+  lint_python
+  DEPENDS conda_flake8
```
litghost: Rather than depending on ``conda_flake8`` directly, use ``get_target_property(FLAKE8_TARGET env FLAKE8_TARGET)``.  Directly depending on ``conda_XXX`` is part of the cause of https://github.com/SymbiFlow/symbiflow-arch-defs/issues/564

common/python_health_helper.sh
```diff
@@ -17,6 +17,11 @@ run() {
   return $res
   }
 
+lint_file() {
+  run "$1 $2" "lint failed on $2"
```
litghost: Rather than have $1, $2, etc everywhere, please assign $1, ... to named variables with meaningful names.

common/python_health_helper.sh
```diff
@@ -4,7 +4,7 @@ usage() {
   if [ -n "$1" ]; then
```
mithro: Can you log an issue to rewrite this shell script in Python?

CMakeLists.txt
elmsfu: Done

common/python_health_helper.sh
elmsfu: Will do. But I don't see particular value in that exercise.

common/python_health_helper.sh
elmsfu: Done

PR #789, opened 2019-05-28
xc7: dff: split dff into {FF|REG}_INIT and update timings

kgugala:
This PR splits FFs definitions in the arch.xml so it corresponds with the actual FPGA arch. Also timings definitions are updated, but they do not have correct delay values.

This PR was originally a part of #738


kgugala:
CI is green. I'll merge it.

PR #790, opened 2019-05-28
Removed blackbox attribute.

litghost:
This was causing simulations to output 1'bx for all CARRY4_VPR outputs.

This does not fix DRAM test error output problem, but it does fix autosim for cases where a sim model exists.


mithro:
Should this be a `(* whitebox *)`?

litghost:
> Should this be a `(* whitebox *)`?

I'm looking into that.

litghost:
> Should this be a `(* whitebox *)`?

No, whitebox has the same behavior.

PR #794, opened 2019-05-29
Add docstrings to graph2 channel

elmsfu:
- Change namedtuples to classes with docstrings
 - Address Flake8 for these files

PR #795, opened 2019-05-29
utils: address flake8 errors

elmsfu:
Address various issues. Added some ignores so it doesn't conflict with yapf style

Ignored/Exceptions:
 - E123 closing bracket does not match indentation of opening bracket’s line
 - E125 continuation line with same indent as next logical line
 - E251 unexpected spaces around keyword / parameter equals
 - W503 line break before binary operator
 - E731 do not assign a lambda expression, use a def
 - E741 [Do not use variables named 'l', 'o', or 'i'](https://www.python.org/dev/peps/pep-0008/#names-to-avoid)

Also changed v2x to add `utils` to `PYTHONPATH`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,4 +1,8 @@
 [flake8]
 max-line-length = 100
 max-doc-lentgh = 100
-ignore = E123, E125, E251, W503, E731, E741
+ignore =
+  E123, # closing bracket indentation mismatch  yapf does this a lot
+  E125, # continuation line indent              yapf does this a lot
+  W503, # line break before binary operator     yapf breaks lines here
+  W504, # line break after binary operator      yapf breaks lines here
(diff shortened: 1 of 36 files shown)


elmsfu:
The most important feedback would be on what errors to ignore globally vs. adding one off exceptions. I already have a hand full of E501 for docstrings.

elmsfu:
This should wait for #794

tox.ini
```diff
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 100
 max-doc-lentgh = 100
+ignore = E123, E125, E251, W503, E731, E741
```
litghost: Can you add a detailed comment about why each of these are ignored?  Can be a simple as "yapf formatting causing this check to be violated", etc.

elmsfu: Done

PR #800, opened 2019-05-30
xc7: Fix alignment in ff.pb_type.xml

mithro:
Alignment was broken in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/738


mithro:
@kgugala - FYI

PR #803, opened 2019-05-31
Add utilities for converting from/to rr graph inodes and routing data…

litghost:
…base

This is useful for diagnosing router behavior.

The amount of time to write the node_map via pickle is very small (~100 msec).

### DIFF
--- a/xc7/utils/prjxray_find_inode.py
+++ b/xc7/utils/prjxray_find_inode.py
@@ -1,0 +1,55 @@
+""" Find an rr node by specifying a wire name (e.g. INT_R_X3Y149/WL1BEG0).
+
+This is useful for examining router behavior.  E.g. what which inode represents
+this wire?
+
+"""
+import argparse
+import pickle
+import sqlite3
+from lib.rr_graph.graph2 import NodeType
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('--rrgraph_node_map', required=True)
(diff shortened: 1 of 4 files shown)


xc7/utils/prjxray_lookup_inode.py
```diff
@@ -0,0 +1,76 @@
+""" Convert rrgraph inode back into graph_node_pkey and print some information.
```
mithro:
```python
#!/usr/bin/env python3
```

xc7/utils/prjxray_find_inode.py
```diff
@@ -0,0 +1,55 @@
+""" Find an rr node by specifying a wire name (e.g. INT_R_X3Y149/WL1BEG0).
```
mithro:
```python
#!/usr/bin/env python3
```

xc7/utils/prjxray_routing_import.py
```diff
@@ -694,6 +700,11 @@ def main():
                 import_graph_edges(conn, graph, node_mapping)
             )
 
+        print('{} Writing node map.'.format(now()))
+        with open(args.write_rr_node_map, 'wb') as f:
+            pickle.dump(node_mapping, f)
```
mithro: Is using a pickle verse a json file more performant?

litghost: Pretty much.  This is a diagnostic leaf, I wanted a low impact both in runtime and disk space.

PR #813, opened 2019-06-05
Add missing 2-bit width to RAMB36E1 

HackerFoo:


HackerFoo:
They just don't use the parity bits, but it affects how the address space is mapped.

HackerFoo:
I removed the 8/16/32-bit widths, but 2 was missing, so I added that, and also eliminated some of the repetition.

PR #815, opened 2019-06-06
travis: add -j2 option to build in parallel and speed it up

elmsfu:


elmsfu:
Trying this to see if it really has significant impact on build time.

mithro:
@elmsfu I think we want something like `make -k -j2 $TARGET || make -j1 $TARGET`

mithro:
This is what I did for the vtr repository;

https://github.com/mithro/vtr-verilog-to-routing/blob/034d34e17d7debb7af8b818791c77ba14838bb01/.github/travis/build.sh#L8-L19

```shell

start_section "vtr.build" "${GREEN}Building..${NC}"
export FAILURE=0
make -k CMAKE_PARAMS="-DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on" -j2 || export FAILURE=1
end_section "vtr.build"

# When the build fails, produce the failure output in a clear way
if [ $FAILURE -ne 0 ]; then
	start_section "vtr.failure" "${RED}Build failure output..${NC}"
	make CMAKE_PARAMS="-DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on" -j1
	end_section "vtr.failure"
	exit 1
fi
```

elmsfu:
> @elmsfu I think we want something like `make -k -j2 $TARGET || make -j1 $TARGET`

@mithro So the `|| make -j1` is to get linear output on failure?

I like the idea, but there is a good chance on travis that the wouldn't happen within the 50 minute window. I think we should merge this to get builds not timing out.

mithro:
> > @elmsfu I think we want something like `make -k -j2 $TARGET || make -j1 $TARGET`
> 
> @mithro So the `|| make -j1` is to get linear output on failure?

Correct.

> I like the idea, but there is a good chance on travis that the wouldn't happen within the 50 minute window. I think we should merge this to get builds not timing out.

`make -k || make -j1` should take almost the same time as `make`. The `make -j1` will only run if a task fails and it will only rerun at most 1 failing task.

elmsfu:
Will open a new PR with different approach.

PR #817, opened 2019-06-07
travis: wrap make in function to handle parallel building

elmsfu:
Have travis build in parallel and handle rebuilding to display error messages clearly as recommended in #815

### DIFF
--- a/.github/travis/common.sh
+++ b/.github/travis/common.sh
@@ -51,6 +51,32 @@ if [ -z "$DATESTR" ]; then
 	fi
 fi
 
+function make_target() {
+  target=$1
+
+  if [ ! -v MAKE_JOBS ]; then
+    export MAKE_JOBS=$(nproc)
+    echo "Setting MAKE_JOBS to $MAKE_JOBS"
+  fi;
+
+  export VPR_NUM_WORKERS=${MAKE_JOBS}
+
+  start_section "symbiflow.$target" "$2"
+  make_status=0
+  make -k -j${MAKE_JOBS} $target || make_status=$?
+  end_section "symbiflow.$target"
+
+  # When the build fails, produce the failure output in a clear way
+  if [ ${MAKE_JOBS} -ne 1 -a $make_status -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


.github/travis/common.sh
```diff
@@ -51,6 +51,32 @@ if [ -z "$DATESTR" ]; then
 	fi
 fi
 
+function make_target() {
+  target=$1
+
+  if [ ! -v MAKE_JOBS ]; then
+    export MAKE_JOBS=$(nproc)
```
mithro: I think you might want to force this to 2 on Travis?

.github/travis/common.sh
```diff
@@ -51,6 +51,32 @@ if [ -z "$DATESTR" ]; then
 	fi
 fi
 
+function make_target() {
+  target=$1
+
+  if [ ! -v MAKE_JOBS ]; then
```
mithro: Might make sense to just call this `CPUS`?

.github/travis/common.sh
```diff
@@ -51,6 +51,32 @@ if [ -z "$DATESTR" ]; then
...
+    export MAKE_JOBS=$(nproc)
+    echo "Setting MAKE_JOBS to $MAKE_JOBS"
+  fi;
+
+  export VPR_NUM_WORKERS=${MAKE_JOBS}
+
+  start_section "symbiflow.$target" "$2"
+  make_status=0
```
mithro: Any reason this is lowercase?

.github/travis/common.sh
elmsfu: I detects the correct number. Not sure why we'd need to force it at the moment.

.github/travis/common.sh
elmsfu: If we want to override it I think this name makes more sense. Longer term, I think the way kokoro handles it may be more flexible. They could be unified, but don't think it's worth that effort at the moment.

.github/travis/common.sh
elmsfu: local to the function, but no strong feeling on it.

PR #828, opened 2019-06-13
devices.cmake: quiet script registration

kgugala:


kgugala:
@mithro it seems that cocoro tests were not run at all here and the CI is green

acomodi:
@kgugala I forced a kokoro-run and it CI went green. I'll proceed with merging.

PR #858, opened 2019-06-21
cmake: update VTR and add new command line options

acomodi:

This PR adds the option to allow dangling combinational nodes to xc7.

CI will fail as the option is not yet in the conda package. Once it is added I will re-run kokoro and merge if green.

### DIFF
--- a/common/cmake/devices.cmake
+++ b/common/cmake/devices.cmake
@@ -922,10 +922,12 @@ function(ADD_FPGA_TARGET)
     set(VPR_ARCH_ARGS "")
   endif()
 
+  set(OUT_NOISY_WARNINGS ${OUT_LOCAL}/noisy_warnings.log)
   separate_arguments(
     VPR_BASE_ARGS_LIST UNIX_COMMAND "${VPR_BASE_ARGS}"
     )
   list(APPEND VPR_BASE_ARGS_LIST --route_chan_width ${ROUTE_CHAN_WIDTH})
+  list(APPEND VPR_BASE_ARGS_LIST --suppress_warnings ${OUT_NOISY_WARNINGS},sum_pin_class:check_unbuffered_edges:load_rr_indexed_data_T_values:check_rr_node:trans_per_R)
   separate_arguments(
     VPR_EXTRA_ARGS_LIST UNIX_COMMAND "${VPR_EXTRA_ARGS}"
     )
--- a/common/cmake/devices.cmake
+++ b/common/cmake/devices.cmake
@@ -922,17 +922,19 @@ function(ADD_FPGA_TARGET)
     set(VPR_ARCH_ARGS "")
   endif()
 
-  set(OUT_NOISY_WARNINGS ${OUT_LOCAL}/noisy_warnings.log)
   separate_arguments(
     VPR_BASE_ARGS_LIST UNIX_COMMAND "${VPR_BASE_ARGS}"
     )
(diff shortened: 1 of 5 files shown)


acomodi:
I have added all the necessary command options to VPR and updated it to the latest version.

Lattice iCE40 will fail while routing picosoc until https://github.com/SymbiFlow/symbiflow-arch-defs/pull/693 gets merged.

xc7/make/arch_define.cmake
```diff
@@ -26,6 +26,8 @@ function(ADD_XC7_ARCH_DEFINE)
       --clustering_pin_feasibility_filter off
       --disable_check_route on
       --strict_checks off
+      --allow_dangling_combinational_nodes on
+      --disable_errors check_unbuffered_edges:check_route
```
litghost: Can we also disable the warnings that generate ~250k warnings?

acomodi:
> Lattice iCE40 will fail while routing picosoc until #693 gets merged.

@litghost ^

>LGTM, small request to disable some verbose warnings that makes VPR logs hard to read. You'll know it when you see it.

Sounds good.

litghost:
> Lattice iCE40 will fail while routing picosoc until #693 gets merged.

Is this just because of the new router flag?  Lets add that flag to this PR so it goes green?

common/cmake/devices.cmake
```diff
@@ -904,10 +904,12 @@ function(ADD_FPGA_TARGET)
     set(VPR_ARCH_ARGS "")
   endif()
 
+  set(OUT_NOISY_WARNINGS ${OUT_LOCAL}/noisy_warnings.log)
   separate_arguments(
     VPR_BASE_ARGS_LIST UNIX_COMMAND "${VPR_BASE_ARGS}"
     )
   list(APPEND VPR_BASE_ARGS_LIST --route_chan_width ${ROUTE_CHAN_WIDTH})
+  list(APPEND VPR_BASE_ARGS_LIST --suppress_warnings ${OUT_NOISY_WARNINGS},sum_pin_class:check_unbuffered_edges:load_rr_indexed_data_T_values:check_rr_node:trans_per_R)
```
litghost: Let's do this in the arch define, rather than globally.

xc7/make/arch_define.cmake
acomodi: Done, the only thing is that I have specified the `suppress_warnings` option in `device.cmake` as they could be suppressed also for ice40 I suppose.
Another reason is that I didn't manage to get cmake to set the noisy_warnings.log output file to the build directory destination.

common/cmake/devices.cmake
acomodi: How can I set the output file in the arch_define.cmake to be in the build directory of a specific test?

litghost: Two ways I can see.  One is to model VPR_ARCH_ARGS like PLACE_TOOL_CMD (see https://github.com/SymbiFlow/symbiflow-arch-defs/pull/858/files/c70a93d58f07f764e45ff6ad9fbc017fd4eac8ff..ed4d055b1066571749bf9fb6902474c6e829c5db#diff-2d565728ea2fbe9d3a017347718a0ba5R969 ), the other is to fully model warning suppression as a first class argument to define_arch

acomodi: I think the first one could be the most correct way to go as it maintains more consistency with the other commands definitions. I'll proceed with that one then.

litghost:
Needs to be merged with master

acomodi:
> Needs to be merged with master

Done, I'll merge as soon as CI gets green.

PR #861, opened 2019-06-26
Imported bram timings

acomodi:

This PR adds timings to BRAMs.

I have taken the True Dual Port timings from a local build of timing fuzzer to test it. `prjxray-db` has to be updated with the most recent timing values after https://github.com/SymbiFlow/prjxray/commit/ca6bbee1931cc06152f0f94463f830d437160ff7.

There are some timing values which are not used in our architecture, such as `REGCLKB` and `REGCLKA`, so I have left them and set a fake timing.


acomodi:
CI failed as there db does not contain timing information on BRAMs yet.
```
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/utils/update_arch_timings.py", line 160, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/utils/update_arch_timings.py", line 138, in main
    bel_timings = get_bel_timings(tq, timings, bels)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/utils/update_arch_timings.py", line 84, in get_bel_timings
    return find_timings(timings, bel, location, site, bels)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/utils/update_arch_timings.py", line 63, in find_timings
    cell = mergedicts(timings['cells'][ct][instance], cell)
KeyError: 'RAMBFIFO36E1'
```

I have also tried to locally run on HW murax with this new timing information, but it did not properly work. I need to investigate the reason. (I have also double-checked that current master builds a correct murax SoC).

mithro:
The upstream database should have BRAM timings now. Can you check?

acomodi:
@mithro Yes indeed, thanks. I will run kokoro again, but even if it turns green these PR still needs time as it breaks at least Murax on HW.

What happens is that no led is switched on and UART is not transmitting/receiving anything when building with this PR.

acomodi:
Picosoc seems to work (even though the baud rate must be set to 460800 now instead of 115200) until I try to access the main menu. As soon as I do something, everything halts and the leds stop blinking.

acomodi:
A curious thing is that by integrating the usage of the equivalent tiles, picosoc works perfectly and does not present the issue described in the [previous comment](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/861#issuecomment-506729701).

acomodi:
CI is green, merging.

PR #870, opened 2019-07-02
Routing BELs timings

kgugala:
This PR adds routing bels timings to common slice. It will not work until https://github.com/SymbiFlow/prjxray/issues/890 is merged and new DB generated with it


kgugala:
@mithro We'd need updated DB (with routing BELs timings) before we can land this PR

xc7/primitives/common_slice/common_slice.pb_type.xml
```diff
@@ -477,7 +539,8 @@
       </metadata>
     </mux>
     <mux name="CARRY_DI0" input="COMMON_SLICE.AO5 COMMON_SLICE.AX" output="CARRY4_VPR.DI0" >
-      <delay_constant max=".105e-9" in_port="AX" out_port="CARRY4_VPR.DI0" />
```
litghost: This is not exactly right.  The previous delay was **not** the CY mux delay, but instead the difference between CARRY4_[ABCD]X and CARRY4_O5.

mithro:
Updated database has been pushed at https://github.com/SymbiFlow/prjxray-db/commit/2e4d3dc3b2355c7b502ef778ff1ce8585e9d3889

mithro:
I rebased this change onto symbiflow-arch-defs which has the latest prjxray-db including the routing bel timing information.

xc7/primitives/common_slice/common_slice.pb_type.xml
kgugala: I have removed those changes from the PR. This will need some more logic to calculate those delays on the fly. I don't want to stall this PR, so I'll open a new one with those

PR #871, opened 2019-07-02
Add QoR tool

litghost:
This adds a simple QoR summary tool and binds Kokoro to gather this summary and the other VPR outputs.

### DIFF
--- a/utils/print_qor.py
+++ b/utils/print_qor.py
@@ -1,0 +1,82 @@
+""" Gathers statistics from VPR runs, like pack/place/route runtimes and Fmax.
+"""
+import argparse
+import csv
+import os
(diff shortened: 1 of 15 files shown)


utils/print_qor.py
```diff
@@ -0,0 +1,82 @@
...
+    critical_path : str
+        Critical path delay in nsec
+    fmax : str
+        Fmax in MHz.
+
+    """
+    try:
+        with open(fname, 'r') as f:
+            for l in f:
+                if l.startswith('Final critical path:'):
```
mithro:
```suggestion
                if not l.startswith('Final critical path:'):
                    continue
```
etc

Issue #848, opened 2019-06-18
Disallow multiple top-level tags in architecture tag


duck2:
The arch file reader implementation looks for one top level tag of each type:
```c++
        /* Process models */
        Next = get_single_child(architecture, "models", loc_data);
        [...]
```

The schema disagrees:
```xsd
  <xs:element name="architecture">
    <xs:complexType>
      <xs:choice minOccurs="0" maxOccurs="unbounded">
        <xs:element name="models" type="models"/>
        <xs:element name="layout" type="layout"/>
        <xs:element name="device" type="device"/>
        <xs:element name="switchlist" type="switchlist"/>
        <xs:element name="switchblocklist" type="switchblocklist"/>
        <xs:element name="segmentlist" type="segmentlist"/>
        <xs:element name="power" type="global_power"/>
        <xs:element name="clocks" type="clocks"/>
        <xs:element name="directlist" type="directlist"/>
        <xs:element name="tiles" type="tiles"/>
        <xs:element name="complexblocklist" type="complexblocklist"/>
      </xs:choice>
    </xs:complexType>
  </xs:element>
```

mithro:
You seem to be correct.

I think an architecture requires a lot of these `<xs:element>` objects?

mithro:
Fixed in #872.

Issue #847, opened 2019-06-18
Fix values in switch type to be xs:float rather than xs:double in the schema


duck2:
The docs and the implementation says the R, Cin, Cout... values in a switch type are floats, but the schema defines them as doubles.

The schema also defines a `Cinternal` attribute which isn't present anywhere in the code or the docs.

```xsd
  <xs:complexType name="switch">
    <xs:sequence>
      <xs:element name="Tdel" minOccurs="0" maxOccurs="unbounded">
        <xs:complexType>
          <xs:attribute name="num_inputs" type="xs:int" use="required"/>
          <xs:attribute name="delay" type="xs:double" use="required"/>
        </xs:complexType>
      </xs:element>
    </xs:sequence>
    <xs:attribute name="type" type="switch_type" use="required"/>
    <xs:attribute name="name" type="xs:string" use="required"/>
    <xs:attribute name="R" type="xs:double" use="required"/>
    <xs:attribute name="Cin" type="xs:double" use="required"/>
    <xs:attribute name="Cinternal" type="xs:double"/>
    <xs:attribute name="Cout" type="xs:double"/>
    <xs:attribute name="Tdel" type="xs:double"/>
    <xs:attribute name="buf_size" type="buf_size" default="auto"/>
    <xs:attribute name="mux_trans_size" type="xs:double"/>
    <xs:attribute name="power_buf_size" type="xs:double"/>
  </xs:complexType>
```

mithro:
`Cinternal` is in the process of being added. See https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/550 and https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/607

mithro:
Fixed in #872.

Issue #846, opened 2019-06-18
Disallow multiple x and y tags in chan_width_str in the schema


duck2:
According to the [implementation](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/libarchfpga/src/physical_types.h#L1132), there can only be one `x` and one `y` tag in a `chan_width_distr` tag. The schema disagrees:

```xsd
    <xs:choice maxOccurs="unbounded">
      <xs:element name="x" type="chan_dist"/>
      <xs:element name="y" type="chan_dist"/>
    </xs:choice>
```

mithro:
I think you are correct.

mithro:
Fixed in #872.

Issue #845, opened 2019-06-18
Remove device_layout tag from the schema


duck2:
`device_layout` is no longer accepted in an arch file, but it still exists in the schema.

```xsd
  <xs:complexType name="layout">
    <xs:choice minOccurs="1">
      <xs:element name="auto_layout" type="auto_layout" maxOccurs="1"/>
      <xs:element name="fixed_layout" type="fixed_layout" maxOccurs="unbounded"/>
      <xs:element name="device_layout" type="fixed_layout" maxOccurs="unbounded"/> <!-- Apparently a legacy name -->
    </xs:choice>
  </xs:complexType>
```

mithro:
Fixed in #872.

PR #872, opened 2019-07-03
Revise arch schema

duck2:
Fixes the following issues found in the schema when generating a parser using [uxsdcxx](https://github.com/duck2/uxsdcxx/);
- [x] #841 - Update the XML schema to not accept offsets on metadata
- [x] #842 - Add is_non_clock_global attribute to model ports in the XML schema
- [x] #843 - Disallow multiple input_ports in a model tag in the schema
- [x] #844 - Add metadata to layout info tags in the schema
- [x] #845 - Remove device_layout tag from the schema
- [x] #846 - Disallow multiple x and y tags in chan_width_str in the schema
- [x] #847 - Fix values in switch type to be xs:float rather than xs:double in the schema
- [x] #848 - Disallow multiple top-level tags in architecture tag

### DIFF
--- a/common/xml/fpga_architecture.xsd
+++ b/common/xml/fpga_architecture.xsd
@@ -12,10 +12,6 @@
     <xs:simpleContent>
       <xs:extension base="xs:string">
         <xs:attribute name="name" type="xs:string" use="required"/>
-        <xs:attribute name="xoffset" type="xs:int"/>
-        <xs:attribute name="x_offset" type="xs:int"/>
-        <xs:attribute name="yoffset" type="xs:int"/>
-        <xs:attribute name="y_offset" type="xs:int"/>
       </xs:extension>
     </xs:simpleContent>
   </xs:complexType>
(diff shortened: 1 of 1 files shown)


common/xml/fpga_architecture.xsd
```diff
@@ -1,836 +1,873 @@
...
-        <xs:element name="clocks" type="clocks"/>
-        <xs:element name="directlist" type="directlist"/>
-        <xs:element name="tiles" type="tiles"/>
-        <xs:element name="complexblocklist" type="complexblocklist"/>
-      </xs:choice>
-    </xs:complexType>
-  </xs:element>
-
-</xs:schema>
+<?xml version="1.0"?>
```
mithro: Why is every single line modified?

duck2: Old schema had CRLF, mine has LF.

duck2: Considering the rest of the repository uses LF newlines, maybe I shouldn't fix this? But it would be good to divide into two commits, first newline changes, then real changes.

mithro: Please don't change line endings.

mithro:
@duck2 - Can you update the pull request description to include the titles for each issue (rather than just the link + number)?

duck2:
> @duck2 - Can you update the pull request description to include the titles for each issue (rather than just the link + number)?

done!

mithro:
Merged! Yay!

mithro:
Also closed all those bugs too!

PR #938, opened 2019-08-12
Fixed fasm2bels to work with new IOB/IOI fasm tags

mkurc-ant:
This PR refactors how IOBs are handled by fasm2bels thus allowing it to work with latest prjxray db (866e227).

Since the IOSTANDARD (and sometimes DRIVE) cannot be decoded from the fasm in an unambigous way, there is a mechanizm to provide those parameters externally (for each IOB). Externally specified parameters are compared against fasm disassembly to check if they are sane.


xc7/fasm2bels/ioi_models.py
```diff
@@ -0,0 +1,96 @@
...
+        iob_site = sites[0]
+    else:
+        iob_site = sites[1 - int(site[-1])]
+
+    return iob_site
+
+
+def process_idelay(top, features):
+    # TODO: Support IDELAY
+    pass
```
litghost: Problaly worth having an assert if [IDELAY.IN_USE](https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/segbits_lioi3.db#L13) is set?

xc7/fasm2bels/iob_models.py
```diff
@@ -126,21 +160,70 @@ def process_iob(top, iob):
...
+
+    # Buffer direction
+    is_input = has_feature_with_part(site, "IN") or has_feature_with_part(
+        site, "IN_ONLY"
+    )
+    is_inout = has_feature_with_part(site, "IN") and has_feature_with_part(
+        site, "DRIVE"
+    )
+    is_output = not has_feature_with_part(site, "IN") and \
+        has_feature_with_part(site, "DRIVE")
```
litghost: Should sanity check this (e.g. only one is set).

mkurc-ant:
@litghost I am open to suggestions how to specify iostandard per port. Right now I did that through a simple JSON file which references IOB site locations (by XY).

As long as we have harnesses I can prepare such a JSON file for each one. But once we do the ROI breakout each design would need a separate one. I've been thinking how to use eg. PCF files.

litghost:
> As long as we have harnesses I can prepare such a JSON file for each one. But once we do the ROI breakout each design would need a separate one. I've been thinking how to use eg. PCF files.

Something like that is going to be required.  We could consider using sdc files.

mithro:
> > As long as we have harnesses I can prepare such a JSON file for each one. But once we do the ROI breakout each design would need a separate one. I've been thinking how to use eg. PCF files.
> 
> Something like that is going to be required. We could consider using sdc files.

I think we should use SDC files too.

```tcl
set_property DRIVE 16 [get_ports PIN_A];  #set drive to 16mA
set_property IOSTANDARD LVCMOS33 [get_ports PIN_A]
set_property PACKAGE_PIN N26 [get_ports PIN_A]
```

Interestingly this can also be set via the attribute like so;
```verilog
//----
// Sets the drive strength on the STATUS output port to 2 mA
(* DRIVE = "2" *) output STATUS,
// Sets the I/O Standard on the STATUS output to LVCMOS12
(* IOSATNDARD = "LVCMOS12" *) output STATUS,
//----
// Designates port CLK to be placed on pin B26
(* PACKAGE_PIN = "B26" *) input CLK;
```

```verilog
 // OBUF: Single-ended Output Buffer
 // Virtex-7
 // Xilinx HDL Language Template, version 2012.2

 OBUF #(
 .DRIVE(12), // Specify the output drive strength
 .IOSTANDARD("LVCMOS12"), // Specify the output I/O standard
 .SLEW("SLOW") // Specify the output slew rate
 ) status_obuf (
 .O(STATUS), // Buffer output (connect directly to top-level port)
 .I(status_int) // Buffer input
 );
 // End of status_obuf instantiation
```

mkurc-ant:
@mithro IOSTANDARD and DRIVE can be specified in the verilog file. That is not a problem. What I meant is that it would be good to have eg. an SDC file used both with VPR and fasm2bels. So in the end we would have a single file that defines all IO stuff for each design.

mkurc-ant:
I spotted a problem: Disassembly of eg. `ff_ce_sr_9_fdce` design shows that the IOB `LIOB33_X0Y43.IOB_Y0` is configured as an input. But according to the harness design this site is not used at all. The IOB_Y1 site is a LED output.

```
LIOB33_X0Y43.IOB_Y0.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL.SLEW.FAST
LIOB33_X0Y43.IOB_Y0.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL_SSTL135.IN_ONLY
LIOB33_X0Y43.IOB_Y0.LVCMOS25_LVCMOS33_LVTTL.IN
LIOB33_X0Y43.IOB_Y0.PULLTYPE.PULLUP
LIOB33_X0Y43.IOB_Y1.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL_SSTL135.SLEW.SLOW
LIOB33_X0Y43.IOB_Y1.LVCMOS33_LVTTL.DRIVE.I12_I16
LIOB33_X0Y43.IOB_Y1.PULLTYPE.NONE
```

This causes issues with equivalence test since the IOB configuration translates to a top level input `LIOB33_X0Y43_IOB_X0Y44_IPAD` which is not present in the original design.

I guess that a check has to be added to fasm2bels to check whether all IOBs are actually connected to something. Unconnected ones should be pruned.

xc7/fasm2bels/ioi_models.py
mkurc-ant: Done.

xc7/fasm2bels/iob_models.py
mkurc-ant: Done.

mkurc-ant:
So I hard coded an exception for site `IOB_X0Y44` which is ignored for now.

xc7/fasm2bels/iob_models.py
```diff
@@ -61,62 +61,96 @@ def get_iob_site(db, grid, tile, site):
...
 
 
-def add_output_parameters(bel, site):
-    """ Adds parameters required for output IOB BELs. """
-    assert 'IOSTANDARD' in bel.parameters
+def append_obuf_iostandard_params(
+        top, site, bel, possible_iostandards, slew="SLOW"
+):
+    """
+    Appends IOSTANDARD, DRIVE and SLEW parameters to the bel. Those parameters
```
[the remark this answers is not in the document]
mkurc-ant: The SLEW is being decoded from the fasm

xc7/fasm2bels/fasm2bels.py
```diff
@@ -271,13 +272,18 @@ def main():
...
+        with open(args.iostandard_defs) as fp:
+            defs = json.load(fp)
+            top.set_iostandard_defs(defs)
 
     for tile, tile_features in tiles.items():
         process_tile(top, tile, tile_features)
 
     top.make_routes(allow_orphan_sinks=args.allow_orphan_sinks)
 
+    top.prune_unconnected_ports()
```
litghost: Can you make this optional and driven by a flag?

xc7/fasm2bels/iob_models.py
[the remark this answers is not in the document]
litghost: Can you clarify the comment then?  The first sentence feels correct, but I interpreted the second sentence to mean that all 3 parameters had to specified in the iostandard_defs dict.

xc7/fasm2bels/ioi_models.py
```diff
@@ -0,0 +1,103 @@
...
+
+def process_ilogic(top, features):
+
+    aparts = features[0].feature.split('.')
+    # tile_name = aparts[0]
+    ioi_site = get_ioi_site(top.db, top.grid, aparts[0], aparts[1])
+
+    site = Site(features, ioi_site)
+
+    # TODO: Support IDDR, ISERDES etc
```
litghost: Same story as IDELAY, please add assert that ISERDES/IDDR IN_USE features are not set.  This applies for the OLOGIC block too.

xc7/fasm2bels/fasm2bels.py
mkurc-ant: Done.

xc7/fasm2bels/iob_models.py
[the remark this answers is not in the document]
mkurc-ant: Right, I clarified the comment. Only IOSTANDARD and DRIVE need to be specified. The SLEW is decoded from fasm.

xc7/fasm2bels/ioi_models.py
mkurc-ant: I added asserts for ISERDES, IDELAY and OSERDES.

PR #939, opened 2019-08-13
Add fast path timing for better hold slack estimation.

litghost:


litghost:
@kgugala I would like your review as I haven't touched this code before.

PR #945, opened 2019-08-14
WIP: Incorporate connection box updated segments

litghost:


litghost:
CI is green, but QoR isn't better and runtime is significantly worse.

litghost:
> CI is green, but QoR isn't better and runtime is significantly worse.

QoR is actually not worse, Fmax and CPD columns were swapped.  Some runtime improvements are coming.  QoR expirements are showing that the new lookahead is helping once abc9 is turned on.  Plan is to land new yosys (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/948/commits/004ecc6ce12bf274f33723b9b9765b59ee4f49ca), then enable abc9, then compare QoR or yosys + abc9 vs yosys + abc9 + new lookahead.  Local experiments show CPD reduction from 16 ns to 11 ns on murax, and improvement on picosoc, but those experiments are uncontrolled.

litghost:
With connection block lookahead, picosoc and murax both close timing at 50 MHz, @mithro.  I need to fix #950, and then clean up this PR and land it.

PR #948, opened 2019-08-19
Update yosys.

litghost:


litghost:
@mithro I'm not sure what the TODO was for, but it seems safe to remove it.  Yosys no longer has the ice40_unlut pass, so we need to change something.

PR #949, opened 2019-08-19
Lower poll interval when redirecting to non-interactive terminal (e.g. CI).

litghost:


mithro:
`fill` should be `file` in the commit message?

litghost:
> `fill` should be `file` in the commit message?

Ya, I noticed the typo at PR time.

mithro:
LGTM.

PR #950, opened 2019-08-19
Enable Yosys abc9 pass.

litghost:
This dramatically improves the synthesis result from a timing
perspective.

PR #952, opened 2019-08-20
Incorporate connection box updated segments astar

litghost:
Variant on #945 with different parameters to see if runtime can be brought under control.


mithro:
This looks like it could use a rebase?

litghost:
> Should this be it's own pull request?

It is?  #950

PR #954, opened 2019-08-21
[XC7] Attempt to enable SRL synthesis.

litghost:
This may result in an SRL16, which will cause a failure.


litghost:
Confirm, yosys is trying to infer SRL16's, need to make a version that only uses SRL32.

PR #957, opened 2019-08-21
Add output of iostandard_defs.json from PCF format.

litghost:
- Adds picosoc vivado target
- Fixes bug in BRAM model for write widths other than 18.


mithro:
LGTM!

PR #961, opened 2019-08-23
SRL16 support

mkurc-ant:
Add support for SRL16 to VPR and fasm2bels with timings.


xc7/fasm2bels/clb_models.py
```diff
@@ -209,57 +250,101 @@ def cleanup_slice(top, site):
 
     """
     carry4 = site.maybe_get_bel('CARRY4')
+    if carry4 is not None:
 
-    if carry4 is None:
-        return
+        # Simplest check is if the CARRY4 has output in used by either the OUTMUX
```
litghost: Let's refactor the cleanup into "cleanup_carry4", "cleanup_srl", assuming that no logic/data is shared between them.

xc7/primitives/slicem/slicem.pb_type.xml
```diff
@@ -256,9 +256,8 @@
         <direct name="D_SRL.O5_to_SLICEM_MODES.DO5" input="DSRL.O5" output="SLICEM_MODES.DO5"/>
         <direct name="D_SRL.O6_to_SLICEM_MODES.DO6" input="DSRL.O6" output="SLICEM_MODES.DO6"/>
 
-        <!-- WE/CE inputs -->
+        <!-- CE input. Don't connect WE to force routing through WE -->
         <direct name="SLICEM_MODES.CE_to_WE_MUX.CE" input="SLICEM_MODES.CE" output="WE_MUX.CE"/>
-        <direct name="SLICEM_MODES.WE_to_WE_MUX.WE" input="SLICEM_MODES.WE" output="WE_MUX.WE"/>
```
litghost: Why force `CE` over `WE` here?

xc7/primitives/slicem/srl/a_srl.pb_type.xml
```diff
@@ -0,0 +1,65 @@
+<pb_type name="ASRL" num_pb="1" xmlns:xi="http://www.w3.org/2001/XInclude">
```
litghost: Please add a comment on why the a_srl is special, for future reference.

xc7/techmap/cells_map.v
```diff
@@ -1756,3 +1756,75 @@ module SRLC32E (
...
+module SRL16E (
+  output Q,
+  input A0, A1, A2, A3,
+  input CE, CLK, D
+);
+  parameter [15:0] INIT = 16'h0000;
+  parameter [ 0:0] IS_CLK_INVERTED = 1'b0;
+
+  // Substitute
+  SRLC16E #
```
litghost: Can you document in a comment/readme what is going on here?

xc7/primitives/slicem/slicem.pb_type.xml
mkurc-ant: This is to maintain compatibility with Vivado which cannot route through `WE` to an SRL.

There is no such limitation in hardware, both `CE` and `WE` work. However, when trying to import disassembled fasm with routing through `WE`, Vivado refuses to use that input which results in "antenna net" error.

I added more elaborate comment to the file.

xc7/primitives/slicem/srl/a_srl.pb_type.xml
mkurc-ant: It is special because it requires different pack patterns for chaining. I added comment at the beginning of the file.

xc7/techmap/cells_map.v
mkurc-ant: I added a comment to the SRL section of the file.

xc7/fasm2bels/clb_models.py
mkurc-ant: Done.

xc7/primitives/slicem/slicem.pb_type.xml
litghost: You have misinterpreted the error.  Vivado can route using the `WE` to the SRL, but will generally refuse if the `CE` port is available.  The example to check is have the SRL CE and FF CE use different net's, but force them to pack into the same `SLICEM`.

mkurc-ant: Ok, I'll check that tomorrow. But still, if I enable `WE` there will be issues with importing the VPR routing back to Vivado in case when there is only SRL in a slice.

litghost: The solution to this problem is have VPR choose CE over WE, either by making sure it is the default (e.g. first in the `<mux>`) or by some timing pressure.  Presumably Vivado chooses CE over WE for that reason.

mkurc-ant: I've re-enabled the WE input of WEMUX.

Actually, the `CE` input should be the default as it is defined in the XML:
```
<mux name="WE_MUX" input="WE_MUX.CE WE_MUX.WE" output="WE_MUX.WE_OUT">
  <metadata>
    <meta name="fasm_mux">
      WE_MUX.CE = WEMUX.CE
      WE_MUX.WE = NULL
    </meta>
  </metadata>
</mux>
```

xc7/tests/srl/srl16_amc31/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+add_file_target(FILE basys3_doutmux_top.v SCANNER_TYPE verilog)
+#add_file_target(FILE basys3_dffmux_top.v SCANNER_TYPE verilog)
```
litghost: Why are the dffmux tests commented out?

mkurc-ant: That is because I cannot verify it on hardware. The VPR refuses to use FF on the same slice as the SRL. I'll try to play more with pack patterns, maybe I can get it to do so.

mkurc-ant: @litghost I discovered that in order to have such a pack pattern we would need to make different COMMON_SLICE and SLICE_FF pb_types for SLICEL and SLICEM. That is because MC31 related pack patterns cannot be specified inside SLICEL as this makes VPR report an error.

I made specialized COMMON_SLICE and SLICE_FF but still couldn't force VPR to pack SRL and FF in the same slice.

I think that having the DFFMUX.MC31 working is not a top priority. I can file an issue with a description what (supposedly) needs to be done.

litghost: > I think that having the DFFMUX.MC31 working is not a top priority. I can file an issue with a description what (supposedly) needs to be done.

This makes sense to me for now.

mkurc-ant: I added comments to CMake files. There is the issue: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/974

PR #964, opened 2019-08-28
Refactored timing import with explicit CELL specification and order.

mkurc-ant:
This PR introduces a change to the `bels.json` format thus allowing to explicitly specify which SDF CELLs and in what order to import.

SDF CELLs are now specified as a list of CELLTYPE.INSTANCE, not as independent list of CELLTYPEs and INSTANCEs. Timings from multiple CELLs are merged as it was done before, but now in the order from last to first. This means that if there are multiple CELLs which specify the same timing, the former ones will take precedence over the latter ones.

For example, in an SDF we have:

```
(CELL
    (CELLTYPE "LUT_OR_MEM6LRAM")
    (INSTANCE SLICEM/A6LUT)
    (DELAY
        (ABSOLUTE
            (IOPATH A1 O6 (0.045::0.056)(0.100::0.124))
            (IOPATH A2 O6 (0.045::0.056)(0.100::0.124))
            (IOPATH A3 O6 (0.045::0.056)(0.100::0.124))
            (IOPATH A4 O6 (0.045::0.056)(0.100::0.124))
            (IOPATH A5 O6 (0.045::0.056)(0.100::0.124))
            (IOPATH A6 O6 (0.045::0.056)(0.100::0.124))
            (IOPATH CLK O6 (0.432::0.538)(0.930::1.153))
        )
    )
    (TIMINGCHECK
        (HOLD DI1 (posedge CLK) (0.155::0.192))
        (SETUP DI1 (posedge CLK) (0.366::0.453))
        (HOLD DI2 (posedge CLK) (0.098::0.122))
        (SETUP DI2 (posedge CLK) (0.309::0.384))
    )
)

(CELL
    (CELLTYPE "LUT_OR_MEM6SHFREG")
    (INSTANCE SLICEM/A6LUT)
    (DELAY
        (ABSOLUTE
            (IOPATH CLK MC31 (0.332::0.414)(0.898::1.114))
            (IOPATH CLK O6 (0.444::0.553)(1.186::1.472))
        )
    )
    (TIMINGCHECK
        (HOLD DI1 (posedge CLK) (0.075::0.093))
        (SETUP DI1 (posedge CLK) (0.137::0.170))
        (HOLD DI2 (posedge CLK) (0.076::0.094))
        (SETUP DI2 (posedge CLK) (0.140::0.173))
    )
)
```

We want CLK, DI1 and DI2 timings from LUT_OR_MEM6SHFREG.SLICEM/A6LUT but A1-A6 timings from LUT_OR_MEM6LRAM.SLICEM/A6LUT.

We define in `bels.json`:

```
"SLICEM": {
    "SRLC32E_VPR":{
        "ASRL":
            [ "LUT_OR_MEM6SHFREG.SLICEM/A6LUT", "LUT_OR_MEM6LRAM.SLICEM/A6LUT"]
    }
}
```

And the timings from the LUT_OR_MEM6SHFREG.SLICEM/A6LUT cell will be overlaid over the timings from the LUT_OR_MEM6LRAM.SLICEM/A6LUT cell.

Moreover, CELL names can be specified using a wildcard pattern. This reduces the complexity of the `bels.json` file. For example, instead of writing:

```
ROUTING_BEL.SLICEL/AMUX
ROUTING_BEL.SLICEL/BMUX
ROUTING_BEL.SLICEL/CMUX
ROUTING_BEL.SLICEL/DMUX
```

we can write

```
ROUTING_BEL.SLICEL/[ABCD]MUX
```

and it will be automatically expanded. Only one wildcard pattern in `[]` braces can be specified.

### DIFF
--- a/utils/update_arch_timings.py
+++ b/utils/update_arch_timings.py
@@ -32,37 +32,74 @@ def remove_site_number(site):
     return site
 
 
-def get_cell_types_and_instance(bel, location, site, bels):
+def gen_all_possibilities(pattern):
+    """
+    Generates all possible combinations of a pattern if it contains a
+    wildcard string in braces eg. "LUT[ABCD]" will yield in "LUTA", "LUTB"
+    and so on.
+
+    >>> list(gen_all_possibilities("LUT"))
+    ['LUT']
+
+    >>> list(gen_all_possibilities("LUT[ABCD]"))
+    ['LUTA', 'LUTB', 'LUTC', 'LUTD']
+    """
+
+    # Match the regex
+    match = re.match("(.*)\[([A-Za-z0-9]+)\](.*)", pattern)
+
+    # Generate combinations
+    if match is not None:
+        for c in match.group(2):
+            yield match.group(1) + c + match.group(3)
+
+    # Not a regex
+    else:
+        yield pattern
+
+
+def get_cell_types_and_instances(bel, location, site, bels):
     """This function searches for a bel type and instance
        translation between VPR and Vivado. The translation
        is defined in the `bels` dictionary. If translation
-       is found celltypes list and bel instance is returned,
-       `None` otherwise"""
+       is found a list of celltypes and bel instances is returned,
+       None otherwise"""
     if site not in bels:
-        return None, None
+        return None
     if bel not in bels[site]:
-        return None, None
+        return None
     if location not in bels[site][bel]:
-        return None, None
+        return None
 
-    celltypes = (bels[site][bel][location]['celltype']).split()
-    instance = bels[site][bel][location]['instance']
+    # Generate a list of tuples (celltype, instance)
+    cells = []
+    for pattern in bels[site][bel][location]:
+        for names in gen_all_possibilities(pattern):
+            cells.append(tuple(names.split(".")))
 
-    return celltypes, instance
+    return cells
 
 
 def find_timings(timings, bel, location, site, bels, corner, speed_type):
     """This function returns all the timings associated with
        the selected `bel` in `location` and `site`. If timings
-       are not found, `None` is returned"""
-    celltype, instance = get_cell_types_and_instance(bel, location, site, bels)
-    if (celltype is None) or (instance is None):
+       are not found, empty dict is returned"""
+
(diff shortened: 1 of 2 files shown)


utils/update_arch_timings.py
```diff
@@ -32,37 +32,74 @@ def remove_site_number(site):
...
+    match = re.match(r"(.*)\[([A-Za-z0-9]+)\](.*)", pattern)
+
+    # Generate combinations
+    if match is not None:
+        for c in match.group(2):
+            yield match.group(1) + c + match.group(3)
+
+    # Not a regex
+    else:
+        yield pattern
```
kgugala: shouldn't this be `return`? This is not a partial result

utils/update_arch_timings.py
```diff
@@ -32,37 +32,74 @@ def remove_site_number(site):
...
 def find_timings(timings, bel, location, site, bels, corner, speed_type):
     """This function returns all the timings associated with
        the selected `bel` in `location` and `site`. If timings
-       are not found, `None` is returned"""
-    celltype, instance = get_cell_types_and_instance(bel, location, site, bels)
-    if (celltype is None) or (instance is None):
+       are not found, empty dict is returned"""
+
+    # Get cells, reverse the list so former timings will be overwritten by
+    # latter onex
```
kgugala: onex?

utils/update_arch_timings.py
mkurc-ant: No, that is correct. If the pattern does not contain a wildcard then it is considered to be a single name of a CELLTYPE.INSTANCE and is returned directly.

utils/update_arch_timings.py
mkurc-ant: A typo... corrected.

utils/update_arch_timings.py
kgugala: this is exactly what I'm saying. If we return the value directly it should be done with `return`. `yield` is confusing here.

mkurc-ant: But this function is used as a generator, it cannot return anything. It must yield even if it is a single item.

kgugala: ok

PR #982, opened 2019-09-12
Fixes some issues with 64x1D test.

litghost:
- Trailing commas are not standard, remove them.
 - error checker is assuming registered RAM, add register to LUT-RAM.


mithro:
CI failure?

litghost:
> CI failure?

We've been having unrelated intermittent CI failures due to what appears to be VM disk failures.  The failure in http://sponge.corp.google.com/acee5057-e70f-4847-b9b9-e5cf8860a789 is hard to explain as anything other than the GCS VM disk failing to behave correctly.  I've been attempting replication of the failures without success on my local machine.  I've started trying to replicate on the large symbiflow VM.

PR #986, opened 2019-09-12
Enable connection box lookahead

litghost:


litghost:
> generation of `rr_graph_xc7a50t-basys3_test.lookahead.bin` and `rr_graph_xc7a50t-basys3_test.place_delay.bin` takes long time.

I have some plans to try and improve this runtime.  There is some oversampling that is occuring with the router lookahead that may be wasted time, and hopeful I can cut the lookahead build from 900 seconds to closer to 300 seconds.  But for now, I'd like to commit something so Vaughn's team of students don't have to work on a branch of symbiflow-arch-defs.

PR #987, opened 2019-09-12
Add utilities for print override delay matrix and connection box lookahead

litghost:


litghost:
> Could it be worth adding CMake targets to run these utilities automatically and output debug files?
> 
> CI failed with the following Icarus error (never seen before):

I've been tracking the failure, it looks like a depedency or file collision failure.  Unrelated to this PR.

litghost:
> CI failed with the following Icarus error (never seen before):

I've been able to replicate this on the large symbiflow VM.  I believe this may be related to the VM change (med to large) we tried this week.

litghost:
> Could it be worth adding CMake targets to run these utilities automatically and output debug files?

The capnp files are located in the VTR repo, which is not currently imported.   I'm hesistent to do that right now just for this tool.

mithro:
Should we update the conda package to include the capnp files?

litghost:
> Should we update the conda package to include the capnp files?

That will be required when/if we write the routing graph directly as capnp.  So yes?

PR #990, opened 2019-09-13
Avoid cache rebuild

litghost:
Optimization on top of #980


litghost:
> Is this tool to deal with the fact that the inputs might change timestamp but not contents?

Yes.  This will be very common as we change pb_types, but not the routing graph.

mithro:
LGTM.

PR #991, opened 2019-09-13
Use ninja build system.

litghost:
Unclear why, but using ninja might help with the flakyness we've been seeing on the large VMs.  It may also speed up builds too.  As a side affect, ninja logs can be used to easily profiler the build times (see https://buildbloat.neocities.org/)

### DIFF
--- a/.github/kokoro/continuous-docs.cfg
+++ b/.github/kokoro/continuous-docs.cfg
@@ -13,6 +13,7 @@ action {
     # File types
     regex: "**/*result*.xml"
     regex: "**/*sponge_log.xml"
+    regex: "**/.ninja_log"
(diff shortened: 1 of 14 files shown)


litghost:
I'm going to re-run kokoro a couple times to see if it is more stable, or just a coincidence.  Ninja does appear to run faster than the Makefiles, so this is probably worth applying to the other jobs too.

litghost:
> Generally LGTM.
> 
> Is Travis is still running with make?

Right now only kokoro xc7 is running with ninja, everything else is Make.

mithro:
(BTW I do think we want to have something continuing to test with just make.)

PR #997, opened 2019-09-16
Yosys: cells property should not return internal cells

kgugala:
Comment says that the cell property does not include Yosys internal cells, but the code is not filtering them.

### DIFF
--- a/utils/vlog/yosys/json.py
+++ b/utils/vlog/yosys/json.py
@@ -51,6 +51,8 @@ class YosysModule:
         """
         clist = []
         for cell, cdata in sorted(self.data["cells"].items()):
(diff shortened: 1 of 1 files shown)


mithro:
Are we actually sure we want to exclude Yosys internal cells?

kgugala:
> Are we actually sure we want to exclude Yosys internal cells?

in this property (`cells`) yes. It's docstring says:
```
List of cells of a module, excluding Yosys-internal cells beginning with $.
```

If you need all the cells you should use `all_cells` described as:
```
List of cells of a module, including Yosis-internal cells
```

Before the fix both `cells` and `all_cells` were doing the same thing

PR #999, opened 2019-09-16
Add Vivado P&R targets.

litghost:
This enables direct post-synthesis Vivado comparisions with VPR easily.

### DIFF
--- a/tests/9-soc/murax/CMakeLists.txt
+++ b/tests/9-soc/murax/CMakeLists.txt
@@ -39,5 +39,13 @@ add_vivado_target(
     CLOCK_PINS io_mainClk
     CLOCK_PERIODS 10.0
     )
+
+add_vivado_pnr_target(
+    NAME murax_basys_vivado_pnr
+    PARENT_NAME murax_basys
+    CLOCK_PINS io_mainClk
+    CLOCK_PERIODS 10.0
+    IOSTANDARD LVCMOS33
+    )
 endif (NOT DEFINED ENV{CI} OR NOT $ENV{CI})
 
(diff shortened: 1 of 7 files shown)


tests/9-soc/murax/CMakeLists.txt
```diff
@@ -39,5 +39,13 @@ add_vivado_target(
     CLOCK_PINS io_mainClk
     CLOCK_PERIODS 10.0
     )
+
+add_vivado_pnr_target(
+    NAME murax_basys_vivado_pnr
+    PARENT_NAME murax_basys
+    CLOCK_PINS io_mainClk
```
mithro: Can we just use an SDC / XDC file instead of putting this in the build rule?

xc7/make/vivado.cmake
```diff
@@ -1,3 +1,98 @@
...
+    COMMAND ${CMAKE_COMMAND} -E remove -f ${WORK_DIR}/design_${NAME}.xpr
+    COMMAND ${PRJXRAY_DIR}/utils/vivado.sh -mode batch -source
+        ${CMAKE_CURRENT_BINARY_DIR}/${NAME}_runme.tcl
+        > ${CMAKE_CURRENT_BINARY_DIR}/${WORK_DIR}/vivado.stdout.log
+    WORKING_DIRECTORY ${WORK_DIR}
+    DEPENDS ${DEPS} ${NAME}_runme.tcl
+    )
+
+  add_custom_target(
+      ${NAME}_load_dcp
```
mithro: `load`? Maybe `show` or `view` would be a better name?

tests/9-soc/murax/CMakeLists.txt
litghost: That is an option, but most of the existing symbiflow tests just use a PCF right now.

mithro: I logged https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1000

xc7/make/vivado.cmake
mithro: FYI - I also logged https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1001

Issue #902, opened 2019-07-30
[xc7] Support RAMB36E1


litghost:
Tasks:
 - [ ] Ensure that RAMB36E1 can be placed and routed
 - [ ] Add simulation model for RAMB18E1 and RAMB36E1
 - [ ] Create test benches to verify pre- and post-place and routed designs using RAMB18E1 and RAMB36E1
 - [ ] Verify RAMB18E1 and RAMB36E1 operation in hardware

Support not required for:
 - RAMB36E1 1-bit cascades
 - ECC support

litghost:
Existing BRAM36 tests appear to work on hardware.  I'm working on fasm2v support next, followed by RAMB sim models so we can add testbenches.

litghost:
Clarification on BRAM36 support, I suspect that only the widest modes are working (e.g. 36 and 72 bit wide), for the same reason that https://github.com/SymbiFlow/symbiflow-arch-defs/commit/9c65a49cd5c6b88d7606a8cb817dbee448050013#diff-2bb469162bf2c7f8904db77a1acc2582 was required.  Once fasm2v support is in, I'll add the following tests:
- [ ] RAMB18 width = 2
- [ ] RAMB18 width = 4
- [ ] RAMB18 width = 9
- [ ] RAMB36 width = 2
- [ ] RAMB36 width = 4
- [ ] RAMB36 width = 9
- [ ] RAMB36 width = 18

PR #1007, opened 2019-09-17
Correct RAMB36E1 write enables and add fasm2v support.

### LINKED ISSUE #905: [xc7/fasm2v] Support RAMB36E1
fasm2v needs support for RAMB36E1. Support already exists for RAMB18E1, and the only difference is in how the upper address lines are routed.  Need to distinguish between RAMB18E1 and RAMB36E1 based on how the upper address lines are routed.

litghost:
Fixes #905

Part of #902

### DIFF
--- a/xc7/fasm2bels/make_routes.py
+++ b/xc7/fasm2bels/make_routes.py
@@ -154,7 +154,10 @@ class Net(object):
         self.parent_nodes[node_pkey] = parent_node_pkey
 
         for wire_pkey in get_wires_in_node(conn, node_pkey):
-            net_map[wire_pkey] = self.source_wire_pkey
+            if wire_pkey not in net_map:
+                net_map[wire_pkey] = set()
+
+            net_map[wire_pkey].add(self.source_wire_pkey)
             self.route_wire_pkeys.add(wire_pkey)
 
     def expand_source(self, conn, check_downstream_default, net_map):
@@ -453,15 +456,16 @@ def expand_sink(
                         )
                     )
 
-                nets[net_map[upstream_sink_wire_pkey]].add_node(
-                    conn=conn,
-                    net_map=net_map,
-                    node_pkey=sink_node_pkey,
-                    parent_node_pkey=get_node_pkey(
-                        conn, upstream_sink_wire_pkey
-                    ),
-                    incoming_wire_pkey=node_wire_pkey
-                )
+                for net in net_map[upstream_sink_wire_pkey]:
+                    nets[net].add_node(
+                        conn=conn,
+                        net_map=net_map,
+                        node_pkey=sink_node_pkey,
+                        parent_node_pkey=get_node_pkey(
+                            conn, upstream_sink_wire_pkey
+                        ),
+                        incoming_wire_pkey=node_wire_pkey
+                    )
                 return
 
     # There are no active pips upstream from this node, check if this is a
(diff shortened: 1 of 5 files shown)


litghost:
Testing on hardware:
 - [ ] scalable proc using RAMB18 (and fasm2v)
 - [ ] scalable proc using RAMB36 (and fasm2v)
 - [ ] bram_test_18
 - [ ] bram_test_36
 - [ ] bram_init_test_18
 - [ ] bram_init_test_36
 - [ ] murax (RAMB18)

Tested fasm2v:
 - [ ] picosoc (RAMB36)

xc7/fasm2bels/make_routes.py
```diff
@@ -154,7 +154,10 @@ def add_node(
         self.parent_nodes[node_pkey] = parent_node_pkey
 
         for wire_pkey in get_wires_in_node(conn, node_pkey):
-            net_map[wire_pkey] = self.source_wire_pkey
+            if wire_pkey not in net_map:
+                net_map[wire_pkey] = set()
```
HackerFoo: I see this pattern a lot. What do you think about using this class for dictionaries of sets?

```python
class SetDict(dict):
    def __missing__(self, key):
        return set()
```

xc7/techmap/cells_map.v
```diff
@@ -1487,6 +1487,59 @@ end
       localparam ZINV_REGCLKB = 1'b0;
   end
 
+  wire [7:0] WEBWE_WIDE;
+  wire [3:0] WEA_WIDE;
+
+  if(WRITE_WIDTH_A < 18) begin
+      assign WEA_WIDE[3] = WEA[0];
```
HackerFoo:
```verilog
assign WEA_WIDE = {4{WEA[0]}};
```

xc7/techmap/cells_map.v
```diff
@@ -1487,6 +1487,59 @@ end
...
+      assign WEA_WIDE[2] = WEA[0];
+      assign WEA_WIDE[1] = WEA[0];
+      assign WEA_WIDE[0] = WEA[0];
+  end else if(WRITE_WIDTH_A == 18) begin
+      assign WEA_WIDE[3] = WEA[1];
```
HackerFoo:
```verilog
assign WEA_WIDE = {2{WEA[1]}, 2{WEA[0]}};
```

xc7/techmap/cells_map.v
```diff
@@ -1487,6 +1487,59 @@ end
...
+      assign WEA_WIDE[2] = WEA[1];
+      assign WEA_WIDE[1] = WEA[0];
+      assign WEA_WIDE[0] = WEA[0];
+  end else if(WRITE_WIDTH_A == 36) begin
+      assign WEA_WIDE[3] = WEA[3];
```
HackerFoo:
```verilog
assign WEA_WIDE = WEA;
```

xc7/fasm2bels/make_routes.py
litghost: I'm not a fan of this kind of magic, because it masks invalid keys down the road, and there isn't a strong readability improvement at this site because you still need to know that net_map is a `SetDict`, rather than just `dict`.

Also this strategy doesn't work because the return is to an object not in the dictionary itself.

```
class SetDict(dict):
    def __missing__(self, key):
        o = set()
        self[key] = o
        return o
```

would work, but again, this is bad pattern IMO.

mithro: FYI Python also has the `defaultdict` object...
```python
from collections import defaultdict
mydict = defaultdict(set)
```

mithro: BTW I'm not a huge fan of defaultdict and agree with @litghost's comment that it masks invalid keys.

PR #1009, opened 2019-09-17
Update to latest prjxray-db (including the new basys 50 MHz harness).

litghost:


mithro:
If CI passes, go ahead and merge.

PR #1015, opened 2019-09-19
Add IO architecture primitives

HackerFoo:


xc7/primitives/idelaye2/idelaye2.pb_type.xml
```diff
@@ -0,0 +1,40 @@
+<!-- set: ai sw=1 ts=1 sta et -->
+<pb_type name="IDELAYE2" blif_model=".subckt IDELAYE2" num_pb="1">
```
mkurc-ant: The `blif_model` here should go away. VPR can't route through leaf pb_types. Instead we should have IDELAYE2 with two modes "passtrhough" and "idelay". The "passthrough" would be as it is now and the "idelay" would contain the leaf pb_type with the `blif_model` set to (probably) sth. like `IDELAYE2_VPR`.

litghost: Reminder that this is the "IDELAY" site, which is not actually blocking any routing, and doesn't support a route through mode.  THE ILOGIC and OLOGIC sites both contain passthrough modes.  So I don't believe any change is required here.

mkurc-ant: Right... I was thinking about ILOGIC

PR #1026, opened 2019-09-24
[xc7] Fixes for SRL and LUT-RAM

litghost:
These fixes were identified via fasm2v differences between VPR and Vivado or errors during Vivado.

### DIFF
--- a/tests/9-scalable_proc/utils/rom_generator.py
+++ b/tests/9-scalable_proc/utils/rom_generator.py
@@ -115,7 +115,7 @@ RAM64X1D #
 )
 dram_{row}_{col}
 (
-.WCLK   (1'b0),
+.WCLK   (CLK),
 .WE     (1'b0), .D(1'b0),
 .DPRA0  (1'b0), .DPRA1(1'b0), .DPRA2(1'b0),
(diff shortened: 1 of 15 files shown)


litghost:
I've identified a bug with the RAM32M/RAM64M fasm2v implementation, I'm working on testing a fix.

litghost:
Fix implemented for DI1 mux.

litghost:
> LGTM. Can you just post here a short info what problems this PR does fix?

- Update scaleable proc template to route clock to LUT-RAM.  This is required because of what Vivado does with clocks set to 0.
- Implemented the full DI1 mux for LUT-RAMs.
- Implemented RAM32M and RAM64M in fasm2m
- Added RAM32X1D test
- Fixed RAM32M and RAM64M LUT-RAM tests (they were failing)
- Fixed 32-bit LUT-RAM initialization
- Added pack pattern for SRL x2, x3 and x4 chains.  This is required because of how Vivado implements SRL32 packing.
- Fixed bug in srl32_chain where the tail of the SRL chain was unused and being discarded because the SRL address was set to 1 (rather than len-1).

PR #1027, opened 2019-09-24
Create a target for checking all diff FASM ("all_xc7_diff_fasm").

litghost:


xc7/make/vivado.cmake
```diff
@@ -82,15 +121,18 @@ function(COMMON_VIVADO_TARGETS NAME WORK_DIR DEPS BITSTREAM)
...
-      COMMAND diff -u
-        ${BITSTREAM_LOCATION}.fasm
-        ${CMAKE_CURRENT_BINARY_DIR}/${WORK_DIR}/design_${NAME}.bit.fasm
-      DEPENDS
-        ${DEPS}
-        ${BITSTREAM_LOCATION}.fasm
-        ${CMAKE_CURRENT_BINARY_DIR}/${WORK_DIR}/design_${NAME}.bit.fasm
-      )
+
+  if(${COMMON_VIVADO_TARGETS_MAKE_DIFF_FASM})
```
mithro: Do you need to canonicalize the file first?

litghost: No.  If bit2fasm is given an identical input bitstream, it will generate identical FASM.  We only care if the bitstream is exactly the same.

mithro: Okay great!

PR #1017, opened 2019-09-20
updated symbiflow archs to integrate the newly added tiles tag in VTR

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

This PR modifies the architectures to accommodate the new `<tiles>` tag addition to VPR.

- XC7: the import files have been changed accordingly
- Ice40, testarch: the architectures go through a final step and are modified with the `update_arch_tiles.py` script.

CI will fail until the new VTR version is added to conda.

PR #1029, opened 2019-09-25
Add FASM features for DI1MUX's to ensure that ADI1 connection is valid.

litghost:
An error will result if the ADI1 connection does not match the BDI1
connection.

Does requires `genfasm` fix, tracked https://github.com/SymbiFlow/vtr-verilog-to-routing/issues/280 .  Plan is to update VTR with this change after https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1017 is merged, (FYI @acomodi)


litghost:
Merged in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1036

PR #1035, opened 2019-09-27
Handle site pins that connect to pips not in the same tile.

litghost:
The vast majority of pips that connect to site pins connect within the
same tile, however this is not true of some of the pips in the IOB/IOI
tiles.

### DIFF
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -219,6 +219,18 @@ CREATE TABLE pip_in_tile(
   FOREIGN KEY(backward_switch_pkey) REFERENCES switch(pkey)
 );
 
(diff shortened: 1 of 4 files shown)


litghost:
@acomodi @HackerFoo With this PR, prjxray_create_edges runs to completion without the ROI present.

litghost:
The QoR results show an unexpected picosoc change (increase in route time, bad CPD), investigating now.

litghost:
Given that the picosoc result matches closely with the outcome of the #1036 which had unrelated changes, I suspect there is a sensitive path in the picosoc router, and depending on inputs, there is a chaotic switchover.  This may indicate that more routing effort is required to stabilize timing.

PR #1037, opened 2019-09-30
Refactor tracks to cache neighboor computation.

litghost:
During edge creation, get_tracks_for_wire_at_coord gets called enough
times that caching the result is a good memory <-> CPU trade off.

Part of prjxray_create_edge scaling work to reach full 50T in reasonable time.

### DIFF
--- a/utils/lib/rr_graph/tracks.py
+++ b/utils/lib/rr_graph/tracks.py
@@ -158,6 +158,8 @@ class Tracks(object):
         self.tracks = tracks
         self.track_connections = track_connections
 
+        self.track_cache = {}
+
     def verify_tracks(self):
         """ Verify that all tracks are connected to all other tracks. """
         track_connections = {}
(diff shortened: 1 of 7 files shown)


HackerFoo:
Is it worth using something like [cachetools](https://pypi.org/project/cachetools/) to remove all the ad-hoc caching code?

litghost:
> Is it worth using something like [cachetools](https://pypi.org/project/cachetools/) to remove all the ad-hoc caching code?

No?  Not only is it a new depedency, but these cache's do need TTL or LRU features.  The existing https://docs.python.org/3/library/functools.html#functools.lru_cache is fine if LRU is needed.

HackerFoo:
I couldn't find a way to make `lru_cache` ignore `self`. I'm not sure if that would cause a problem or not.

litghost:
> I couldn't find a way to make `lru_cache` ignore `self`. I'm not sure if that would cause a problem or not.

Correct, hence why the `lru_cache` we use are nested functions.  The outer scope serves as the "self".  However, in most cases the only thing required in the outer scope is a connection database reference, which we wouldn't want to cache over anyways.

HackerFoo:
Still, it would be nice to use some common annotation or wrapper, since we do a fair amount of caching/memoizing.

mithro:
FWIW - I'm fine with introducing a dependency on extra Python modules like the cachetools package that @HackerFoo linked too.

I'm also okay with @litghost using adhoc caching for now and then refactoring / cleaning this in the future to use a common package.

utils/lib/rr_graph/tracks.py
```diff
@@ -158,6 +158,8 @@ def __init__(self, tracks, track_connections):
         self.tracks = tracks
         self.track_connections = track_connections
 
+        self.track_cache = {}
```
mithro: Maybe?
```suggestion
        self._track_cache = {}
```

utils/lib/rr_graph/tracks.py
```diff
@@ -225,16 +227,37 @@ def is_wire_adjacent_to_track(self, idx, coord):
             assert False, track
 
     def get_tracks_for_wire_at_coord(self, coord):
-        """Returns which track indicies and direction a wire at a coord can
-        be connected too.
+        """Returns map of direction to track indicies for valid connections.
+
+        There may be multiple connections to tracks at a particular coordinate,
+        this method only returns a possible connection in each direction this
+        this track.
```
mithro: `in each direction this this track.`

utils/lib/rr_graph/tracks.py
```diff
@@ -225,16 +227,37 @@ def is_wire_adjacent_to_track(self, idx, coord):
...
+
+        Parameters
+        ----------
+        coord : (int, int)
```
mithro: FYI - We should use [typing annotation](https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html) rather than doing it in the docstring (where it works).

xc7/utils/prjxray_assign_tile_pin_direction.py
```diff
@@ -283,8 +283,9 @@ def handle_edges_to_channels(
 
                 tracks_model = channel_wires_to_tracks[track_pkey]
                 available_pins = set(
```
mithro: nit: Should this be a mutable or immutable set?

PR #1038, opened 2019-09-30
Refactor channel2 track packing algorithm.

litghost:
New algorithm is significantly faster, and avoids creating Interval
objects.

Part of prjxray_create_edge scaling work to reach full 50T in reasonable time.

### DIFF
--- a/utils/lib/rr_graph/channel2.py
+++ b/utils/lib/rr_graph/channel2.py
@@ -5,7 +5,6 @@ an excessive number of dummy tracks to fill empty space (channel.Channel = >2M
 versus channel2.Channel ~70k).
 
 """
-from intervaltree import IntervalTree, Interval
 
 
 class Channel(object):
@@ -29,11 +28,11 @@ class Channel(object):
     ptc=0
      tracks[4] = (0, 10)
     ptc=1
-     tracks[2] = (4, 5)
      tracks[0] = (1, 3)
+     tracks[2] = (4, 5)
     ptc=2
-     tracks[3] = (4, 4)
      tracks[1] = (1, 1)
+     tracks[3] = (4, 4)
     >>> for ptc, min_v, max_v in channel_model.fill_empty(0, 10):
     ...     print('ptc={} ({}, {})'.format(ptc, min_v, max_v))
     ptc=1 (0, 0)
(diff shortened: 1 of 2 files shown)


utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,74 @@ def __init__(self, tracks):
...
-        If track won't fit in any existing tree, allocate a new tree for the track"""
-        for idx, tree in enumerate(self.trees):
-            if not tree.overlaps(track[0], track[1] + 1):
-                tree.add(
-                    Interval(begin=track[0], end=track[1] + 1, data=track[2])
-                )
-                return
+    def _start_track(self, track):
+        self.trees.append(list())
+        self.trees[-1].append(track)
```
HackerFoo: Isn't this equivalent to `self.trees.append([track])`?

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,74 @@ def __init__(self, tracks):
...
 
-        self.trees.append(IntervalTree())
-        self.trees[-1].add(
-            Interval(begin=track[0], end=track[1] + 1, data=track[2])
-        )
+    def _add_track_to_last_tree(self, track, idx=-1):
+        self.trees[idx].append(track)
```
HackerFoo: Is the last tree `self.trees[-1]`? Because the name seems misleading if supplied with an `idx` argument.

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,74 @@ def __init__(self, tracks):
...
+        if len(by_low) > 0:
+            high = max(by_low)
+
+        while len(by_low) > 0:
+            track_low = min(by_low)
+            track_high, key = pop(track_low)
+
+            self._start_track((track_low, track_high, key))
+
+            while True:
```
HackerFoo:
```python
while track_high is not None:
```

otherwise line 96 would fail. Even if that couldn't happen, it's cleaner than `while True`.

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,74 @@ def __init__(self, tracks):
...
+
+            return track
+
+        for low, high, key in self.tracks:
+            if low not in by_low:
+                by_low[low] = []
+
+            by_low[low].append((high, key))
+
+        if len(by_low) > 0:
```
HackerFoo:
```python
if by_low:
```

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,74 @@ def __init__(self, tracks):
...
+            high = max(by_low)
+
+        while len(by_low) > 0:
```
HackerFoo:
```python
while by_low:
```

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,74 @@ def __init__(self, tracks):
...
         """pack all tracks"""
-        for track in self.tracks[::-1]:
-            self._place_track(track)
+
+        by_low = {}
+
+        def pop(low):
+            track = by_low[low].pop()
+
+            if len(by_low[low]) == 0:
```
HackerFoo:
```python
if not by_low[low]:
```

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,74 @@ def __init__(self, tracks):
...
+                track_high = None
+                for track_low in range(start, high + 1):
+                    if track_low in by_low:
+                        track_high, key = pop(track_low)
+                        self._add_track_to_last_tree(
+                            (track_low, track_high, key)
+                        )
+                        break
+
+                if track_high is None:
```
HackerFoo: You can remove this with the above change.

utils/lib/rr_graph/channel2.py
litghost: The function name is wrong, I'll update.

HackerFoo:
Could you please document the changes and how they improve the algorithm?

utils/lib/rr_graph/channel2.py
litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1038#discussion_r329696800

litghost:
> how they improve the algorithm?

The previous algorithm had an O(N*logN) search combined with high constant factor in managing the Interval objects.  In terms of runtime we went from > 30 seconds for 20k items to < 10 msec.

I'll add a comment on the new algorithms strategy.

HackerFoo:
> The previous algorithm had an O(N*logN) search combined with high constant factor in managing the Interval objects. In terms of runtime we went from > 30 seconds for 20k items to < 10 msec.

Nice!

utils/lib/rr_graph/channel2.py
litghost: Ya, fixed.

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,93 @@ def __init__(self, tracks):
...
+         1. Sort tracks by length, shortest tracks first.  Popping from back
+            of python lists is O(1).
+         2. Create stack for each starting values, inserting in length order.
+         3. Starting with the lowest starting value, greedly pack tracks
+            Algorithm weaknesses:
+             - Linear scan for lowest starting value
+             - Linear scan for packing
+
+            Both weaknesses are O(Number grid dim * Number of tracks) in
+            pathilogical case, however grid dimensions tend to be fairly small,
```
mithro:
```suggestion
            pathological case, however grid dimensions tend to be fairly small,
```

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,93 @@ def __init__(self, tracks):
...
+            (e.g. 50T is 150), so scans are practically fast.
+
+            If grid dimension size grows large, revisit how to find the lowest
```
mithro:
```suggestion
            If the grid dimension size grows large, revisit how to find the lowest
```

utils/lib/rr_graph/channel2.py
```diff
@@ -53,38 +52,93 @@ def __init__(self, tracks):
...
+            starting value and next bucket pack.  Relevant operation is given
+            coordinate, find next largest non-empty bucket.
+         3a. Pop largest track from smallest starting value, creating a new
```
mithro: Can you make these a proper nested list?

https://github.com/ralsina/rst-cheatsheet/blob/master/rst-cheatsheet.rst#lists

litghost: This isn't RST, and it unclear how we would even render this as RST?

utils/lib/rr_graph/channel2.py
HackerFoo: Huh. I guess Python [doesn't use linked lists](https://wiki.python.org/moin/TimeComplexity). Okay.

utils/lib/rr_graph/channel2.py
mithro: Docstring formatting is RST. This is because they end up being parsed by docutils/sphinx.

PR #1040, opened 2019-09-30
Update Yosys to latest merge with master.

litghost:


mithro:
Looks like something broke?

mkurc-ant:
I'm testing `srl32_init` and:
- `testbench_srl32_init_tb` passes
- `testbinch_srl32_init_tb` fails
- manually runned simulation of the `top_synth.v` also fails.

mkurc-ant:
I don't think that the failure is SRL related. I'm observing that in the test design, reference data from them ROM is generated one prescaler tick before compared to data from SRL. I've compared waveforms from testbench and testbinch.

mkurc-ant:
Apparently cause of the failure is weakness in the design code. There are some uninitialized registers which are being initialized to 1 by the bumped Yosys version. This does not affect functionality of the design but affects the assert which checks the error signal regardless of state of the reset signal. (the reset signal is generated inside the DUT).

PR #1042, opened 2019-10-01
Initial refactor of create edges.

litghost:
This is part of the refactoring to improve create edges time.

- Remove usage of multiprocess now that channel creation is fast
- Remove alive tracks set, and let sqlite filter (via index).
- Refactor edge creation into function.

### DIFF
--- a/xc7/utils/prjxray_create_edges.py
+++ b/xc7/utils/prjxray_create_edges.py
@@ -36,7 +36,7 @@ from prjxray.site_type import SitePinDirection
 from prjxray_constant_site_pins import yield_ties_to_wire
 from lib.connection_database import get_track_model, get_wire_in_tile_from_pin_name
 from lib.rr_graph.graph2 import NodeType
-import multiprocessing
+import sqlite3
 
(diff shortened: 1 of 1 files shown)


xc7/utils/prjxray_create_edges.py
```diff
@@ -1057,9 +1056,15 @@ def mark_track_liveness(
 
     print('{} Track aliveness committed'.format(now()))
 
-    print("{}: Build channels".format(datetime.datetime.now()))
-    build_channels(conn, pool, alive_tracks)
-    print("{}: Channels built".format(datetime.datetime.now()))
+    write_cur.execute("""BEGIN EXCLUSIVE TRANSACTION;""")
```
mithro: Should these index commands be part of the initial table SQL file? Or does it actually make sense from a performance point of view to only create them at this point?

litghost: Exactly.  In general, deferring index creation means there are no surprise rebuilding.  There is a possibly a way to supress that behavior, but this works.

PR #1043, opened 2019-10-01
Move logic of prjxray_create_edges.py into library out of utility

litghost:
Builds on #1042, simply moves functions out of utility file into auxiliary library.   This is useful for experiments and testing.


litghost:
> This is just moving code right?

Yep, but as shown by https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1043/commits/c4244cefbb72641a791be42ad2eb23d3c267ff68, even code movement can introduce lint regressions.

PR #1044, opened 2019-10-01
Lower peak memory usage of create edges

litghost:
- Refactor build channels to avoid high peak memory usage.
- Refactor create edge loop into function.
- Iteratively commit edges, rather than all at the end.
 - This lowers the peak memory consumption without minimal runtime
increase.


litghost:
https://github.com/SymbiFlow/symbiflow-arch-defs/compare/62a47bbda5336f668520c32478eb2feffef07b28..f0e700248af9a27d61c002345a0ae41bd20cac9e

xc7/utils/prjxray_edge_library.py
```diff
@@ -0,0 +1,1834 @@
...
+    for (track_pkey, ) in c.execute("""SELECT pkey FROM track;"""):
+        write_cur.execute(
+            "UPDATE track SET alive = ? WHERE pkey = ?;",
+            (track_pkey in alive_tracks, track_pkey)
+        )
+    write_cur.execute("""COMMIT TRANSACTION;""")
+
+    print('{} Track aliveness committed'.format(now()))
+
+    write_cur.execute("""BEGIN EXCLUSIVE TRANSACTION;""")
```
HackerFoo: Does each line need to be executed separately?

xc7/utils/prjxray_edge_library.py
```diff
@@ -0,0 +1,1834 @@
...
+            phy_tile_pkey, pip_in_tile_pkey, backward) VALUES (?, ?, ?, ?, ?, ?)""",
+        edges
+    )
+    write_cur.execute("""COMMIT TRANSACTION;""")
+
+
+def create_edge_indices(conn):
+    write_cur = conn.cursor()
+
+    write_cur.execute("""BEGIN EXCLUSIVE TRANSACTION;""")
```
HackerFoo: Does each line need to be executed separately?

xc7/utils/prjxray_edge_library.py
litghost: Yes, https://docs.python.org/2/library/sqlite3.html#sqlite3.Cursor.execute

PR #1048, opened 2019-10-02
Support for PLLE2_ADV for fasm2bels.

mkurc-ant:
This PR adds support for the PLLE2_ADV decoding to fasm2bels. Requires https://github.com/SymbiFlow/prjxray/pull/1077 to work correctly (routing decoding issues).

The decoding of COMPENSATION parameter may require improvement as the relation between fasm tags and this parameter is somewhat unclear.

### DIFF
--- a/xc7/fasm2bels/cmt_models.py
+++ b/xc7/fasm2bels/cmt_models.py
@@ -4,6 +4,7 @@ from .verilog_modeling import Bel, Site
 
 # =============================================================================
 
+
 def get_pll_site(db, grid, tile, site):
     """ Return the prjxray.tile.Site object for the given PLL site. """
     gridinfo = grid.gridinfo_at_tilename(tile)
@@ -44,7 +45,10 @@ def process_pll(conn, top, tile_name, features):
         return
 
     # Create the site
-    site = Site(pll_features, get_pll_site(top.db, top.grid, tile=tile_name, site='PLLE2_ADV'))
+    site = Site(
+        pll_features,
+        get_pll_site(top.db, top.grid, tile=tile_name, site='PLLE2_ADV')
+    )
 
     # If the PLL is not used then skip the rest
     if not site.has_feature("IN_USE"):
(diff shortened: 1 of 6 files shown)


xc7/fasm2bels/cmt_models.py
```diff
@@ -0,0 +1,145 @@
...
+
+            # Add output source
+            site.add_source(pll, 'CLK'+clkout, 'CLK'+clkout)
+
+            # Calculate the divider and duty cycle
+            high_time = decode_multi_bit_feature(features, 'CLK{}_CLKOUT1_HIGH_TIME'.format(clkout))
+            low_time  = decode_multi_bit_feature(features, 'CLK{}_CLKOUT1_LOW_TIME'.format(clkout))
+
+            divider = high_time + low_time
+            duty = high_time / (low_time + high_time)
```
litghost: This needs to handle the EDGE feature, https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/segbits_cmt_top_l_upper_t.db#L27
The EDGE enables half-step divider changes, see XAPP888.

mkurc-ant: Fixed that. Another thing is that DivReg also contains an EDGE bit but there seems no control attribute on the PLLE2_ADV that can control it.

mkurc-ant:
With added dumping of PPIPs for INT_INTERFACE_[LR] tiles (in https://github.com/SymbiFlow/prjxray/pull/1077) i managed to successfully decode the PLL and its routing (PLLE2_ADV minitest). Moreover I then feed the results through Vivado which yielded in working design in HW. However, I had to disable some DRC checks regarding PLL input frequency specification which is not stored in the bitstream hence cannot be decoded.

xc7/fasm2bels/cmt_models.py
litghost: When DIVCLK_DIVIDE is odd, DIVREG.EDGE is set I believe?

xc7/fasm2bels/cmt_models.py
```diff
@@ -0,0 +1,164 @@
...
+    gridinfo = grid.gridinfo_at_tilename(tile)
+    tile_type = db.get_tile_type(gridinfo.tile_type)
+
+    sites = list(tile_type.get_instance_sites(gridinfo))
+    assert len(sites) == 1, sites
+
+    return sites[0]
+
+
+def decode_multi_bit_feature(features, target_feature):
```
litghost: I think we have a couple variants of this amoung the models.  Can we consolidate?

xc7/fasm2bels/cmt_models.py
```diff
@@ -0,0 +1,164 @@
...
+    if site.has_feature('COMPENSATION.INTERNAL'):
+        pll.parameters['COMPENSATION'] = '"INTERNAL"'
+
+    # Built-in inverters
+    pll.parameters['IS_CLKINSEL_INVERTED'] =\
+        "1'b1" if site.has_feature('INV_CLKINSEL') else "1'b0"
+    pll.parameters['IS_PWRDWN_INVERTED'] =\
+        "1'b1" if site.has_feature('ZINV_PWRDWN') else "1'b0"
+    pll.parameters['IS_RST_INVERTED'] =\
+        "1'b1" if site.has_feature('ZINV_RST') else "1'b0"
```
litghost: Missing parameter for BANDWIDTH, which is determined (somehow) via the FILT regs.  Might need to a make a table, which would also be useful when going from verilog -> eblif -> VPR.

litghost:
> With added dumping of PPIPs for INT_INTERFACE_[LR] tiles (in [SymbiFlow/prjxray#1077](https://github.com/SymbiFlow/prjxray/pull/1077)) i managed to successfully decode the PLL and its routing (PLLE2_ADV minitest). Moreover I then feed the results through Vivado which yielded in working design in HW. However, I had to disable some DRC checks regarding PLL input frequency specification which is not stored in the bitstream hence cannot be decoded.

Awesome!

xc7/fasm2bels/cmt_models.py
mkurc-ant: Done, but it seems that "OPTIMIZED" and "HIGH" are the same.

PR #1050, opened 2019-10-03
add prjxray_create_pinmap.py

HackerFoo:


xc7/utils/prjxray_create_pinmap_csv.py
```diff
@@ -0,0 +1,70 @@
+#! /usr/bin/env python3
+""" Tool generate a pin map CSV from channels.db and a *_package_pins.csv for pin placement. """
+from __future__ import print_function
+import argparse
+import sys
+import csv
+import sqlite3
+
+
+def get_vpr_coords_from_tile_name(conn, tile_name):
```
litghost: This will not work after the IOI tile is split, as the one phy_tile will be located into two VPR tile locations.

PR #1047, opened 2019-10-02
Initial support for RIOB33M routing

litghost:
This PR will hopefully add support for routing to RIOB33M pins.

Approximate work:
 - [x] Have `prjxray_form_channels` generate database with split RIOB33M tiles
 - [x] Have `prjxray_assign_pin_direction` generate pin directions for new IO PAD tiles
 - [x] Have `prjxray_create_edges` generate graph_nodes and graph_edges for new IO PAD tiles
 - [x] Have initial RIOPAD_M tile
 - [x] Update CMake flow to handle lack of synth_tiles
 - [x] Add CMake option to generate partial routing graph for debugging ease
 - [x] Add lower right CMT device/device type that includes RIOPAD_M tile type
 - [x] Verify prjxray_routing_import completes with RIOPAD_M in arch
 - [x] Verify routing graph connects RIOPAD_M IPIN/OPINs to arch
 - [x] Integrate pinmap.csv tool from @HackerFoo to generate valid IO placement constraints
 - [x] Merge RIOB33M tile into RIOPAD_M
 - [x] Update prjxray_import_tile to handle merged tiles and within tile connections
 - [x] Hack VPR to new search coordinates
 - [x] Verify VPR routes between RIOPAD_M
 - [x] Verify that constant network is integrated
 - [x] Integrate new "fasm_placeholders" features to emit valid FASM prefixes for IOB and IOI sites.
 - [ ] Have VPR that integrates lookahead fix that works with existing ROI and https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/285
 - [x] Add fixed FASM features as needed to emit required FASM features.

PR #1051, opened 2019-10-03
Add site instance table to connection database.

litghost:


litghost:
> Don't necessarily know _why_ you are doing this - so might be better to get @mkurc-ant or @HackerFoo to review.

It is used for package pin mapping (and future LOC mapping): https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1047/commits/3cf23c61ea11dc6b83e53735747bc7d03a983818

PR #1054, opened 2019-10-04
equivalent tiles: modified import scripts to add direct pin mappings

acomodi:

This PR adds the direct pin mappings to the architecture files.


acomodi:
@litghost FYI, CI will be red as VtR is still not present in conda. As soon as it does, I will add the new VtR conda-package to this PR.

Issue #1068, opened 2019-10-11
Merge conflict prevents producing octopus merge


mithro:
If there is a conflict when trying to do an octopus merge then the merging WIP tool fails.

_Originally posted by @litghost in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1067#issuecomment-540823673_

mithro:
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/utils/update_tools/update_tools.py

litghost:
```
subprocess.CalledProcessError: Command 'cd /usr/local/google/home/keithrothman/cat_x/update_wip/vtr-verilog-to-routing && git merge wip/add-issue-template wip/avoid-criticality-issue wip/avoid_unordered_set wip/connection_box_serdes_merge_squashed wip/disable-check-route-option wip/equivalent-sites wip/genfasm-placeholders wip/prune_non_configurable_nodes wip/refactor_heap2 wip/rr-graph-binary-read-write wip/symbiflow-badger wip/use_expansion_for_place_delay && cd -' returned non-zero exit status 1.
```

New code doesn't handle failed octopus merge

PR #1069, opened 2019-10-11
update_tool: fixed wrong Octopus merge when having conflicts

acomodi:

This PR fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1068. The strategy now is to fallback to a Sequential merge of all the `wip/` branches whenever there is a conflict during the Octopus Merge.

In this way we are sure that all the branches are included in the `master+wip-next` branch.

### DIFF
--- a/utils/update_tools/update_tools.py
+++ b/utils/update_tools/update_tools.py
@@ -21,6 +21,7 @@ branch.
 """
 
 import os
+import subprocess
 import argparse
 import git
 
@@ -69,7 +70,7 @@ def merge_branches(location, branches):
     The branches have to be in string format
     """
 
-    os.system("cd {} && git merge {} && cd -".format(location, branches))
+    subprocess.check_call("cd {} && git merge {} && cd -".format(location, branches), shell=True)
 
 
 def rebase_continue_rec(g):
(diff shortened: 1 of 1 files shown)


utils/update_tools/update_tools.py
```diff
@@ -63,6 +63,15 @@ def solve_conflicts(g, branch=""):
     g.add(".")
 
 
+def merge_branches(location, branches):
+    """
+    Merge one or more branches into the current branch.
+    The branches have to be in string format
+    """
+
+    os.system("cd {} && git merge {} && cd -".format(location, branches))
```
litghost: Why use os.system here, rather than gitpython?

litghost: Looks like you need to set octopus=True?https://github.com/gitpython-developers/GitPython/issues/171

acomodi: I have tried multiple ways to do this, but it seems kinda complex to avoid the above described issue. `merge_base` should only find the common ancestor of the branches that are specified, without performing the merge. Probably `merge_tree` can be used, but I am no sure it works for Octopus merges.

IMO it would be better to leave it with the `os.system` call, at least we are sure that it works as we expect.

litghost: Okay, rather than using `os.system`, please use `subprocess.check_call(xxx, shell=True)`

acomodi: Done

PR #1060, opened 2019-10-09
[WIP] BUFG support

litghost:
Work on variant of #960 

Current state of VTR hack: 
[hack.patch.zip](https://github.com/SymbiFlow/symbiflow-arch-defs/files/3705457/hack.patch.zip)

PR #1070, opened 2019-10-11
xc7: add clock tiles BUFG and BUFH

acomodi:
This PR is part of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1060 and adds the BUFH and BUFG tile types.

Once this lands the next patches can be added to have a working ROI-less and clock-tree driven implementations.


xc7/make/arch_define.cmake
```diff
@@ -83,6 +83,7 @@ function(ADD_XC7_ARCH_DEFINE)
         --pcf \${INPUT_IO_FILE} \
         --eblif \${OUT_EBLIF} \
         --top \${TOP} \
+        --allow_orphan_sinks \
```
litghost: Can we try dropping this option?  I don't believe it is required anymore for fasm2bels to work.

acomodi: Done, should I locally check that fasm2bels works locally or is it already tested in CI?

litghost: Locally.  We cannot test ROI-less stuff until the lookahead search is updated.

litghost:
Let's reword the commit message on 3022931c524885338f7556b56a4df83a3424bd83, and maybe break up the commits.

PR #1073, opened 2019-10-12
Integrate VPR with explicitly lookahead search coordinates

litghost:
This also enables testing of ROI-less binaries.

@acomod / @mkurc-ant This should enable the ROI-less stuff without needing VPR patches.


acomodi:
@litghost I have included the changes introduced with this PR in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1075 so that the clock network is tested in CI, which will hopefully turn green.

litghost:
Merged as part of #1075

PR #1077, opened 2019-10-15
Pin libxml2 to version with fix.

### LINKED ISSUE #1066: lxml generates exception when writing more than 2 GB incrementally
There is a int32_t overflow bug in libxml2 (https://gitlab.gnome.org/GNOME/libxml2/issues/112) that causes an exception to be generated when more than 2 GB of XML is written.  The upstream maintainers have not merged a fix yet (see https://gitlab.gnome.org/GNOME/libxml2/merge_requests/54 or https://gitlab.gnome.org/GNOME/libxml2/merge_requests/55).  For now, libxml2 needs to be replaced.

Instructions:

1. Have symbiflow-arch-defs setup conda (e.g. `make all_conda`)
2. Clone `https://github.com/SymbiFlow/libxml2-feedstock` into a new directory
3. Execute the following commands:
```
export CONDA_DIR=<path to conda>
export PATH=$CONDA_DIR/bin:$PATH
# Make sure conda-build is installed in conda
conda install conda-build

# Build replacement package
cd <libxml2-feedstock directory>
conda-build .

# Install replacement package
conda install --use-local libxml2
```
If the conda environment is destroyed, the steps above will need to be repeated.

litghost:
Fixes #1066

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,7 @@ add_conda_package(
 add_conda_package(
   NAME libxml2
   PROVIDES xmllint
+  PACKAGE_SPEC "libxml2 2.9.9 h14c3975_5"
   )
 add_conda_package(
   NAME pytest


litghost:
@mkurc-ant Workaround in #1066 should not be required after this PR is merged.

PR #1081, opened 2019-10-15
Relax router effort when creating the place delay matrix.

litghost:
Testing to see that picosoc and murax still behave after this.


litghost:
This change may decrease full 50T placer delay computation time.  This change has no affect on the router lookahead computation.

Issue #1065, opened 2019-10-10
BUFH route through


acomodi:
With https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1060 we are able to build a simple_ff test using the clocking network and ROI-less.

The next step is to get rid of the explicit BUFH instantiation that will prevent the whole 50T to have a valid clocking network. This because the BUFH is now placed in a specific location and cannot be accessed by other clock region.

The idea to solve this is to let VPR route through the BUFH tiles creating a CHAN to CHAN connection. Then with check_features add the relative feature of the BUFH that have been used during routing.

acomodi:
@litghost FYI

litghost:
Connecting the BUFH IPIN to OPIN is the simplest answer, but it won't work because of https://github.com/SymbiFlow/vtr-verilog-to-routing/blob/master%2Bwip/vpr/src/route/route_timing.cpp#L1859

Instead, a CHAN to CHAN (or CHAN to OPIN) edge is required.  Luckily there is a convenient graph node created during create edges [here](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/utils/prjxray_edge_library.py#L347).  `find_wire_node` creates a CHAN node before each IPIN and a CHAN node after each OPIN that should be suitable for BUFH route through.

On another note, the prjxray-db already has a pip that represents the BUFH route through, but it has been ignored up till now because it is marked "is_psuedo", which represents that it is a site route through.  This is in fact what we are doing, so it can be used.  Example psuedo-pip for BUFH route through: https://raw.githubusercontent.com/SymbiFlow/prjxray-db/master/artix7/tile_type_CLK_HROW_BOT_R.json , search for CLK_HROW_BOT_R.CLK_HROW_CK_MUX_OUT_L0->>CLK_HROW_CK_HCLK_OUT_L0

Note that the pip has timing information representing the timing cost of routing through the BUFH.

acomodi:
@litghost I have tackled this issue today, but I still have some things I do not fully understand:

[...]

PR #1084, opened 2019-10-16
Bufh route-through

acomodi:

This PR is a very initial attempt to let the BUFH being routed without placing it.

Issue tracker: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1065


acomodi:
@litghost FYI.
With these changes, the edge between the two nodes has been created. The only problem now is that, as you have predicted, the nodes that are being connected are IPINs and OPINs.

I could not create the CHANX/Y before and after the I/OPINS as a `track_pkey` is required to use the `find_wire_node` function that creates the CHANX/Y graph nodes. I'll continue working on this branch and push new updates as I progress.

acomodi:
I have thought that maybe it is possible to transform those I/OPINs nodes in actual CHANX or CHANY nodes. I guess that from a VPR point of view it should make no difference.

acomodi:
> I have thought that maybe it is possible to transform those I/OPINs nodes in actual CHANX or CHANY nodes. I guess that from a VPR point of view it should make no difference.

I take my words back. I think that I was looking at this in a wrong way. The pip that I am exploiting is relative to only site nodes, that is why no track is found in the connectors.
What I really need to do, I believe, is to get all the tile IPINs and use the CHAN that should be already be provided there and connect it to the OPIN of the BUFH site.

litghost:
There is a Chan for the site pins via https://github.com/SymbiFlow/symbiflow-arch-defs/blob/f179e823b4da40d5de3ec3c2ec883552e3adc3fb/xc7/utils/prjxray_edge_library.py#L347 , use that chan

acomodi:
@litghost The problem is, where do I take the track_pkey from? It should be necessary for the `find_wire_node` function

xc7/utils/prjxray_edge_library.py
```diff
@@ -1801,9 +1814,13 @@ def create_and_insert_edges(
             if 'PADOUT0' in pip.name or 'PADOUT1' in pip.name:
                 continue
 
-            if pip.is_pseudo:
+            if pip.is_pseudo and "CLK_HROW_CK" not in pip.name:
                 continue
 
+            dless = None
+            if "CLK_HROW_CK" in pip.name:
+                dless = delayless_switch_pkey
```
litghost: This is the wrong switch to use, why not use the switch associated with the psuedo pip, which had timing through the feature?

acomodi: All right

acomodi:
The `simple_ff_x1y0` is routed successfully. The problem is `print_route` now:

```
                    if (!is_io_type(device_ctx.grid[ilow][jlow].type) && (rr_type == IPIN || rr_type == OPIN)) {
                        int pin_num = device_ctx.rr_nodes[inode].ptc_num();
                        int xoffset = device_ctx.grid[ilow][jlow].width_offset;
                        int yoffset = device_ctx.grid[ilow][jlow].height_offset;
                        ClusterBlockId iblock = place_ctx.grid_blocks[ilow - xoffset][jlow - yoffset].blocks[0];
                        VTR_ASSERT(iblock);
                        t_pb_graph_pin* pb_pin = get_pb_graph_node_pin_from_block_pin(iblock, pin_num);
                        t_pb_type* pb_type = pb_pin->parent_node->pb_type;
                        fprintf(fp, " %s.%s[%d] ", pb_type->name, pb_pin->port->name, pb_pin->pin_number);
                    }
```

This is the piece of code that fails, specifically the VTR_ASSERT. 
I believe that I should also create the CHANX for the OPIN, as the problem, I think, is given by the fact that the OPIN is at the BUFH location which was never placed, hence no Block is present in there.

litghost:
Why not use the Chan on opin, same as the IPin?

acomodi:
Ya, I have started a build with this fix, should solve the issue

xc7/utils/prjxray_routing_import.py
```diff
@@ -709,7 +709,7 @@ def create_get_pip_wire_names(conn):
     def get_pip_wire_names(pip_pkey):
         cur.execute(
             """SELECT src_wire_in_tile_pkey, dest_wire_in_tile_pkey
-            FROM pip_in_tile WHERE pkey = ? AND is_pseudo = 0;""",
+            FROM pip_in_tile WHERE pkey = ? AND (is_pseudo = 0 OR name LIKE '%CLK_HROW_CK%');""",
```
litghost: I actually think this should just be `WHERE pkey = ?`, I think the previous `is_pseudo = 0` was not required given that this function takes a primary key.

acomodi: Done.

xc7/utils/prjxray_edge_library.py
```diff
@@ -688,6 +688,26 @@ def connect_at(
                     yield (src_node, switch_pkey, dest_node, pip.pip_pkey)
                     return
 
+        elif self.pins and other_connector.pins and pip.is_pseudo:
+            for pin_dir in self.pins.edge_map:
+                if pin_dir in other_connector.pins.edge_map:
+                    src_node = self.pins.edge_map[pin_dir]
+                    dest_node = other_connector.pins.edge_map[pin_dir]
+
+                    src_wire_switch_pkey, src_wire_node = self.find_wire_node(
```
litghost: Need to modify `find_wire_node` to assert if `track_graph_node_pkey` is needed, but is None.

xc7/utils/prjxray_edge_library.py
```diff
@@ -688,6 +688,26 @@ def connect_at(
...
+                        src_wire_pkey, src_node, None
+                    )
+
+                    dest_wire_switch_pkey, dest_wire_node = self.find_wire_node(
+                        dest_wire_pkey, dest_node, None
+                    )
+
+                    yield (
+                        src_wire_node, src_wire_switch_pkey, dest_wire_node,
```
litghost: Do not use src_wire_switch_pkey, use the switch belonging to the pip itself.  Should be able to use 
```
                    switch_pkey = pip.get_pip_switch(
                        src_wire_pkey, dest_wire_pkey
                    )
``` or equiv.

Comment (acomodi) [reply xc7/utils/prjxray_edge_library.py]: Done

File: xc7/utils/prjxray_edge_library.py (hunk 3 above)
Comment (acomodi) [reply xc7/utils/prjxray_edge_library.py]: Done.

**Comment by acomodi** [comment]
@litghost I have added the fasm feature. The `simple_ff` tests works on HW, I'll do some other tests, but it seems this PR is doing the job correctly.

File: xc7/utils/prjxray_edge_library.py (hunk 1 above)
Comment (acomodi) [reply xc7/utils/prjxray_edge_library.py]: Done

**Comment by acomodi** [comment]
`counter_basys3_x1y0` test works as well on HW

File: xc7/utils/prjxray_routing_import.py [hunk 5]
```diff
@@ -230,6 +231,28 @@ def check_feature(feature):
...
+        x_loc_str = bufhce_loc_string[0]
+        if 'L' in x_loc_str:
+            x_loc = 0
+        elif 'R' in x_loc_str:
+            x_loc = 1
+        else:
+            assert False, "Impossible to determine X location of BUFHCE"
+
+        y_loc = bufhce_loc_string[1:]
+        bufhce_loc = 'BUFHCE_X{}Y{}'.format(x_loc, y_loc)
```
litghost: This code only works for the lowest block of BUFHCE, because the y_loc increases on the higher CMTs, but the segbits is modulo the height of the number of BUFHCE's (e.g. 12).  The higher ones should by y_loc % 12, but let's build that table from the connection database.  So please add something like `populate_bufg_rebuf_map` or `populate_hclk_cmt_tiles` to create a lookup.

xc7/utils/prjxray_routing_import.py
```diff
@@ -230,6 +231,28 @@ def check_feature(feature):
 
         return ' '.join((feature, enable_buffer_feature))
 
+    if CLK_HROW_CK_MUX_REGEX.fullmatch(feature_path[-1]):
+        bufhce_loc_string = feature_path[-1].split('_')[-1]
```
litghost: Because you already have a regular expression, please add capture groups to the `CLK_HROW_CK_MUX_REGEX` and use those capture groups, instead of ad-hoc string slicing.

xc7/utils/prjxray_edge_library.py
```diff
@@ -688,6 +692,28 @@ def connect_at(
                     yield (src_node, switch_pkey, dest_node, pip.pip_pkey)
                     return
 
+        elif self.pins and other_connector.pins and pip.is_pseudo:
+            switch_pkey = pip.get_pip_switch(src_wire_pkey, dest_wire_pkey)
```
litghost: Let's a comment what is going on in this block.

xc7/utils/prjxray_edge_library.py
```diff
@@ -688,6 +692,28 @@ def connect_at(
...
+
+            for pin_dir in self.pins.edge_map:
+                if pin_dir in other_connector.pins.edge_map:
+                    src_node = self.pins.edge_map[pin_dir]
+                    dest_node = other_connector.pins.edge_map[pin_dir]
+
+                    src_wire_switch_pkey, src_wire_node = self.find_wire_node(
```
litghost: In the same way the edge to channel or channel to edge should be CHAN to IPIN or OPIN to CHAN, let's assert that this src_node is an IPIN and dest_node is an OPIN.

xc7/utils/prjxray_edge_library.py
```diff
@@ -979,7 +1006,7 @@ def make_connection(
 
     pip_obj = find_pip(tile_type, pip.name)
 
-    assert not pip_obj.is_pseudo
+    assert not pip_obj.is_pseudo or "CLK_HROW_CK" in pip.name
```
litghost: Let's add a comment on what is going on here

xc7/utils/prjxray_edge_library.py
```diff
@@ -1801,7 +1829,7 @@ def create_and_insert_edges(
             if 'PADOUT0' in pip.name or 'PADOUT1' in pip.name:
                 continue
 
-            if pip.is_pseudo:
+            if pip.is_pseudo and "CLK_HROW_CK" not in pip.name:
```
litghost: Comment

xc7/utils/prjxray_routing_import.py
```diff
@@ -230,6 +231,28 @@ def check_feature(feature):
 
         return ' '.join((feature, enable_buffer_feature))
 
+    if CLK_HROW_CK_MUX_REGEX.fullmatch(feature_path[-1]):
```
litghost: Let's add a comment in words what is going on here.

xc7/utils/prjxray_routing_import.py
acomodi: Done

xc7/utils/prjxray_edge_library.py
acomodi: Done

xc7/utils/prjxray_edge_library.py
acomodi: Done

xc7/utils/prjxray_edge_library.py
acomodi: Done

xc7/utils/prjxray_routing_import.py
acomodi: Done

PR #1087, opened 2019-10-16
Update to new VPR that includes faster place delay matrix comp.

litghost:


litghost:
Affect on 7-series was good (~20 minute runtime reduction), but segfaults showed up in testarch and ice40.  These were discovered and fixed when running against VTR CI, new VTR master+wip is being created now.

PR #1098, opened 2019-10-21
Update print_connection_map_lookahead.py to match latest capnp.

litghost:


litghost:
> LGTM. CI though went red, probably for some unrelated infrastructure failure.

Flaky failure in XC7 that I reran and turned, and it didn't update the overarching job as green.

PR #1102, opened 2019-10-22
Add OSERDES support

acomodi:

This PR is to add initial support to OSERDES and ISERDES primitives.


acomodi:
@litghost @mkurc-ant FYI. I have opened this early implementation to support IOSERDESes. The PR is still incomplete, but if there are suggestions or things that should be changed, please let me know.

Currently VPR fails during XML parsing as the clocking and timing information is incorrect (I am unsure on what are the ports that should be clocked).

Moreover, I will slightly modify the [oserdes minitest](https://github.com/SymbiFlow/prjxray/tree/master/minitests/oserdes) to test it.

As soon as the OSERDES works I'll proceed with the ISERDES addition.

acomodi:
I have an issue for which, when I instantiate more than one `OSERDES`, the packer correctly packs the first one and fails to pack the second one with the following error:

```
        S
        B(222:cluster-external source (LB_SOURCE)-->42:'BLK-TL-IOPAD_S[0].OLOGICE3_T1[0]') (42:'BLK-TL-IOPAD_S[0].OLOGICE3_T1[0]'-->180:'OLOGICE3[0].T1[0]') (180:'OLOGICE3[0].T1[0]'-->206:'OSERDESE2[0].T1[0]') (206:'OSERDESE2[0].T1[0]'-->259:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T1[0]))
        B(222:cluster-external source (LB_SOURCE)-->43:'BLK-TL-IOPAD_S[0].OLOGICE3_T2[0]') (43:'BLK-TL-IOPAD_S[0].OLOGICE3_T2[0]'-->181:'OLOGICE3[0].T2[0]') (181:'OLOGICE3[0].T2[0]'-->207:'OSERDESE2[0].T2[0]') (207:'OSERDESE2[0].T2[0]'-->260:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T2[0]))
        B(222:cluster-external source (LB_SOURCE)-->44:'BLK-TL-IOPAD_S[0].OLOGICE3_T3[0]') (44:'BLK-TL-IOPAD_S[0].OLOGICE3_T3[0]'-->182:'OLOGICE3[0].T3[0]') (182:'OLOGICE3[0].T3[0]'-->208:'OSERDESE2[0].T3[0]') (208:'OSERDESE2[0].T3[0]'-->261:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T3[0]))
        B(222:cluster-external source (LB_SOURCE)-->45:'BLK-TL-IOPAD_S[0].OLOGICE3_T4[0]') (45:'BLK-TL-IOPAD_S[0].OLOGICE3_T4[0]'-->183:'OLOGICE3[0].T4[0]') (183:'OLOGICE3[0].T4[0]'-->209:'OSERDESE2[0].T4[0]') (209:'OSERDESE2[0].T4[0]'-->262:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T4[0]))
Routing finished
        S
        B(222:cluster-external source (LB_SOURCE)-->42:'BLK-TL-IOPAD_S[0].OLOGICE3_T1[0]') (42:'BLK-TL-IOPAD_S[0].OLOGICE3_T1[0]'-->180:'OLOGICE3[0].T1[0]') (180:'OLOGICE3[0].T1[0]'-->206:'OSERDESE2[0].T1[0]') (206:'OSERDESE2[0].T1[0]'-->259:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T1[0]))
        B(222:cluster-external source (LB_SOURCE)-->43:'BLK-TL-IOPAD_S[0].OLOGICE3_T2[0]') (43:'BLK-TL-IOPAD_S[0].OLOGICE3_T2[0]'-->181:'OLOGICE3[0].T2[0]') (181:'OLOGICE3[0].T2[0]'-->207:'OSERDESE2[0].T2[0]') (207:'OSERDESE2[0].T2[0]'-->260:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T2[0]))
        B(222:cluster-external source (LB_SOURCE)-->44:'BLK-TL-IOPAD_S[0].OLOGICE3_T3[0]') (44:'BLK-TL-IOPAD_S[0].OLOGICE3_T3[0]'-->182:'OLOGICE3[0].T3[0]') (182:'OLOGICE3[0].T3[0]'-->208:'OSERDESE2[0].T3[0]') (208:'OSERDESE2[0].T3[0]'-->261:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T3[0]))
        B(222:cluster-external source (LB_SOURCE)-->45:'BLK-TL-IOPAD_S[0].OLOGICE3_T4[0]') (45:'BLK-TL-IOPAD_S[0].OLOGICE3_T4[0]'-->183:'OLOGICE3[0].T4[0]') (183:'OLOGICE3[0].T4[0]'-->209:'OSERDESE2[0].T4[0]') (209:'OSERDESE2[0].T4[0]'-->262:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].T4[0]))
        B(222:cluster-external source (LB_SOURCE)-->47:'BLK-TL-IOPAD_S[0].OLOGICE3_TCE[0]') (47:'BLK-TL-IOPAD_S[0].OLOGICE3_TCE[0]'-->185:'OLOGICE3[0].TCE[0]') (185:'OLOGICE3[0].TCE[0]'-->211:'OSERDESE2[0].TCE[0]') (211:'OSERDESE2[0].TCE[0]'-->264:cluster-internal sink (LB_SINK accessible via architecture pins: BLK-TL-IOPAD_S[0]/OLOGICE3[0]/OSERDESE2[0].TCE[0]))
                        No possible routing path from 'OSERDESE2[0].OQ[0]' to cluster-external sink (LB_SINK): needed for net 'o_dat' from net pin 'oserdes_test2.oserdes.OQ[0]'
Routing finished
        S
Routing was impossible!
                        FAILED Detailed Routing Legality
                FAILED No candidate primitives available
        FAILED_SEED: Block Type BLK-TL-IOPAD_S
# Packing took 0.08 seconds (max_rss 25.1 MiB, delta_rss +2.3 MiB)
Error 1:
Type: Packing
File: /tmp/really-really-really-really-really-really-really-really-really-really-really-really-really-long-path/conda/conda-bld/vtr_1571664918758/work/vpr/src/pack/cluster.cpp
Line: 2057
Message: Can not find any logic block that can implement molecule.
        Atom oserdes_test2.oserdes (OSERDESE2_VPR)


```

The problem is in the `OQ` signal which cannot be routed to the `.outpad`.

This is weird, as, if only one `OSERDES` is isntantiated, the build completes successfully. I am thinking that no new cluster is created and VPR is trying to pack everything in the same cluster, hence the OQ signal is already occupied by the first complex block.

litghost:
> I have an issue for which, when I instantiate more than one `OSERDES`, the packer correctly packs the first one and fails to pack the second one with the following error:

You need a pack pattern between the OBUF and the OSERDES.  If the packer tried to pack the OSERDES first, it will notice that it cannot get to the fabric from OQ, and fail.  If instead, the packer packs the OBUF first, and then the OSERDES, everything is fine.  The solution is make the OBUF + OSERDES a molecule, and then it all works out.

acomodi:
@litghost where to add the pack pattern though? I cannot do that directly in the OLOGICE3 primitive as it will be included in the three different IOPAD pb_types, each of one containing the corresponding IOB33 pb_types with their OBUFs.

The problem is that the OLOGICE3 primitive should not be aware in which pb_type it will be included, I guess.

litghost:
> @litghost where to add the pack pattern though? I cannot do that directly in the OLOGICE3 primitive as it will be included in the three different IOPAD pb_types, each of one containing the corresponding IOB33 pb_types with their OBUFs.
> 
> The problem is that the OLOGICE3 primitive should not be aware in which pb_type it will be included, I guess.

I completely agree.  This is a common problem I've had when defining pack patterns, where what I really want to say to VPR is "make a molecule between this atom and every downstream atom".  @kgugala Can we use V2X here to make defining this kind of pb_type agnostic pack pattern easier?  Basically an annotation on OLOGICE3 OQ that is "make a pack pattern for each variant"?

This would come in handle in the SLICEL/SLICEM too.

alessandrocomodi:
@litghost I think I found the reason why it fails:

This is the first atom that is packed:
```
Complex block 0:                                                                                                                                                       
        Seed: 'oserdes_test1.oserdes' (OSERDESE2_VPR)                                                                                                                                                                                                                                                     Try pack molecule: 'oserdes_test1.oserdes' (OSERDESE2_VPR)                                                                                             
                        Placed atom 'oserdes_test1.oserdes' (OSERDESE2_VPR) at BLK-TL-IOPAD[0][default]/OLOGICE3[0][OSERDES]/OSERDESE2[0]                                                                                                                                                 
                PASSED pack molecule                                                                                                                                   
        PASSED_SEED: Block Type BLK-TL-IOPAD                                                                                                                                                                                                                                              Not enough resources expand FPGA size to (150 x 162)                                                                                                                   
                Try pack molecule: 'out:led[2]' (.output)                                                                                                                                                                                                                                                         Placed atom 'out:led[2]' (.output) at BLK-TL-IOPAD[0][default]/IOB33[0][NO_OBUF]/outpad[0]                                                                                                                                                                        
                PASSED pack molecule                                                                                                                                                                                                                                                      
        PASSED: 'out:led[2]' (.output)                                                                                                                                                       
        No related molecule found and unrelated clustering disabled       
```

The IOB33 mode is `NO_OBUF` which connects the `O` pin to the `outpad`

for the second atom (the one which fails) this is the output:
```
Complex block 1:                                                                                                                                                                                                                                                                                  Seed: 'oserdes_test2.oserdes' (OSERDESE2_VPR)                                                                                                                                                      
                Try pack molecule: 'oserdes_test2.oserdes' (OSERDESE2_VPR)                                                                                                                                                                                                                                        Placed atom 'oserdes_test2.oserdes' (OSERDESE2_VPR) at BLK-TL-IOPAD[0][default]/OLOGICE3[0][OSERDES]/OSERDESE2[0]                                                                 
                PASSED pack molecule                                                                                                                                                                                                                                                      
        PASSED_SEED: Block Type BLK-TL-IOPAD                                                                                                                                                                                                                                              
                Try pack molecule: 'out:o_dat' (.output)                                                                                                                                                                                                                                  
                        Placed atom 'out:o_dat' (.output) at BLK-TL-IOPAD[0][default]/IOB33[0][OBUF]/outpad[0]                                                                                       
                PASSED pack molecule                                                                                                                                                                                                                                                      
        PASSED: 'out:o_dat' (.output)    
```

The `OBUF` mode is selected, and it makes the routing impossible from the `O` pin to the `.outpad` as the pb_type definition is the following:

```
        <mode name="OBUF">
          <pb_type blif_model=".subckt OUTBUF_VPR" name="OUTBUF_VPR" num_pb="1">
            <input name="IN" num_pins="1"/>
            <input name="TRI" num_pins="1"/>
            <output name="OUT" num_pins="1"/>
            <delay_constant in_port="OUTBUF_VPR.IN" max="10e-12" out_port="OUTBUF_VPR.OUT"/>
          </pb_type>
          <pb_type blif_model=".output" name="outpad" num_pb="1">
            <input name="outpad" num_pins="1"/>
            <metadata>
              <meta name="type">bel</meta>
              <meta name="subtype">output</meta>
            </metadata>
          </pb_type>
          <interconnect>
            <direct input="IOB33M.O" name="O" output="OUTBUF_VPR.IN"/>
            <direct input="IOB33M.T" name="T" output="OUTBUF_VPR.TRI"/>
            <direct input="OUTBUF_VPR.OUT" name="to_outpad" output="outpad.outpad">
              <pack_pattern in_port="OUTBUF_VPR.OUT" name="OUTBUF_to_outpad" out_port="outpad.outpad"/>
            </direct>
          </interconnect>
        </mode>
```

alessandrocomodi:
By removing the `OBUF` mode, packing completes succsessfully

litghost:
> @litghost I think I found the reason why it fails:

Sorry, I don't follow your explanation, but then again my head hasn't been in this problem.  Can you explain a different way?

alessandrocomodi:
So, for the second instance of the OSERDESE2, the packer selects the OBUF mode for the IOB33M, which implements the OUTBUF_VPR model, which is actually strange, because the first OSERDESE2 used the NO_BUF mode of the IOB33.

The packer tries to reach the `outpad` input pin, but the problem is that, for the `OBUF` mode, the `.outpad` input is connected to OUTBUF_VPR.OUT and not to `IOB33M.O`, causing the packer to fail in finding a valid route from the `OSERDESE2.OQ -> OLOGICE3 -> IOB33M -> OBUF.O` pins.

litghost:
But isn't that just a bug in VPR then?  Why didn't it try the NO_OBUF mode?

acomodi:
@litghost yes indeed, I am looking into that to debug what is happening

acomodi:
Turns out that applying the correct `pack_pattern` the packer can succeed. There was no bug in VPR. The only problem now is to find the most suitable way to create the pack_pattern chain between the `NO_OBUF` pb_type of the `IOB33[MS]?` and the OSERDESE2 output pin

kgugala:
> > @litghost where to add the pack pattern though? I cannot do that directly in the OLOGICE3 primitive as it will be included in the three different IOPAD pb_types, each of one containing the corresponding IOB33 pb_types with their OBUFs.
> > The problem is that the OLOGICE3 primitive should not be aware in which pb_type it will be included, I guess.
> 
> I completely agree. This is a common problem I've had when defining pack patterns, where what I really want to say to VPR is "make a molecule between this atom and every downstream atom". @kgugala Can we use V2X here to make defining this kind of pb_type agnostic pack pattern easier? Basically an annotation on OLOGICE3 OQ that is "make a pack pattern for each variant"?
> 
> This would come in handle in the SLICEL/SLICEM too.

@litghost @acomodi 
There is no such functionality right now, but I don't think it will be hard to implement

acomodi:
I have temporary overcome the `pack_pattern` issue with explicitly instantiating the `OBUF` in the design. I also think that a `pack_pattern` should not be a long term solution as the OSERDESE2 output should not be linked to a specific iob33 mode.
The fact is that, IMO, the packer should be able to find a correct solution if it exists (in this case placing the molecule in the `NO_OBUF` mode instead of the `OBUF` one, even without a `pack_pattern`.

Anyway, with the packing problem solved I have incurred in the following unroutable situations using the latest master+wip release (https://github.com/SymbiFlow/vtr-verilog-to-routing/releases/tag/v8.0.0_rc1_1140_g596caca0d):

```
# Routing
  RR Graph Nodes: 1287364
  RR Graph Edges: 7937856
Confirming router algorithm: TIMING_DRIVEN.
Warning 103: No routing path for connection to sink_rr 903203, retrying with full device bounding box
Warning 104: No routing path for connection to sink_rr 490069, retrying with full device bounding box
Warning 105: No routing path for connection to sink_rr 1121972, retrying with full device bounding box
Cannot route from BLK-TL-SLICEM.AQ[0] (RR node: 1121985 type: SOURCE location: (137,129) class: 38 capacity: 1) to BLK-TL-SLICEM.CX[0] (RR node: 1121972 type: SINK location: (137,129) class: 25 capacity: 1) -- no possible path
Failed to route connection from '$techmap7003$auto$simplemap.cc:420:simplemap_dff$361.cesr_mux.ce' to '$techmap7003$auto$simplemap.cc:420:simplemap_dff$361.cesr_mux.ce' for net 'oserdes_test0.comparator.sreg[21]' (#283)
Routing failed.
# Routing took 15.54 seconds (max_rss 4261.7 MiB, delta_rss +0.0 MiB)
Circuit is unroutable with a channel width factor of 500.
VPR failed to implement circuit
The entire flow of VPR took 33.55 seconds (max_rss 4261.7 MiB)

```

This unroutable situation affects the SLICE[LM] tiles and I am wondering if this is an issue with the design or the rr-graph look-ahead.

Here is the piece of code of the design that produces the failure:
```
// Shift register for reference data, shift strobe generator.
reg  [31:0] sreg;
reg  [ 4:0] sreg_sel;
wire        sreg_dat;
reg         sreg_sh;

always @(posedge CLK)
    sreg <= (sreg << 1) | i_dat_ref;

always @(posedge CLK)
    if (RST)
        sreg_sel <= 0;
    else if(sreg_sh)
        sreg_sel <= sreg_sel + 1;

assign sreg_dat = sreg[sreg_sel];

```

acomodi:
I have double-checked and routing succeeds when using the conda VtR and the search locations.

acomodi:
A similar routing error shows up for the `counter_basys3_x1y0_test`:
```
# Routing
  RR Graph Nodes: 1287364
  RR Graph Edges: 7937856
Confirming router algorithm: TIMING_DRIVEN.
Warning 101: No routing path for connection to sink_rr 957419, retrying with full device bounding box
Warning 102: No routing path for connection to sink_rr 247128, retrying with full device bounding box
Warning 103: No routing path for connection to sink_rr 1021785, retrying with full device bounding box
Cannot route from BLK-TL-SLICEL.CMUX[0] (RR node: 1021817 type: SOURCE location: (104,145) class: 38 capacity: 1) to BLK-TL-SLICEL.AX[0] (RR node: 1021785 type: SINK location: (104,145) class: 6 capacity: 1) -- no possible path
Failed to route connection from '$auto$alumacc.cc:485:replace_alu$29.slice[0].carry4_1st_full' to '$auto$alumacc.cc:485:replace_alu$29.slice[0].carry4_1st_full' for net '$abc$330$auto$alumacc.cc:485:replace_alu$29.slice[0].carry4_1st_full.O[2]' (#47)
Routing failed.
# Routing took 0.65 seconds (max_rss 4259.3 MiB, delta_rss +0.0 MiB)
Circuit is unroutable with a channel width factor of 500.
VPR failed to implement circuit
The entire flow of VPR took 22.20 seconds (max_rss 4259.3 MiB)
xc7/tests/counter/CMakeFiles/counter_basys3_x1y0_bin.dir/build.make:94: recipe for target 'xc7/tests/counter/counter_basys3_x1y0/artix7-xc7a50t-roi-virt-xc7a50t-test/top.route' failed

```

The problem is that for the [ROI-breakout](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1105) architecture, the same test did not fail (otherwise CI would have turned red). This suggests me that there could be a lookahead problem when using `GRAPH_LIMIT` and constraining the architecture to use the bottom row clock-regions only. I will try to expand the test to the whole 50T and see if the routing issue shows up again for the IOSERDES

litghost:
Are you integrated at a point that include this commit: https://github.com/SymbiFlow/symbiflow-arch-defs/commit/8aed0b4b8605cc58fd6b6cb2f448e0678f9313a4 ?

acomodi:
@litghost @mkurc-ant The following is the `fasm2frames` error I get when merging the [BUFR PR](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1103) with this one:

```
prjxray.fasm_assembler.FasmLookupError: Segment DB LIOI3, key LIOI3.OLOGIC_Y0.OSERDES.IN_USE not found from line 'LIOI3_X0Y23.OLOGIC_Y0.OSERDES.IN_USE'
Segment DB LIOI3, key LIOI3.OLOGIC_Y0.OSERDES.DATA_RATE_OQ.SDR not found from line 'LIOI3_X0Y23.OLOGIC_Y0.OSERDES.DATA_RATE_OQ.SDR = 1'b1'
Segment DB LIOI3, key LIOI3.OLOGIC_Y0.OSERDES.DATA_RATE_TQ.SDR not found from line 'LIOI3_X0Y23.OLOGIC_Y0.OSERDES.DATA_RATE_TQ.SDR = 1'b1'
Segment DB LIOI3, key LIOI3.OLOGIC_Y0.OSERDES.DATA_WIDTH.W8 not found from line 'LIOI3_X0Y23.OLOGIC_Y0.OSERDES.DATA_WIDTH.W8 = 1'b1'
Segment DB LIOI3, key LIOI3.OLOGIC_Y1.OSERDES.DATA_RATE_TQ.BUF not found from line 'LIOI3_X0Y3.OLOGIC_Y1.OSERDES.DATA_RATE_TQ.BUF'
Segment DB LIOI3_TBYTESRC, key LIOI3_TBYTESRC.OLOGIC_Y1.OSERDES.DATA_RATE_TQ.BUF not found from line 'LIOI3_TBYTESRC_X0Y43.OLOGIC_Y1.OSERDES.DATA_RATE_TQ.BUF'
Segment DB HCLK_IOI3, key HCLK_IOI3.BUFR_X0.IN_USE not found from line 'HCLK_IOI3_X1Y26.BUFR_X0.IN_USE'
Segment DB HCLK_IOI3, key HCLK_IOI3.BUFR_X0.BUFR_DIVIDE.D8 not found from line 'HCLK_IOI3_X1Y26.BUFR_X0.BUFR_DIVIDE.D8 = 1'b1'
Segment DB HCLK_IOI3, key HCLK_IOI3.HCLK_IOI_BUFR3_CE.HCLK_RCLK_DIV_CE3 not found from line 'HCLK_IOI3_X1Y26.HCLK_IOI_BUFR3_CE.HCLK_RCLK_DIV_CE3'
Segment DB HCLK_IOI3, key HCLK_IOI3.HCLK_IOI_BUFR3_CLR.HCLK_RCLK_DIV_CLR3 not found from line 'HCLK_IOI3_X1Y26.HCLK_IOI_BUFR3_CLR.HCLK_RCLK_DIV_CLR3'
Segment DB HCLK_IOI3, key HCLK_IOI3.HCLK_IOI_RCLK_BEFORE_DIV3.HCLK_IOI_RCLK_IMUX3 not found from line 'HCLK_IOI3_X1Y26.HCLK_IOI_RCLK_BEFORE_DIV3.HCLK_IOI_RCLK_IMUX3'
Segment DB HCLK_IOI3, key HCLK_IOI3.HCLK_IOI_RCLK2RCLK3.HCLK_IOI_RCLK_OUT3 not found from line 'HCLK_IOI3_X1Y26.HCLK_IOI_RCLK2RCLK3.HCLK_IOI_RCLK_OUT3'
Segment DB HCLK_IOI3, key HCLK_IOI3.HCLK_IOI_CK_BUFRCLK3.HCLK_IOI_RCLK2RCLK3 not found from line 'HCLK_IOI3_X1Y26.HCLK_IOI_CK_BUFRCLK3.HCLK_IOI_RCLK2RCLK3'
Segment DB HCLK_IOI3, key HCLK_IOI3.HCLK_IOI_RCLK2IO3.HCLK_IOI_CK_BUFRCLK3 not found from line 'HCLK_IOI3_X1Y26.HCLK_IOI_RCLK2IO3.HCLK_IOI_CK_BUFRCLK3'
xc7/tests/serdes/CMakeFiles/oserdes_basys3_x1y0_bin.dir/build.make:66: recipe for target 'xc7/tests/serdes/oserdes_basys3_x1y0/artix7-xc7a50t-roi-virt-xc7a50t-test/top.frames' failed

```

- The `OSERDES` issue should be fixed as I haven't updated to latest db with the fix for the OSERDES names;
- I think that `BUFR_X0` should be `BUFR_Y0` so it is a matter of changing the fasm feature in the primitive;
- The pips that are not found are actually missing from the DB. I think this is because the BUFR is not actually needed in the LiteX design, therefore we did not focus on those pips (I guess). An update to the fuzzer should be required IMO.

litghost:
> * I think that `BUFR_X0` should be `BUFR_Y0` so it is a matter of changing the fasm feature in the primitive;

The Yx is correct, as the BURF's are stacked in Y not X.  So the pb_type is wrong in this case.

litghost:
> * The pips that are not found are actually missing from the DB. I think this is because the BUFR is not actually needed in the LiteX design, therefore we did not focus on those pips (I guess). An update to the fuzzer should be required IMO.

I think some of those are ppips, which @mkurc-ant added, and a new db update will pull them in.  One that I think is not a ppips is :
```
Segment DB HCLK_IOI3, key HCLK_IOI3.HCLK_IOI_RCLK_BEFORE_DIV3.HCLK_IOI_RCLK_IMUX3 not found from line 'HCLK_IOI3_X1Y26.HCLK_IOI_RCLK_BEFORE_DIV3.HCLK_IOI_RCLK_IMUX3'
```

acomodi:
@litghost Ya, I have double-checked and the only real pip is the one pointed by you. I'll modify the fuzzer to include it as well

litghost:
> I think some of those are ppips, which @mkurc-ant added, and a new db update will pull them in. One that I think is not a ppips is :

Latest db didn't get those ppips, so that also might need to be tweaked.

mkurc-ant:
@litghost @acomodi BTW: I'm not sure whether the site ordering (thus their prefixes) are consistent with the prjxray database. Probably they are not.

acomodi:
OSERDES is working on HW, at least the one we need (`DDR` and `DATA_WIDTH=8`). The test need some tweaks to allow testing all the DATA_WIDTHs as well as DDR and SDR data rates.

To do so I have merged in my local branch the following PRs:
- [ROI breakout](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1105)
- [PLL_ADV](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1116)

I have modified the [oserdes minitest](https://github.com/SymbiFlow/prjxray/tree/master/minitests/oserdes) to allow dividing the clock with the PLL and using the PMODs instead the IOBUF to drive the i/o signals of the OSERDES.

As soon as the PLL support and the ROI-breakout land, OSERDESes can be supported on master.

For the time being I could focus also on the support of ISERDESEs which should be similar.

acomodi:
After some testing I have realized that the OSERDES is not correctly functioning. The problem is the CLK sources of the OSERDES which must respect some rules (coming from the same PLL, or being driven by a mixed combination of BUFIO and BUFR), which the test isn't currently taking into account.

The problem though arises when I try to connect the PLL clk/clkdivide input and outputs to BUFG tiles. What happens is that VPR fails for an unroutable situation:
```

# Routing
  RR Graph Nodes: 3539203
  RR Graph Edges: 22966468
Confirming router algorithm: TIMING_DRIVEN.
Warning 103: No routing path for connection to sink_rr 800473, retrying with full device bounding box
Warning 104: No routing path for connection to sink_rr 2023565, retrying with full device bounding box
Warning 105: No routing path for connection to sink_rr 3507863, retrying with full device bounding box
Cannot route from BLK-TL-CLK_BUFG_TOP_R.CLK_BUFG_BUFGCTRL0_O[0] (RR node: 2023507 type: SOURCE location: (79,106) class: 0 capacity: 1) to BLK-TL-CMT_TOP_L_UPPER_T.CMT_TOP_R_UPPER_T_PLLE2_CLKIN1[0] (RR node: 3507863 type: SINK location: (140,62) class: 150 capacity: 1) -- no possible path
Failed to route connection from 'bufg_clk' to 'pll' for net 'CLK' (#4)
Routing failed.
# Routing took 114.17 seconds (max_rss 11919.5 MiB, delta_rss +0.0 MiB)
Circuit is unroutable with a channel width factor of 500.
VPR failed to implement circuit

```

I believe that some segments definitions are still missing, as the `rr_graph_walk` tool correctly finds the path

litghost:
> I believe that some segments definitions are still missing, as the `rr_graph_walk` tool correctly finds the path

You can run `route_diag` on that src/sink rr pair, with debug rr sink enabled and compiled with `VTR_ENABLE_DEBUG_LOGGING=ON`, and it will print which rr node generated a bad lookahead.  That should point to which wire name needs a new segment.

acomodi:
> > I believe that some segments definitions are still missing, as the `rr_graph_walk` tool correctly finds the path
> 
> You can run `route_diag` on that src/sink rr pair, with debug rr sink enabled and compiled with `VTR_ENABLE_DEBUG_LOGGING=ON`, and it will print which rr node generated a bad lookahead. That should point to which wire name needs a new segment.

@litghost What do you mean by `debug rr_sink enabled`? Because I have ran route_diag with the following command:
```
/home/alessandro/projects/symbiflow/tmp/vtr-symbiflow/build/utils/route_diag/route_diag /home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/xc7/archs/artix7/devices/xc7a50t-virt/arch.timing.xml /home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/xc7/tests/serdes/oserdes_basys3_x1y0/artix7-xc7a50t-roi-virt-xc7a50t-test/top.eblif --device xc7a50t-test --read_rr_graph /home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/xc7/archs/artix7/devices/rr_graph_xc7a50t_test.rr_graph.real.xml --min_route_chan_width_hint 100 --max_router_iterations 500 --routing_failure_predictor off --router_high_fanout_threshold -1 --constant_net_method route --route_chan_width 500 --max_criticality 0.9 --clock_modeling route --place_delay_model delta_override --router_lookahead connection_box_map --clustering_pin_feasibility_filter off --disable_check_route on --strict_checks off --allow_dangling_combinational_nodes on --disable_errors check_unbuffered_edges:check_route --congested_routing_iteration_threshold 0.8 --timing_tradeoff 0.6 --astar_fac 1.0 --bb_factor 10 --suppress_warnings /home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/xc7/tests/serdes/oserdes_basys3_x1y0/artix7-xc7a50t-roi-virt-xc7a50t-test/noisy_warnings.log,sum_pin_class:check_unbuffered_edges:load_rr_indexed_data_T_values:check_rr_node:trans_per_R --read_router_lookahead /home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/xc7/archs/artix7/devices/rr_graph_xc7a50t_test.lookahead.bin --read_placement_delay_lookup /home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/xc7/archs/artix7/devices/rr_graph_xc7a50t_test.place_delay.bin --sink_rr_node 3507863 --source_rr_node 2023507
```

and got the following output:
```
Warning 100: No routing path for connection to sink_rr 3507863, retrying with full device bounding box
Cannot route from BLK-TL-CLK_BUFG_TOP_R.CLK_BUFG_BUFGCTRL0_O[0] (RR node: 2023507 type: SOURCE location: (79,106) class: 0 capacity: 1) to BLK-TL-CMT_TOP_L_UPPER_T.CMT_TOP_R_UPPER_T_PLLE2_CLKIN1[0] (RR node: 3507863 type: SINK location: (140,62) class: 150 capacity: 1) -- no possible rout...
```

I compiled with `VTR_ENABLE_DEBUG_LOGGIN=ON`

acomodi:
Nevermind, I found the option. It was not in the documentation.

acomodi:
The wire is the following:
```
rr inode: 1424402
graph_node_pkey inode: 3196603
NodeType: NodeType.CHANY
Wires (1):
  CLK_BUFG_TOP_R_X60Y53/CLK_BUFG_BUFGCTRL0_O

```

mithro:
> Nevermind, I found the option. It was not in the documentation.

If this is missing from the vtr documentation, please do log a bug about it being missing.

acomodi:
> > Nevermind, I found the option. It was not in the documentation.
> 
> If this is missing from the vtr documentation, please do log a bug about it being missing.

Ok. Also, As `route_diag` is in the `utils` directory it would come in hand to add a new section to the docs related to it (as for the FASM utility https://vtr-verilog-to-routing.readthedocs.io/en/latest/utils/index.html). I'll open a bug upstream to have it integrated in the docs.

litghost:
>       Expanding to node 1424402 (RR node: 1424402 type: CHANY location: (79,93) <-> (79,121) track: 50 len: 28 longline: 0 seg_type: OUTPINFEED dir: BI_DIR capacity: 1)
>         Lookahead from OUTPINFEED (RR node: 1424402 type: CHANY location: (79,93) <-> (79,121) track: 50 len: 28 longline: 0 seg_type: OUTPINFEED dir: BI_DIR capacity: 1) to BLK-TL-CMT_TOP_L_UPPER_T.CMT_TOP_R_UPPER_T_PLLE2_CLKIN1[0] (RR node: 3507863 type: SINK location: (140,62) class: 150 capacity: 1) is non-finite, expected_cost = -nan, to->R_upstream = 0.000000

So the bug in this case is that `CLK_BUFG_TOP_R_X60Y53/CLK_BUFG_BUFGCTRL0_O` should not be marked as `OUTPINFEED`, but as something else.

acomodi:
With the new segments the router completed sucessfully.

There are still some other segments to be added (e.g. BUFG_I[01]) as the router fails to drive a clock divided in logic and routed to the BUFG input.

Tomorrow I'll test on HW the current build.

xc7/utils/prjxray_edge_library.py
```diff
@@ -1647,6 +1675,24 @@ def annotate_pin_feeds(conn, ccio_sites):
     cur.execute("SELECT pkey FROM segment WHERE name = ?", ("OUTPINFEED", ))
     outpinfeed_pkey = cur.fetchone()[0]
 
+    cur.execute(
```
litghost: Should probably make a simply name -> pkey map, and convert these into a loop to build the map.

xc7/utils/prjxray_edge_library.py
```diff
@@ -31,6 +31,34 @@ def get_node_type(conn, graph_node_pkey):
     return c.fetchone()[0]
 
 
+def get_pins(conn, site_type, site_pin):
+    """ Returns a set of the pin graph_nodes related to the input site type and pin names."""
+
+    c = conn.cursor()
+    c.execute(
+        """
+WITH pll_opins(wire_in_tile_pkey) AS (
```
litghost: This inner table name is now too specific?

acomodi:
OSERDES still do not work on HW. I have added fasm2bel support for it to debug the bitstream, but at the `bit_v` step the following failure shows up, related to an orphan sink:
```
Generating oserdes_basys3_x1y0/artix7-xc7a50t-roi-virt-xc7a50t-test/top_bit.v
  File "/home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/env/conda/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/build/env/conda/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/xc7/fasm2bels/__main__.py", line 4, in <module>
    main()
  File "/home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/xc7/fasm2bels/fasm2bels.py", line 296, in main
    top.make_routes(allow_orphan_sinks=args.allow_orphan_sinks)
  File "/home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/xc7/fasm2bels/verilog_modeling.py", line 1380, in make_routes
    net_map=self.net_map,
  File "/home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/xc7/fasm2bels/make_routes.py", line 700, in make_routes
    allow_orphan_sinks=allow_orphan_sinks
  File "/home/alessandro/projects/symbiflow/tmp/symbiflow-arch-defs/xc7/fasm2bels/make_routes.py", line 610, in expand_sink
    assert False, (sink_node_pkey, tile_name, wire_name, sink_wire_pkey)
AssertionError: (2068119, 'RIOI3_TBYTESRC_X43Y93', 'IOI_OLOGIC0_CLKDIV', 5920677)

```

The `IOI_OLOGIC0_CLKDIV` sink seems to have no source.

I am thinking of also make the OSERDES test similar to the `ff_sr_ce` one, that produces different verilog sources for each possible configurations of the OSERDES.
This because I have experienced troubles for the router to route all the different CLKDIVs headed to the various OSERDES (many times the router hit a highly congested situation, reaching 400+ iterations).

acomodi:
I have create another device that uses only the bottom row of the xc7 to avoid getting `impossible placement` due to https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1153 when running Vivado targets.

I still get an issue when building using vivado and the output from fasm2bels. The problem is that the Router fails to route the constant nets (GLOBAL_LOGIC_0) due to pin conflicts.

The pin are related to the OSERDES inputs whenever they are set to 0 (therefore they want to connect to HARD0, but I suppose they are conflicting with routes coming from HARD1). The conflict happen for the pins that have an inverter (e.g D1-D8 and T1-T4).

litghost:
> The pin are related to the OSERDES inputs whenever they are set to 0 (therefore they want to connect to HARD0, but I suppose they are conflicting with routes coming from HARD1). The conflict happen for the pins that have an inverter (e.g D1-D8 and T1-T4).

Ya, you need leverage the inverter if the inputs are constant.  This should be done at the techmap-ing level, see https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/techmap/cells_map.v#L1958-L2005

acomodi:
All clear, I know what the issue is then, I forgot to add the INV_D1 ^ D1, etc, to the input pins as well. This will solve it, thanks

litghost:
> I have create another device that uses only the bottom row of the xc7 to avoid getting `impossible placement` due to #1153 when running Vivado targets.

We have some guidance from Vauhgn on the placement issue here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1045#issuecomment-553949121

If you have OSERDES working except for the placement issues, maybe you can take a look at what is the smallest change suitable to solving our placement issues around the PLL?

litghost:
If the issue is purely about clock routing and not placement, then maybe investigate how https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/928 can help us.  We likely need to add routing graph annotations.  I have a message from Mustafa indicating he will be update the code today ish (Nov 15th 2019) addressing some review comments.

acomodi:
@litghost So, the issues are actually two:

1. Placement of PLL and BUFG: If the PLL is instantiated in the TOP row (Y1 and Y2 clock regions) and BUFGs in the BOT row (Y0 region) routing will fail (or at least route through INT tiles) as there is no route that leads from the PLL input/outputs to the BUFG using the dedicated clock tree.

2. I have seen the clock routing issue for which a clock was routed from a BUFHCE belonging to Y1 region to a bunch of slices in the Y0 region. This happened only in the OSERDES test, but I don't want to exclude that it may happen for other complex tests.

For the first problem I will look into Vaughn suggestions, but a very quick workaround is to add only the bottom row CLK_BUFG tiles and PLLs.

For the second problem, which is IMO more urgent, we definetly need to integrate Mustafa's changes. As soon as review comments are addressed I'll try to integrate its branch within master+wip and see if that solves our issue.

litghost:
> For the second problem, which is IMO more urgent, we definetly need to integrate Mustafa's changes. As soon as review comments are addressed I'll try to integrate its branch within master+wip and see if that solves our issue.

I believe Mustafa's changes require some annotations in the graph.  I don't believe the rr graph writer or reader support these annotations, but if you look at the relevant code, you may be able to figure out what annotations will eventually be required, and start the work to prepare to emit them.

acomodi:
> I believe Mustafa's changes require some annotations in the graph. I don't believe the rr graph writer or reader support these annotations, but if you look at the relevant code, you may be able to figure out what annotations will eventually be required, and start the work to prepare to emit them.

Sure, I will look into that than

acomodi:
By runnng `fasm_diff` this is the failure I get:

```
 LIOI3_X0Y25.IDELAY_Y0.IDELAY_TYPE_FIXED
 LIOI3_X0Y25.IDELAY_Y1.IDELAY_TYPE_FIXED
 LIOI3_X0Y25.ILOGIC_Y1.ZINV_D
+LIOI3_X0Y25.IOI_OCLKM_0.IOI_LEAF_GCLK3
 LIOI3_X0Y25.IOI_OCLK_0.IOI_LEAF_GCLK3
 LIOI3_X0Y25.IOI_OLOGIC0_CLKDIV.IOI_LEAF_GCLK4
 LIOI3_X0Y25.OLOGIC_Y0.IS_D1_INVERTED
@@ -4070,6 +4071,7 @@
 LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_RATE_OQ.DDR
 LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_RATE_TQ.SDR
 LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_WIDTH.DDR.W6_8
+LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_WIDTH.SDR.W2_4_5_6
 LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_WIDTH.W8
 LIOI3_X0Y25.OLOGIC_Y0.OSERDES.IN_USE
 LIOI3_X0Y25.OLOGIC_Y0.OSERDES.SRTYPE.SYNC
```

When using OSERDES Vivado routes also through the `OCLK_M` pip, which does not happen using VPR. I think that this can be solved by modifying the `check_features` function during `prjxray_routing_import`. 

The other difference (`LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_WIDTH.SDR.W2_4_5_6`) is kinda weird as the selected DATA_RATE is `DDR`, I believe that this feature could be related to something more than just DATA_WIDTH.

xc7/techmap/cells_map.v
```diff
@@ -2703,8 +2931,7 @@ output [15:0] DO
   .CLKOUT2_CLKOUT1_OUTPUT_ENABLE(_TECHMAP_CONSTMSK_CLKOUT2_ == 1'b0),
   .CLKOUT3_CLKOUT1_OUTPUT_ENABLE(_TECHMAP_CONSTMSK_CLKOUT3_ == 1'b0),
   .CLKOUT4_CLKOUT1_OUTPUT_ENABLE(_TECHMAP_CONSTMSK_CLKOUT4_ == 1'b0),
-  .CLKOUT5_CLKOUT1_OUTPUT_ENABLE(_TECHMAP_CONSTMSK_CLKOUT5_ == 1'b0)
-
+  .CLKOUT5_CLKOUT1_OUTPUT_ENABLE(_TECHMAP_CONSTMSK_CLKOUT5_ == 1'b0),
```
litghost: I don't believe trailing colons is legal?

xc7/utils/prjxray_routing_import.py
```diff
@@ -228,6 +229,14 @@ def check_feature(feature):
     if rebuf_key in REBUF_SOURCES:
         return ' '.join([feature] + REBUF_NODES[REBUF_SOURCES[rebuf_key]])
 
+    m = IOI_OCLK.fullmatch(feature_path[1])
```
litghost: I think this is better modelled as a change in the techmap, because I believe this routing is a function of the mode?

acomodi: Ok, I will try to implement as suggested. OCLKM should be directed to the CLKB port, so actually it should be possible to add that to the pb_type as well and to the techmap.

acomodi:
> Overall looking really good. I have a different suggestion for the OCLKM solution, and the CI is currently red?

So, CI is red for the `diff_fasm` target. There are two reasons:

1. these features behave in an apparently different way then we thought (they seem not to be strictly related to the selected DATA_RATE):
```
LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_WIDTH.DDR.W6_8
LIOI3_X0Y25.OLOGIC_Y0.OSERDES.DATA_WIDTH.SDR.W2_4_5_6
```

2. The prjxray-db version used in master still has wrong ppips defined, as soon as that gets bumped this will be solved: 
https://github.com/SymbiFlow/prjxray-db/blob/d97fb8bd886ccd245a99bf504f45d4579ac96d85/artix7/ppips_hclk_ioi3.db#L5-L16 which generates the following diff:
```
@@ -448,6 +448,8 @@
 HCLK_IOI3_X1Y26.BUFR_Y1.BUFR_DIVIDE.BYPASS
 HCLK_IOI3_X1Y26.BUFR_Y2.BUFR_DIVIDE.BYPASS
 HCLK_IOI3_X1Y26.BUFR_Y3.BUFR_DIVIDE.BYPASS
+HCLK_IOI3_X1Y26.HCLK_IOI_CK_IGCLK11.HCLK_IOI_CK_BUFHCLK11
+HCLK_IOI3_X1Y26.HCLK_IOI_CK_IGCLK2.HCLK_IOI_CK_BUFHCLK2
 HCLK_IOI3_X1Y26.HCLK_IOI_LEAF_GCLK_TOP4.HCLK_IOI_CK_IGCLK11
 HCLK_IOI3_X1Y26.HCLK_IOI_LEAF_GCLK_TOP5.HCLK_IOI_CK_IGCLK2
```

litghost:
> 2. The prjxray-db version used in master still has wrong ppips defined, as soon as that gets bumped this will be solved:

Bump this yourself in the PR?

litghost:
> 1. these features behave in an apparently different way then we thought (they seem not to be strictly related to the selected DATA_RATE):

I believe they function like:
 - SDR 6 -> set `DATA_WIDTH.SDR.W2_4_5_6`
 - DDR 6 -> set `DATA_WIDTH.SDR.W2_4_5_6` and `DATA_WIDTH.DDR.W6_8`

litghost:
> > 1. these features behave in an apparently different way then we thought (they seem not to be strictly related to the selected DATA_RATE):
> 
> I believe they function like:
> 
> * SDR 6 -> set `DATA_WIDTH.SDR.W2_4_5_6`
> * DDR 6 -> set `DATA_WIDTH.SDR.W2_4_5_6` and `DATA_WIDTH.DDR.W6_8`

This doesn't jive with https://github.com/SymbiFlow/prjxray/blob/master/fuzzers/036-iob-ologic/generate.py#L28-L41

Anyways, we should fix this.

xc7/techmap/cells_map.v
acomodi: Typo, done.

xc7/utils/prjxray_routing_import.py
acomodi: By trying to add the CLKB port as well, I get a fasm lookup error. The CLK signal is taken from the OCLK_0 port instead of the OCLKM_0. 

```
prjxray.fasm_assembler.FasmLookupError: Segment DB LIOI3, key IOI3.IOI_OLOGIC0_CLKB.IOI_OCLK_0 not found from line IOI3_X0Y25.IOI_OLOGIC0_CLKB.IOI_OCLK_0'
```

Being OCLKM a pip, I am unsure on how to do this in the techmap.
Moreover, OCLKM is dedicated to only route the clock signal that goes through the inverter, both for ILOGIC and OLOGIC. I do believe, that, whenever a CLK is used in the I/OLOGIC, the same signal needs to reach the inverted CLK port which will give "stability" to the primitive. In fact, I have seen that, without the OCLKM pip enabled, the OSERDES was alive, but was highly unstable. IMO it is safe to allow `check_features` to always add the OCLKM pip feature anytime the OCLK is enabled.

litghost:
Please rebase onto master.  I merged the IOBUF PR, and it generated conflicts.

xc7/utils/prjxray_routing_import.py
litghost: Sounds good to me.  We can always change it later.

acomodi:
> LGTM, rebase/merge with master, and then merge once green

Done, let's wait for CI

PR #1110, opened 2019-10-24
xc7: fix dram timings and models

kgugala:
This PR adds timings injection to DRAM pb_types. Also, it modifies the DRAM models removing combinational connection of write port signals to output (the docs says all writes are sequential/reads are combinational)

### DIFF
--- a/xc7/bels.json
+++ b/xc7/bels.json
@@ -92,43 +92,43 @@
         },
         "DPRAM32_O6": {
             "A_DRAM":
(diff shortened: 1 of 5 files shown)


mithro:
@kgugala I'm still *(personally)* very confused with how the combinational and sequential pathways interact on things like RAM blocks and how they should be described to VtR. We should improve the VtR documentation with a really good example showing this type of async RAM block...

PR #1111, opened 2019-10-24
Add xc7 two router diagnostic tooling

litghost:
- annotate_vpr_log.py is designed to convert vpr logs (like the detailed
  routing diagnostic output) and change rr node ids to wire names.
- animate_router_pop.py is designed create a simple Vivado tcl script for
  animating the router search behavior for visualizing the VPR router behavior.


mithro:
Cool!

PR #1112, opened 2019-10-24
Refactor connection box tools

litghost:
- Adds tool for plotting connection box data
- Adds tool for convert connection box data in `.mat` file
- Refactors common logic into library.


utils/plot_connection_box.py
```diff
@@ -0,0 +1,114 @@
+import argparse
```
mithro: `#!/usr/bin/env python3`?

litghost: Done

litghost:
@HackerFoo / @mithro Please re-review.  I've refactored this PR a bit

PR #1116, opened 2019-10-28
Support for PLLE2_ADV

mkurc-ant:
This PR adds support for the PLLE2_ADV primitive to SymbiFlow.
It also bumps the prjxray db to `d97fb8b`

Already done:
- [x] Models of `CMT_TOP_[LR]_UPPER_T` tiles
- [x] Model of the `PLLE2_ADV` primitive
- [x] Techmap `PLLE2_ADV` -> `PLLE2_ADV_VPR` (and `PLLE2_BASE` -> `PLLE2_ADV`)
- [x] Calculation of PLL register values from given dividers, duty cycles etc.
- [x] A minitest

ToDo:
- [ ] ~~Figure out how to map `COMPENSATION` parameter to fasm features.~~ This requires more effort and probably more work on the fuzzer 032
- [ ] ~~Make `CMT_TOP_[LR]_UPPER_T` equivalent (sites are the same but routing not necessarily is)~~ Postponed for now.
- [x] Add real timings to the PLL model (timings vary across different COMPENSATION modes, for now the worst timings were chosen for the BUF_IN mode).

Issues:
- [x] The PLL site prefix in the db differs from the actual site name. This has been worked around by adding an option to the tile import script so that the prefix is not generated there.
- [x] ~~Figure out when to activate `*_CLOCK_ENABLE` features. It's not possible to detect whether a cell output is used from inside a techmap. Add those features to edges ?~~ It actually can be done.
- [ ] Yosys does not support floating point values in parameters, they are immediately converted to strings. Therefore It makes it impossible to specify eg. clock duty cycle as a real number. For now those have to be specified as integers but multiplied by 1000. https://github.com/YosysHQ/yosys/issues/1469
- [x] How to obtain values of `FILTREG1` and `LOCKREG3` (they seem to be const so they are hard-coded in the techmap)

This PR requires the database from https://github.com/SymbiFlow/prjxray/pull/1077 to work.


mkurc-ant:
VPR segfault... I suspect an issue with block usage statistics (It happened to me before)

mkurc-ant:
There are different timings for clock input -> output propagation for each COMPENSATION setting. Implementing this in VPR would require different models of the PLL for each of them. For now the worst timings were chosen.

xc7/utils/prjxray_tile_import.py
```diff
@@ -1447,6 +1449,13 @@ def main():
         '--filter_x', help="Filter imported sites by their x coordinate."
     )
 
+    parser.add_argument(
+        '--no_fasm_prefix',
+        action="store_true",
+        default=False,
```
litghost: Default is not required with `action="store_true"`, as `store_true` implies default=False

mkurc-ant: Fixed.

PR #1130, opened 2019-11-05
Add initial FASM2BELs README.

litghost:


xc7/fasm2bels/README.md
```diff
@@ -0,0 +1,60 @@
+FASM 2 BELs
+-----------
+
+fasm2bels is a tool designed to convert FASM generated from a bitstream back
```
mithro: fasm2bels is a tool designed to take a FASM file into Vivado.

It does this by generating a file describing the BEL connections (techmapped Verilog) and TCL commands for Vivado which lock the BEL placements.

This makes it possible to perform simulation and analysis of a FASM file inside Vivado.

In the absence of bugs, it is expected that after consuming the BEL connections and TCL constraints Vivado will produce a bitstream identical to the bitstream the FASM file would generate.

litghost:
> Just one comment on how I think we should change the initial description to better describe what this tool does.
> 
> The import fact is that this tool takes a FASM and allows you to read it into Vivado. The fact it generates some Verilog as part of this process is just an implementation detail.

I meant to incorporate your comment, but I had failed to commit the changes, done now.

PR #1134, opened 2019-11-07
Update VTR that contains latest sampling logic and reverts clock prop.

litghost:


litghost:
New lookahead is behaving really well!

2/5 50T lookahead computation time: 4826.888 sec -> 364.981 sec
picosoc route time : 1681.66 sec -> 416.95 sec
picosoc CPD:  18.076 ns -> 17.992 ns
murax route time: 574.93 sec -> 227.76 sec
murax CPD: 11.1028 ns -> 10.7015 ns

acomodi:
Great!

PR #1139, opened 2019-11-07
Change update_tools to avoid requiring force pushes all the time.

litghost:
The previous version of update_tools maintained a simplified history by force pushing overtop of `master+wip` and `master+wip-next`.  This results in github PRs not working correctly, and continuously orphaning trees.

This approach avoids the force push by making a revert commit to `master+wip`, and then merging with `master`.  This achieves the original goal of keeping the history sane, but avoids orphaning the old `master+wip` integration point.

See https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/319 as an example PR with this new methodology.

### DIFF
--- a/utils/update_tools/update_tools.py
+++ b/utils/update_tools/update_tools.py
@@ -104,6 +104,20 @@ def rebase_branch(g, branch):
         rebase_continue_rec(g)
 
 
+def revert_to_master(g, repo, remote):
+    """ Preserving history, revert {remote}/master+wip to {remote}/master
+    """
+    g.reset(['--hard', '{}/master'.format(remote)])
+    g.reset(['{}/master+wip'.format(remote)])
+    g.add(['.'])
+    g.commit(
+        "-sm \"Revert master+wip to master ({})\"".format(
+            repo.commit('{}/master'.format(remote)).hexsha
+        )
+    )
+    g.merge(['{}/master'.format(remote)])
+
+
 def main():
     parser = argparse.ArgumentParser()
 
@@ -117,6 +131,12 @@ def main():
 
     parser.add_argument(
         '--remote', help="Optional remote repository to use instead of origin"
+    )
+
+    parser.add_argument(
+        '--branch_name',
+        required=True,
+        help="Name of branch to push to github with new master+wip"
     )
 
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


litghost:
New example PR is https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/320

mithro:
`/home/travis/build/SymbiFlow/symbiflow-arch-defs/utils/simple_router_analysis.py:5:1: F401 'numpy as np' imported but unused`

mithro:
If @acomodi thinks it looks good, then I'm happy. 

Need to make sure the [SymbiFlow process for managing branches on non-SymbiFlow owned projects (and how to merge)](https://docs.google.com/document/d/1JVWkXxeSanAna8IRt0I49sG4aFS3Unlq693-HsQ0B0U/edit) document is updated.

utils/update_tools/update_tools.py
```diff
@@ -160,12 +165,21 @@ def main():
...
 
+    # Preserving history, revert {remote}/master+wip to {remote}/master
     g.reset(['--hard', '{}/master'.format(remote)])
+    g.reset(['{}/master+wip'.format(remote)])
+    g.add(['.'])
+    g.commit(
+        "-sm \"Revert master+wip to master ({})\"".format(
+            repo.commit('{}/master'.format(remote)).hexsha
+        )
+    )
```
acomodi: I think it is worth to encapsulate this in a function as this step should be repeated in case there are merge-conflicts and a sequential merge needs to be applied.

utils/update_tools/update_tools.py
```diff
@@ -119,6 +119,11 @@ def main():
         '--remote', help="Optional remote repository to use instead of origin"
     )
 
+    parser.add_argument(
+        'branch_name',
```
acomodi: Maybe worth to keep this consistent with the other arguments (e.g. `--branch_name <branch>`)

acomodi: In addition, I believe that, to avoid resetting master+wip by mistake, an assertion like `assert branch_name != 'master+wip'` could be useful.

utils/update_tools/update_tools.py
litghost: Sure

utils/update_tools/update_tools.py
litghost: Yep, same thought goes for `master`, done

litghost:
This strategy has a flaw, seeing if I can find a solution

litghost:
> This strategy has a flaw, seeing if I can find a solution

I've found a fix, please re-review.

litghost:
> why is the integration point needed?

Couple reasons, but they summarize to preventing git from doing nothing when something is needed.  In particular, the new branch needs to be merged back onto master to prevent the rebased merges from looking like they were reverted.  With the integration point on master, and then everything rebased onto there, the revert commit is no longer seen in the history, and everything works correctly.

PR #1142, opened 2019-11-08
Initial ROI breakout PR

litghost:
This PR enables ROI break on the basys3-x0y1 graph (which is now misnamed).

Remaining work to submit this PR:
 - [ ] Rename basys3-x0y1 to just the 50T underlying part
 - [ ] Remove search locations now that VPR no longer actually uses those arguments

### DIFF
--- a/xc7/tests/counter/counter_basys3.v
+++ b/xc7/tests/counter/counter_basys3.v
@@ -9,9 +9,12 @@ module top (
     localparam BITS = 4;
     localparam LOG2DELAY = 22;
 
+    wire bufg;
+    BUFG bufgctrl(.I(clk), .O(bufg));
+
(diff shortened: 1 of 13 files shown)


xc7/tests/counter/counter_basys3.v
```diff
@@ -9,9 +9,12 @@ module top (
     localparam BITS = 4;
     localparam LOG2DELAY = 22;
 
+    wire bufg;
```
mithro: I'd call this `clk_buf` or similar.

litghost:
I'll do  "Rename basys3-x0y1 to just the 50T underlying part" and "Remove search locations now that VPR no longer actually uses those arguments" in follow up PRs.

PR #1148, opened 2019-11-12
IOBUF support

mkurc-ant:
This PR adds support for IOBUF to SymbiFlow.

- [x] Minitest.
- [x] Blackbox / simulation model.
- [x] Architecture definition.
- [x] Techmap.
- [x] fasm2bels.
- [ ] Automatic inference of 3-state IOBUFs

With this PR the flow of SymbiFlow changes a bit. Upon sysnthesis the design is written to a JSON file. Then, the file is processed by the `split_inouts.py` utility which splits inout ports into separate inputs and outputs. After processing, the design is written to another JSON file which is then converted to EBLIF using Yosys.

This processing should be moved to a Yosys plugin in the future.

Each top-level inout port of the design is split into one input and one output port. Port names are given in the following way: `inout A` -> `input A_$inp` and `output A_$out`.


mkurc-ant:
I'm still having issues with making Yosys infer IOBUFs from assign statements though. For such a design:
```
inout wire pin

...

wire pin_i;
wire pin_o;
wire pin_t;

assign pin_i = pin;
assign pin = (pin_t == 1'b1) ? pin_o : 1'bz;
```

In some cases I end up with an `OBUFT`-like cell for the 3-state output and the input taken directly from the pin (not through any buffer). In other cases I have a properly connected `IOBUF` cell but after optimization its output gets disconnected (a bug?).

litghost:
> I'm still having issues with making Yosys infer IOBUFs from assign statements though. For such a design:

I know I suggested the iobufmap to begin with, but if we simply require the IOBUF explicitly, would your script do the right thing?

common/cmake/devices.cmake
```diff
@@ -12,7 +12,8 @@ function(DEFINE_ARCH)
   # ~~~
   # DEFINE_ARCH(
   #    ARCH <arch>
-  #    YOSYS_SCRIPT <yosys_script>
+  #    YOSYS_SYNTH_SCRIPT <yosys_script>
+  #    YOSYS_CONV_SCRIPT <yosys_script>
```
litghost: Need to update documentation on these arguments

common/cmake/devices.cmake
```diff
@@ -980,31 +985,62 @@ function(ADD_FPGA_TARGET)
...
       VERBATIM
     )
+
+    set(SPLIT_INOUTS ${symbiflow-arch-defs_SOURCE_DIR}/utils/split_inouts.py)
+
+    add_custom_command(
+      OUTPUT ${OUT_JSON}
+      DEPENDS ${OUT_JSON_SYNTH} ${QUIET_CMD} ${QUIET_CMD_TARGET} ${SPLIT_INOUTS} ${PYTHON3} ${PYTHON3_TARGET} simplejson
+      COMMAND
+        ${CMAKE_COMMAND} -E env PYTHONPATH=${PRJXRAY_DIR}:${symbiflow-arch-defs_SOURCE_DIR}/utils ${PYTHON3} ${SPLIT_INOUTS} -i ${OUT_JSON_SYNTH} -o ${OUT_JSON}
```
litghost: What is PRJXRAY_DIR required here?   It seems weird to have a 7-series specific thing in the (in theory) generic CMake function?

utils/split_inouts.py
```diff
@@ -0,0 +1,262 @@
...
+            return n0 + 1
+
+    # No gap was found, return max + 1.
+    return sorted_nets[-1] + 1
+
+
+# =============================================================================
+
+
+def find_and_split_inout_ports(design, module_name):
```
litghost: Docstring

utils/split_inouts.py
```diff
@@ -0,0 +1,262 @@
...
+        attrs = module["attributes"]
+        if "top" in attrs and attrs["top"] == 1:
+            return name
+
+    return None
+
+
+def get_nets(bits):
+    """
+    Returns a set of numbers corresponding to net indices.
```
litghost: This function is very weird.  Can you provide some doctest examples?

utils/split_inouts.py
```diff
@@ -0,0 +1,262 @@
...
+    """
+    Looks for the top-level module in the design. Returns its name.
+    """
+
+    for name, module in design["modules"].items():
+        attrs = module["attributes"]
+        if "top" in attrs and attrs["top"] == 1:
+            return name
+
+    return None
```
litghost: Should throw an exception rather than return None?

utils/split_inouts.py
```diff
@@ -0,0 +1,262 @@
...
+def get_nets(bits):
+    """
+    Returns a set of numbers corresponding to net indices.
+    """
+    return set([n for n in bits if isinstance(n, int)])
+
+
+def get_free_net(nets):
+    """
+    Given a set of used net indices, returns a new, free index.
```
litghost: This explaination is insufficient to understand what you are doing.  Needs more.

utils/split_inouts.py
```diff
@@ -0,0 +1,262 @@
...
+
+# =============================================================================
+
+
+def find_and_split_inout_ports(design, module_name):
+
+    # Get the module
+    module = design["modules"][module_name]
+
+    # Find indices of all used nets
```
litghost: I think you need to document the income JSON format here in the comment (or at least link to other documentation)

utils/split_inouts.py
```diff
@@ -0,0 +1,262 @@
...
+        netnames[name] = {
+            "hide_name": 0,
+            "bits": port["bits"],
+            "attributes": {}
+        }
+
+    return port_map, net_map
+
+
+def remap_connections(design, module_name, net_map):
```
litghost: Docstring

common/cmake/devices.cmake
litghost: Upon review of split_inouts, it doesn't seem to use PRJXRAY at all? So you should just be able to drop the path?

common/cmake/devices.cmake
mkurc-ant: Done.

common/cmake/devices.cmake
mkurc-ant: Right, It ended up here as a mistake. Removed.

utils/split_inouts.py
mkurc-ant: Done

utils/split_inouts.py
mkurc-ant: Added docstring and doctest.

utils/split_inouts.py
mkurc-ant: Added throwing an exception.

utils/split_inouts.py
mkurc-ant: Added docstring and doctest.

utils/split_inouts.py
mkurc-ant: Added.

utils/split_inouts.py
mkurc-ant: I linked up Yosys documentation to the docstring of the file. Couldn't find any "official" documentation of the JSON design format other than Yosys' manual for the `write_json` command.

Issue #1019, opened 2019-09-21
[xc7] Invalid usage of FAN[0-7] or BYP[0-7] nodes 


litghost:
Currently the router can use FAN[0-7] and BYP[0-7] nodes in a way that generates antennee nets.  In particular, if the FAN[0-7] and BYP[0-7] connect to a site pin, that node should only be used if the router is visiting the site pin.  Current the router thinks the connections between the FAN/BYP routing node to the IPIN is configurable, but this is not the case.

Solution is unclear, but likely involves making the FAN/BYP into IPIN nodes, and collapsing the nodes in between.

litghost:
Vaughn had a good suggestion to address this:

1. Because FAN_ALT -> FAN -> IPIN are non-configurable, update the VPR switch to be buffering non-configurable
2. Add a pass after net terminals are assigned to set the occupancy of all IPINs (and their associated non-configurable sets) to 1.  This will force the router to avoid these rr nodes.

PR #1151, opened 2019-11-13
WIP: Use "buffer" switch type for non-configurable switches (e.g. ppips).

litghost:
This is part 1 of #1019

### DIFF
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -307,9 +307,11 @@ CREATE TABLE graph_node(
   capacity INT,
   capacitance REAL,
   resistance REAL,
+  site_pin_graph_node_pkey INT,
   FOREIGN KEY(connection_box_wire_pkey) REFERENCES wire(pkey),
   FOREIGN KEY(track_pkey) REFERENCES track(pkey),
   FOREIGN KEY(node_pkey) REFERENCES node(pkey)
+  FOREIGN KEY(site_pin_graph_node_pkey) REFERENCES graph_node(pkey)
 );
 
 -- Table of wires.  This table is the complete list of all wires in the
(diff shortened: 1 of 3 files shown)


acomodi:
@litghost: I am trying to get this to work.

I have locally fixed VTR to accepct Cinternal on buffer switches, but it fails now with the following error:
```
Type: Routing
File: /home/alessandro/projects/symbiflow/vtr-symbiflow/vpr/src/route/check_rr_graph.cpp
Line: 546
Message: Non-configurable BUFFER type switch must have only one driver.  Actual fan-in was 2 (expected 1).
  Possible cause is complex block output pins connecting to:
    RR node: 12223 type: CHANX location: (7,8) <-> (9,8) track: 826 len: 2 longline: 0 seg_type: OUTPINFEED dir: BI_DIR capacity: 1
```

This because, apparently, `buffer` type switches must be unidirectional for VPR.

I am unsure whether this is something to be solved in this PR or in VPR.

litghost:
> 
> 
> @litghost: I am trying to get this to work.
> 
> I have locally fixed VTR to accepct Cinternal on buffer switches, but it fails now with the following error:
> 
> ```
> Type: Routing
> File: /home/alessandro/projects/symbiflow/vtr-symbiflow/vpr/src/route/check_rr_graph.cpp
> Line: 546
> Message: Non-configurable BUFFER type switch must have only one driver.  Actual fan-in was 2 (expected 1).
>   Possible cause is complex block output pins connecting to:
>     RR node: 12223 type: CHANX location: (7,8) <-> (9,8) track: 826 len: 2 longline: 0 seg_type: OUTPINFEED dir: BI_DIR capacity: 1
> ```
> 
> This because, apparently, `buffer` type switches must be unidirectional for VPR.
> 
> I am unsure whether this is something to be solved in this PR or in VPR.

Ah, I think I remember this now.  So this is a bug (I believe) in check rr graph.  Basically, the fan in edge in this case is part of a short.

So the structure looks like:
```
CHAN -> buffer -> CHAN <- short -> CHAN
```
Check route has no valid solution to this configuration.  A short requires bidirectional edges, and buffer requires fan in = 1.  Double check that I'm remembering this correctly, and file a bug upstream.

litghost:
@acomodi In the short term, just disable check rr route when generating the lookahead.  We already disable it during P&R to save time.

acomodi:
@litghost All clear, thanks.

PR #1154, opened 2019-11-14
Add PLL site equivilance.

### LINKED ISSUE #1131: Make CMT_TOP_[LR]_UPPER_T tiles equivalent
As of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1116 those two tile types are imported as separate but they contain identical PLLE2_ADV site. They should be made equivalent.

litghost:
This also needs the fix from https://github.com/SymbiFlow/prjxray/pull/1142

Fixes #1131


litghost:
@mkurc-ant Please review

PR #1158, opened 2019-11-19
XDC plugin integration

tmichalak:
This PR is supposed to make use of the xdc and fasm plugins which are implemented with in the following PR(https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/1)
What remains to be done is: 
 - adding to prjxray the generation of the `iobanks` entry that is expected to be present in the part's json
 - making sure the yosys-plugins conda package (https://github.com/SymbiFlow/conda-packages/pull/55) is built and used correctly


common/cmake/devices.cmake
```diff
@@ -963,6 +963,13 @@ function(ADD_FPGA_TARGET)
     if(NOT "${ADD_FPGA_TARGET_INPUT_IO_FILE}" STREQUAL "")
       add_file_target(FILE ${ADD_FPGA_TARGET_INPUT_IO_FILE})
     endif()
+    if(NOT "${ADD_FPGA_TARGET_INPUT_XDC_FILE}" STREQUAL "")
+      add_file_target(FILE ${ADD_FPGA_TARGET_INPUT_XDC_FILE})
+      get_file_location(INPUT_XDC_FILE ${ADD_FPGA_TARGET_INPUT_XDC_FILE})
+      if(NOT "${ADD_FPGA_TARGET_PART}" STREQUAL "")
+        set(PART_JSON ${PRJXRAY_DB_DIR}/${ARCH}/${ADD_FPGA_TARGET_PART}.json)
```
litghost: Get this from the `${BOARD}` target.  Maybe 
`get_property_required(PART_JSON ${BOARD} PART_JSON)`

and add the PART_JSON here: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/make/device_define.cmake#L56

common/cmake/devices.cmake
```diff
@@ -997,8 +1005,11 @@ function(ADD_FPGA_TARGET)
           symbiflow-arch-defs_SOURCE_DIR=${symbiflow-arch-defs_SOURCE_DIR}
           OUT_EBLIF=${OUT_EBLIF}
           OUT_SYNTH_V=${OUT_SYNTH_V}
+          OUT_FASM_EXTRA=${OUT_FASM_EXTRA}
+          PART_JSON=${PART_JSON}
+          INPUT_XDC_FILE=${INPUT_XDC_FILE}
           ${ADD_FPGA_TARGET_DEFINES}
-          ${QUIET_CMD} ${YOSYS} -p "${COMPLETE_YOSYS_SCRIPT}" -l ${OUT_EBLIF}.log ${SOURCE_FILES}
+          ${QUIET_CMD} ${YOSYS} -p "plugin -i xdc" -p "plugin -i fasm" -p "${COMPLETE_YOSYS_SCRIPT}" -l ${OUT_EBLIF}.log ${SOURCE_FILES}
```
litghost: Why not just modify "$COMPLETE_YOSYS_SCRIPT"?

litghost: E.g. https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/yosys/synth.tcl

common/cmake/devices.cmake
tmichalak: Done

common/cmake/devices.cmake
tmichalak: Done

common/cmake/devices.cmake
```diff
@@ -975,6 +975,11 @@ function(ADD_FPGA_TARGET)
     if(NOT "${ADD_FPGA_TARGET_INPUT_IO_FILE}" STREQUAL "")
       add_file_target(FILE ${ADD_FPGA_TARGET_INPUT_IO_FILE})
     endif()
+    if(NOT "${ADD_FPGA_TARGET_INPUT_XDC_FILE}" STREQUAL "")
+      add_file_target(FILE ${ADD_FPGA_TARGET_INPUT_XDC_FILE})
+      get_file_location(INPUT_XDC_FILE ${ADD_FPGA_TARGET_INPUT_XDC_FILE})
+      get_target_property_required(PART_JSON ${BOARD} PART_JSON)
```
litghost: `PART_JSON` will not be not be defined the option `EXPLICIT_ADD_FILE_TARGET` is specified

litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1181#discussion_r352749731

common/cmake/devices.cmake
```diff
@@ -1327,13 +1338,17 @@ function(ADD_FPGA_TARGET)
...
       COMMAND ${GENFASM_CMD}
       COMMAND
         ${CMAKE_COMMAND} -E copy ${OUT_LOCAL}/vpr_stdout.log
           ${OUT_LOCAL}/genhlc.log
+      COMMAND cat ${OUT_FASM} ${OUT_FASM_EXTRA} > ${OUT_FASM_CONCATENATED}
+      COMMAND
+        ${CMAKE_COMMAND} -E copy ${OUT_FASM_CONCATENATED} ${OUT_FASM}
       WORKING_DIRECTORY ${OUT_LOCAL}
     )
     add_custom_target(${NAME}_fasm DEPENDS ${OUT_FASM})
```
litghost: This should probably be OUT_FASM_CONCATENATED now

common/cmake/devices.cmake
```diff
@@ -1327,13 +1338,17 @@ function(ADD_FPGA_TARGET)
     # Generate FASM
     # -------------------------------------------------------------------------
     set(OUT_FASM ${OUT_LOCAL}/${TOP}.fasm)
+    set(OUT_FASM_CONCATENATED ${OUT_LOCAL}/${TOP}.concat.fasm)
```
litghost: `fasm2bels` keys off of `OUT_FASM`.  Make the `OUT_FASM` variable the final fasm file, etc.

common/cmake/devices.cmake
tmichalak: I changed from `get_target_property` to `get_target_property_required`. What else is needed?

litghost: If the option `EXPLICIT_ADD_FILE_TARGET`, this part of the control flow will never be executed, and the variable `PART_JSON` will not get defined.

common/cmake/devices.cmake
tmichalak: I am eventually copying the `top.concat.fasm` to `top.fasm` like (here)[https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1158/files#diff-2d565728ea2fbe9d3a017347718a0ba5R1349], so `OUT_FASM` is final.

litghost: Sure.  Please copy the original un-cat'd file to a new name then.

common/cmake/devices.cmake
```diff
@@ -1327,13 +1338,20 @@ function(ADD_FPGA_TARGET)
...
       DEPENDS ${OUT_ROUTE} ${OUT_PLACE} ${VPR_DEPS} ${GENFASM_TARGET}
       COMMAND ${GENFASM_CMD}
       COMMAND
         ${CMAKE_COMMAND} -E copy ${OUT_LOCAL}/vpr_stdout.log
           ${OUT_LOCAL}/genhlc.log
+      COMMAND
+        ${CMAKE_COMMAND} -E copy ${OUT_FASM} ${OUT_FASM_GENFASM}
+      COMMAND cat ${OUT_FASM} ${OUT_FASM_EXTRA} > ${OUT_FASM_CONCATENATED}
+      COMMAND
+        ${CMAKE_COMMAND} -E copy ${OUT_FASM_CONCATENATED} ${OUT_FASM}
```
litghost: Change this from a copy to a rename, to avoid leaving behind duplicate files.

common/cmake/devices.cmake
tmichalak: Ok, something like [this](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1158/commits/3e9ef98df26c301fe2f0fc3adfccae8586b15bb0#diff-2d565728ea2fbe9d3a017347718a0ba5R1350)?

litghost: Yep, perfect

common/cmake/devices.cmake
litghost: This comment is the last outstanding issue

tmichalak: Not sure what `EXPLICIT_ADD_FILE_TARGET` is used for?

litghost: It is used whenever files are shared between targets.  The implicit calls to `add_file_target` result in errors if two instances of `add_fpga_target` both share the same verilog file.  For example, if the min and full LiteX share verilog files, then `EXPLICIT_ADD_FILE_TARGET` should be specified, etc.

tmichalak: Is simply moving the `get_target_property_required(PART_JSON..` from inside of the `EXPLICIT_ADD_FILE_TARGET` check a solution here?

litghost: Yes, but I only want `get_target_property_required(PART_JSON..` called if an `XDC` file is specified.

tmichalak: Ok - [this](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1158/commits/8d4a4a7cfa92c78fb64b5311b58d962ba27f00b6) should be fine then.

PR #1162, opened 2019-11-20
Support for differential IO

mkurc-ant:
With this PR I plan to add support for OBUFDS and OBUFTDS.

TODO:
- [x] Cleanup IOB33 architecture definition.
- [x] Architecture definition for differential output
- [x] Techmap for OBUFTDS and OBUFDS
- [x] Simple test
- [x] fasm2bels support.


mkurc-ant:
At first I tried to add an additional mode "OBUFTDS_M" to `IOB33M` with the `OBUFTDS_VPR` leaf pb_type and `.output` in it plus a "OBUFTDS_S" mode to `IOB33S` with the second (complementary) `.output`. The complementary connection goes from `OBUFTDS_VPR.OB` through `DIFFO_OUT` and `DIFFO_IN` pins to the `.output` in the slave IOB. There is a pack pattern for that connection.

Unfortunately the VPR packs IOB33M and IOB33S in a way that the output `BLK-TL-LIOPAD_M.IOB_DIFFO_OUT0[0]` would be connected to the input `BLK-TL-LIOPAD_S.IOI_OLOGIC1_D1[0]` which is off course unroutable.

litghost:
> At first I tried to add an additional mode "OBUFTDS_M" to `IOB33M` with the `OBUFTDS_VPR` leaf pb_type and `.output` in it plus a "OBUFTDS_S" mode to `IOB33S` with the second (complementary) `.output`. The complementary connection goes from `OBUFTDS_VPR.OB` through `DIFFO_OUT` and `DIFFO_IN` pins to the `.output` in the slave IOB. There is a pack pattern for that connection.
> 
> Unfortunately the VPR packs IOB33M and IOB33S in a way that the output `BLK-TL-LIOPAD_M.IOB_DIFFO_OUT0[0]` would be connected to the input `BLK-TL-LIOPAD_S.IOI_OLOGIC1_D1[0]` which is off course unroutable.

I think the DIFFO_IN/_OUT pins likely need to be directly connected to top level ports (e.g. like a carry chain).  This may require special logic, or just hand coding the relevant pb_type.

mkurc-ant:
I've fixed the problem this way: In the techmap I'm splitting OBUFTDS into two cells OBUFTDS_M_VPR and OBUFTDS_S_VPR. The second one is a pass through but it ensures that VPR chooses the correct mode for the IOB33S pb_type which utilizes the differential connection.

The pack,place,route flow is working. Some FASM features are broken for now and genfasm fails. But that's easy to fix.

mkurc-ant:
In the differential output mode, the connection between IOB33M and IOB33S can be either through DIFFO_OUT/DIFFO_IN or O_OUT/O_IN. I've modelled the connection through DIFFO but Vivado shows that the second one is used. I think that it is not a problem for us as those connections are not routed.

I also haven't modelled the T_OUT to T_IN connection for 3-state control. But this does not matter either as that one is not routed either.

xc7/fasm2bels/iob_models.py
```diff
@@ -77,27 +77,52 @@ def append_obuf_iostandard_params(
...
+            print("-------------------|--------|------|")
+            for i, d, s in possible_iostandards:
+                print(
+                    " {}| {}| {}|".format(
+                        i.ljust(18),
+                        str(d).ljust(7), s.ljust(5)
+                    )
+                )
+
+            # Demote NSTD-1 to warning
```
litghost: Why?

xc7/fasm2bels/iob_models.py
```diff
@@ -116,45 +141,49 @@ def append_ibuf_iostandard_params(top, site, bel, possible_iostandards):
                 "IOSTANDARD setting provided for {} do not match"
                 "its counterpart decoded from the fasm".format(site.site.name)
             )
-            return
 
-        bel.parameters["IOSTANDARD"] = '"{}"'.format(iostandard)
+            print("Requested:")
```
litghost: What is going on with the `print`'s?

litghost:
@mkurc-ant What's the plan to fix #1169 ?  Is your plan to fix it in a follow up PR?

xc7/fasm2bels/iob_models.py
mkurc-ant: The demotion happens only when the IOSTANDARD defined in JSON does not match to what can be decoded from FASM features. In that case the standard is not set at all.

I don't consider it a permanent solution but it makes Vivado tests pass (tests don't care about IOSTANDARDs untill they are not run in HW). A proper fix for that would require setting iostandards via eg. XDC file per design and then passing it to fasm2bels. That way we won't have mismatches.

xc7/fasm2bels/iob_models.py
mkurc-ant: I wanted fasm2bels to be more informative when an IOSTANDARD mismatch occurs.

mkurc-ant:
@litghost Honestly, I haven't thought about the solution for https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1169 yet. I want to run more tests to get more details.

xc7/fasm2bels/iob_models.py
litghost: Should, probably write to stderr?

mkurc-ant:
Got strange CI error: `/tmpfs/src/github/symbiflow-arch-defs/build/xc7/archs/artix7/devices/xc7a50t-arty-swbut-roi-virt/arch.timing.xml:522 Zero or more than one sites corresponding to a tile.`

mkurc-ant:
https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1189

xc7/fasm2bels/iob_models.py
litghost: Given that we are working on XDC support and IOSTANDARD's, this issue should be resolved shortly.  Can you work with @tmichalak to make a detailed issue explaining what needs to be done?

Once the issue is created, please add a TODO here

Issue #1161, opened 2019-11-20
Inference of 3-state IOBUFs in Yosys


mkurc-ant:
The current support for IOBUFs (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1148) requires them to be explicitly instantiated and connected directly to a top-level IO port.

Using a top-level `inout` port directly in the following way won't work:
```
inout A;

...

wire A_i;
wire A_t;
wire A_o;

assign A_i = A;
assign A = (A_t == 1'b1) ? A_o : 1'bz;
```

I tried to play with the `iopadmap` Yosys command but failed to find a working solution. In some cases I ended up with an OBUFT-like cell for the 3-state output and the input taken directly from the pin (not through any buffer). In other cases I had a properly connected IOBUF cell but after optimization its output got disconnected (a bug?).

mkurc-ant:
The generated verilog for the LiteX design contains such assign statements:
```
assign eth_mdio = main_ethphy_data_oe ? main_ethphy_data_w : 1'bz;
assign main_ethphy_data_r = eth_mdio;
```

mkurc-ant:
https://github.com/YosysHQ/yosys/issues/1530

PR #1170, opened 2019-11-22
IOBUF inference using Yosys

mkurc-ant:
Fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1161


xc7/yosys/synth.tcl
```diff
@@ -1,5 +1,21 @@
 yosys -import
 
+# Infer 3-state IOBUFs.
```
litghost: I don't have a problem with any of this, I'm wondering if this should be moved into `synth_xilinx` in upstream yosys?  The changes all seem to be independent of symbiflow-arch-defs?

mkurc-ant: You're probably right. I'll check what steps does the `synth_xilinx` command do regarding IOs and see how I can augment it.

In my opinion the fact that i have to map `$tribuf` to `$_TBUF_` "manually" prior to the `iopadmap` is kind of a workaround. Besides that the `iopadmap` is not behaving properly as it always leaves some dangling wires connected directly to the io pad.

litghost: Part of the reason I think this change might belong to upstream yosys is they can comment whether what you are doing is the "proper" way.  I am not really equiped to say one way or another.

mithro: @mkurc-ant I think the correct person to talk to is @mwkmwkmwk who has been doing a lot of work in this area?

mkurc-ant: @mithro Ok, I'll do. Meanwhile the only example of how to use the `iopadmap` I found is that reddit post: https://www.reddit.com/r/yosys/comments/4hwlkx/new_support_for_tristate_buffers_in_iopadmap/

mkurc-ant:
There are two recent PRs to the upstream Yosys that addresses the issue I'm having with the inference (https://github.com/YosysHQ/yosys/pull/1527, https://github.com/YosysHQ/yosys/pull/1528). I've modified the Yosys script to use the `-iopad` option of `synth_xilinx` and when used with yosys that has those two PRs merged it works.

I've created a PR to our Yosys fork (https://github.com/SymbiFlow/yosys/pull/45) which contains those changes.

We can either wait for those PRs to get merged upstream and use the upstream Yosys or merge https://github.com/SymbiFlow/yosys/pull/45 and use out master+wip one.

litghost:
With https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1195, I believe you can rebase this PR and the inferred IOBUF should work?

xc7/yosys/synth.tcl
```diff
@@ -2,9 +2,7 @@ yosys -import
 
 # -flatten is used to ensure that the output eblif has only one module.
 # Some of symbiflow expects eblifs with only one module.
-synth_xilinx -vpr -flatten -abc9 -nosrl -noclkbuf -nodsp
-
-write_verilog $::env(OUT_SYNTH_V).premap.v
+synth_xilinx -vpr -flatten -abc9 -nosrl -noclkbuf -nodsp -iopad
```
litghost: Why did you remove the premap output?

mkurc-ant: I must have done that by accident. Fixed.

PR #1174, opened 2019-11-25
Add SERDES IDELAY IDELAYCTRL support

acomodi:

This PR is a WIP to add initial ISERDES support to symbiflow-arch-defs.

TODO:
- [x] Adding fasm2bel support
- [x] Verify it works on HW


xc7/techmap/cells_map.v
```diff
@@ -2159,6 +2159,100 @@ module OSERDESE2 (
...
+      .DATA_WIDTH_W4_6              (DATA_WIDTH == 4 || DATA_WIDTH == 6),
+      .DATA_WIDTH_W5_7              (DATA_WIDTH == 5 || DATA_WIDTH == 7),
+      .DATA_WIDTH_W8                (DATA_WIDTH == 8),
+      .NUM_CE_N2                    (NUM_CE == 2)
+  ) _TECHMAP_REPLACE_ (
+   .BITSLIP     (BITSLIP),
+   .CE1         (CE1),
+   .CE2         (CE2),
+   .CLK         (CLK),
+   .CLKB        (CLKB),
```
litghost: You are missing inverters?

acomodi: Done

acomodi:
ISERDES are currently working on HW. There are still two outstanding issues.

1. pack patterns to create the molecules.
2. ISERDES fuzzer needs to be revisited for several reasons. Some working modes (SDR wit h `DATA_WIDTH = 7`) does not work, probably for wrong bit settings. There is also a fasm2bels issue when reading fasm to assess the `DATA_WIDTH` of the ISERDES as the feature comprises multiple widths in some cases (e.g. `DATA_WIDTH.W4_6`). this PR should solve this issue https://github.com/SymbiFlow/prjxray/pull/978

acomodi:
I have added a script to utils (`add_pack_patterns.py`) to add an additional modification step that adds the required pack patterns for the IOSERDES to IOB33. This way the first issue of the previous comment (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1174#issuecomment-559852667) is temporary solved.

common/cmake/devices.cmake
```diff
@@ -272,12 +273,16 @@ function(DEFINE_DEVICE_TYPE)
...
+  # an additional step to update some special `directs` with the correct pack pattern
+  # that is needed to successfully run the packing step.
+  #
   # DEFINE_DEVICE_TYPE defines a dummy target <arch>_<device_type>_arch that
   # will build the merged architecture file for the device type.
   #
   # If UPDATE_TIMINGS is set merged arch.xml file will be processed to inject
   # timing values using data from prjxray-db/$ARCH/timigs/*sdf files
-  set(options UPDATE_TIMINGS UPDATE_TILES)
+  set(options UPDATE_TIMINGS UPDATE_TILES ADD_PACK_PATTERNS)
```
litghost: Why not use the already existing SCRIPTS mechanism?

utils/add_pack_patterns.py
```diff
@@ -0,0 +1,98 @@
+import lxml.etree as ET
```
litghost: Top level comment

utils/add_pack_patterns.py
```diff
@@ -0,0 +1,98 @@
...
+import argparse
+import re
+
+IOPAD_OLOGIC_REGEX = re.compile("OLOGICE3\.OQ_to_IOB33[MS]?\.O")
+IOPAD_ILOGIC_REGEX = re.compile("IOB33[MS]?\.I_to_ILOGICE3\.D")
+
+
+def get_top_pb_type(element):
```
litghost: Function docstring

utils/add_pack_patterns.py
```diff
@@ -0,0 +1,98 @@
+import lxml.etree as ET
+import argparse
+import re
+
+IOPAD_OLOGIC_REGEX = re.compile("OLOGICE3\.OQ_to_IOB33[MS]?\.O")
```
litghost: Comment explaining what is going on here?

litghost: Also this utility is in the generic folder, but seems xc7 specific?  Should be located in xc7/utils, or made general enough to be applicable to any arch?

common/cmake/devices.cmake
acomodi: So, I have tried to use them, the problem was that if I add multiple scripts, cmake complains about the list that I pass to the `define_device_type` function

litghost: So please fix that bug rather than add a new argument?

litghost: FYI, you cannot pass a list of lists in CMake.  You can only pass the **name** of a list, and then copy that list during looping.  I can find an example if needed.

acomodi: Sure, I'll fix it than. Can you please provide the example?

litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/make/project_xray.cmake#L483-L486

utils/add_pack_patterns.py
acomodi: Done. I think this is going to be only xc7 specific. Being it a temporary patch, I would say that it shouldn't be used in the flow for the long-term.

utils/add_pack_patterns.py
acomodi: Done

utils/add_pack_patterns.py
acomodi: Done

common/cmake/devices.cmake

common/cmake/devices.cmake
```diff
@@ -269,15 +268,10 @@ function(DEFINE_DEVICE_TYPE)
...
-  # an additional step and is updated to have include the `tiles` tags.
-  #
   # DEFINE_DEVICE_TYPE defines a dummy target <arch>_<device_type>_arch that
   # will build the merged architecture file for the device type.
-  #
-  # If UPDATE_TIMINGS is set merged arch.xml file will be processed to inject
-  # timing values using data from prjxray-db/$ARCH/timigs/*sdf files
-  set(options UPDATE_TIMINGS UPDATE_TILES)
+
+  set(options UPDATE_TIMINGS UPDATE_TILES ADD_PACK_PATTERNS)
```
litghost: Still need to remove the relevant options?

acomodi: Yep, done.

common/cmake/devices.cmake
```diff
@@ -304,9 +297,8 @@ function(DEFINE_DEVICE_TYPE)
     FILE ${DEFINE_DEVICE_TYPE_ARCH_XML}
     OUTPUT ${DEVICE_MERGED_FILE}
   )
-
   get_target_property_required(PYTHON3 env PYTHON3)
-  get_target_property(PYTHON3_TARGET env PYTHON3_TARGET)
+  get_target_property_required(PYTHON3_TARGET env PYTHON3_TARGET)
```
litghost: Leave this as `get_target_property`

acomodi: OK. I am trying to get Travis CI to pass. I have moved the update_tiles script usage to be suitable for the scripts injection mechanism. CI fails with 
```
Generating arch.tiles.xml
  File "../../../../utils/update_arch_tiles.py", line 51, in <module>
    from lxml import etree as ET
ModuleNotFoundError: No module named 'lxml'
```

I though maybe that was the issue

acomodi: The problem is that I cannot reproduce the error locally

ice40/devices/top-routing-virt/CMakeLists.txt
```diff
@@ -29,14 +29,17 @@ set(TIMING_IMPORT_CMD "${PYTHON3} ${ICE40_IMPORT_TIMING} --read_sdf ${SDF_FILE}
 add_file_target(FILE ${SDF_FILE} GENERATED)
 append_file_dependency(DEPS ${SDF_FILE})
 
+set(UPDATE_TILES "${symbiflow-arch-defs_SOURCE_DIR}/utils/update_arch_tiles.py")
+set(UPDATE_TILES_CMD "${PYTHON3} ${UPDATE_TILES} --in_xml /dev/stdin --out_xml /dev/stdout")
```
litghost: You probably need to delay the deref of `${PYTHON3}`, so change `${PYTHON3}` to `\${PYTHON3}`

litghost: In this case it is not required, but you have to ensure that each place you moved a script, the `PYTHON3` was defined, or delay the deref.

common/cmake/devices.cmake
litghost: If I had to guess your change resulted in `${PYTHON3}` being dereferenced at a location where it was not defined.  Double check

ice40/devices/top-routing-virt/CMakeLists.txt
acomodi: Great, thanks, this explains the issue. I'll fix that

common/cmake/devices.cmake
```diff
@@ -269,15 +268,9 @@ function(DEFINE_DEVICE_TYPE)
...
-  # If the UPDATE_TILES option is active, the architecture XML file will go through
-  # an additional step and is updated to have include the `tiles` tags.
-  #
   # DEFINE_DEVICE_TYPE defines a dummy target <arch>_<device_type>_arch that
   # will build the merged architecture file for the device type.
-  #
-  # If UPDATE_TIMINGS is set merged arch.xml file will be processed to inject
-  # timing values using data from prjxray-db/$ARCH/timigs/*sdf files
-  set(options UPDATE_TIMINGS UPDATE_TILES)
+
```
litghost: You still want `options` defined, just defined to an empty string, so `set(options "")`

testarch/devices/clutff-unidir-s4/CMakeLists.txt
```diff
@@ -1,8 +1,17 @@
 add_file_target(FILE arch.xml SCANNER_TYPE xml)
 
+get_target_property_required(PYTHON3 env PYTHON3)
+get_target_property(PYTHON3_TARGET env PYTHON3_TARGET)
+
+set(UPDATE_TILES "${symbiflow-arch-defs_SOURCE_DIR}/utils/update_arch_tiles.py")
+set(UPDATE_TILES_CMD "${PYTHON3} ${UPDATE_TILES} --in_xml /dev/stdin --out_xml /dev/stdout")
+set(UPDATE_TILES_DEPS "")
```
litghost: At a minimum you still want to depend on the script itself

xc7/make/device_define.cmake
```diff
@@ -169,19 +169,23 @@ function(ADD_XC7_DEVICE_DEFINE_TYPE)
...
   set(UPDATE_ARCH_TIMINGS ${symbiflow-arch-defs_SOURCE_DIR}/utils/update_arch_timings.py)
+  set(UPDATE_PACK_PATTERNS ${symbiflow-arch-defs_SOURCE_DIR}/xc7/utils/add_pack_patterns.py)
   set(PYTHON_SDF_TIMING_DIR ${symbiflow-arch-defs_SOURCE_DIR}/third_party/python-sdf-timing)
   set(BELS_MAP ${symbiflow-arch-defs_SOURCE_DIR}/xc7/bels.json)
 
+  set(ADD_PACK_PATTERN "${PYTHON3} ${UPDATE_PACK_PATTERNS} --in_arch /dev/stdin")
+  set(PACK_PATTERN_DEPS "")
+
   set(TIMING_IMPORT "${PYTHON3} ${UPDATE_ARCH_TIMINGS} --sdf_dir ${SDF_TIMING_DIRECTORY} --bels_map ${BELS_MAP} --out_arch /dev/stdout --input_arch /dev/stdin")
   set(TIMING_DEPS "")
```
litghost: This should have a dependency on BELS_MAP

acomodi:
With the two latest commits I have added also idelay support. As the IDELAY instantiation prohibits to have the `D` input of the ISERDES to be used, I have split the ISERDES in two different modes and split also the models/techmap.

With https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1191 there will be also IDELAYCTRL support (which needs to be paired with the LOC placement constraints PR https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1186)

litghost:
> two different ISERDES modes

Why is that required?

acomodi:
> > two different ISERDES modes
> 
> Why is that required?

Mainly because we need to use a different techmap. If no `D` input is provided, it will be set to `$false` in the blif, generating an unroutable situation. If there is a way to avoid this, the issue can be solved by only using pack patterns.

xc7/make/device_define.cmake
acomodi: Done

testarch/devices/clutff-unidir-s4/CMakeLists.txt
acomodi: Done

common/cmake/devices.cmake

common/cmake/devices.cmake

xc7/make/device_define.cmake
```diff
@@ -169,19 +169,23 @@ function(ADD_XC7_DEVICE_DEFINE_TYPE)
...
+  set(UPDATE_PACK_PATTERNS ${symbiflow-arch-defs_SOURCE_DIR}/xc7/utils/add_pack_patterns.py)
   set(PYTHON_SDF_TIMING_DIR ${symbiflow-arch-defs_SOURCE_DIR}/third_party/python-sdf-timing)
   set(BELS_MAP ${symbiflow-arch-defs_SOURCE_DIR}/xc7/bels.json)
 
+  set(ADD_PACK_PATTERN "${PYTHON3} ${UPDATE_PACK_PATTERNS} --in_arch /dev/stdin")
+  set(PACK_PATTERN_DEPS UPDATE_PACK_PATTERNS)
+
   set(TIMING_IMPORT "${PYTHON3} ${UPDATE_ARCH_TIMINGS} --sdf_dir ${SDF_TIMING_DIRECTORY} --bels_map ${BELS_MAP} --out_arch /dev/stdout --input_arch /dev/stdin")
-  set(TIMING_DEPS "")
+  set(TIMING_DEPS BELS_MAP UPDATE_ARCH_TIMINGS)
```
litghost: I believe this should be the files themselves, not the variable names

acomodi: Oh sure, my mistake. Should `add add_file_target` I suppose

litghost: `add_file_target` is not required here, as these aren't generated files, and the dependency list goes directly to the add_custom_command

acomodi: Done

litghost:
Looks like you broke something in the CI:
```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs/build/ice40/devices/top-routing-virt && /tmpfs/src/github/symbiflow-arch-defs/build/env/conda/bin/python3 /tmpfs/src/github/symbiflow-arch-defs/ice40/utils/ice40_import_bel_timing.py --read_sdf timings_hx1k.sdf --write_arch_xml /dev/stdout --read_arch_xml /dev/stdin < /tmpfs/src/github/symbiflow-arch-defs/build/ice40/devices/top-routing-virt/arch.tiles.xml > arch.timing.xml
usage: ice40_import_bel_timing.py [-h] [--read_sdf READ_SDF]
                                  [--read_arch_xml READ_ARCH_XML]
                                  [--write_arch_xml WRITE_ARCH_XML] [-v V]
ice40_import_bel_timing.py: error: argument --read_sdf: can't open 'timings_hx1k.sdf': [Errno 2] No such file or directory: 'timings_hx1k.sdf'
```

The CI uses ninja and it looks like ninja + ice40 targets are broken.

acomodi:
Ok, I will use ninja to build locally to solve the issue. I'll check if this happens also for xc7, because the sdf script is called in a similar way.

litghost:
Good amount of diff_fasm failures.  

Failures appear to come in two flavers:
 - Interconnect connections that go missing
   - If I had to guess, the interconnect stuff is a a result of driving an unused site (e.g. IDELAY), which could be avoided with an IDELAY mode that makes no connections?
 - IDELAY values that go missing
  - If I had to guess, if the IDELAY is in a certain mode, the IDELAY_VALUE stuff is always just set to 0?

```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc7/tests/serdes && diff -u /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc7/tests/serdes/serdes_basys3_5_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/top.bit.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc7/tests/serdes/serdes_basys3_5_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/design_serdes_basys3_5_sdr_vivado.bit.fasm
--- /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc7/tests/serdes/serdes_basys3_5_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/top.bit.fasm	2019-12-03 16:19:51.467911771 -0800
+++ /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc7/tests/serdes/serdes_basys3_5_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/design_serdes_basys3_5_sdr_vivado.bit.fasm	2019-12-03 16:24:45.662583875 -0800
@@ -393,29 +393,14 @@
 INT_L_X0Y24.LH12.LV_L9
 INT_L_X0Y24.NR1BEG1.NN6END1
 
-INT_L_X0Y25.BYP_ALT6.GFAN1
 INT_L_X0Y25.CLK_L0.GCLK_L_B1
-INT_L_X0Y25.CLK_L1.FAN_BOUNCE5
 INT_L_X0Y25.CTRL_L0.WW4END2
 INT_L_X0Y25.CTRL_L1.WW4END2
 INT_L_X0Y25.ER1BEG_S0.LOGIC_OUTS_L7
-INT_L_X0Y25.FAN_ALT5.GFAN1
 INT_L_X0Y25.GFAN0.GND_WIRE
-INT_L_X0Y25.GFAN1.GND_WIRE
 INT_L_X0Y25.IMUX_L0.GFAN0
-INT_L_X0Y25.IMUX_L12.GFAN1
-INT_L_X0Y25.IMUX_L25.GFAN0
-INT_L_X0Y25.IMUX_L26.GFAN0
-INT_L_X0Y25.IMUX_L30.GFAN1
-INT_L_X0Y25.IMUX_L32.GFAN0
-INT_L_X0Y25.IMUX_L33.GFAN0
 INT_L_X0Y25.IMUX_L34.WL1END1
-INT_L_X0Y25.IMUX_L35.GFAN0
-INT_L_X0Y25.IMUX_L36.GFAN1
-INT_L_X0Y25.IMUX_L38.GFAN1
-INT_L_X0Y25.IMUX_L39.GFAN1
 INT_L_X0Y25.IMUX_L40.EL1END0
-INT_L_X0Y25.IMUX_L41.GFAN0
 INT_L_X0Y25.IMUX_L42.EL1END1
 INT_L_X0Y25.IMUX_L43.NR1END1
 INT_L_X0Y25.IMUX_L44.WL1END2
@@ -907,10 +892,7 @@
 LIOI3_X0Y17.OLOGIC_Y0.OSERDES.DATA_RATE_TQ.BUF
 
 LIOI3_X0Y25.IDELAY_Y0.IDELAY_TYPE_FIXED
-LIOI3_X0Y25.IDELAY_Y1.DELAY_SRC_IDATAIN
 LIOI3_X0Y25.IDELAY_Y1.IDELAY_TYPE_FIXED
-LIOI3_X0Y25.IDELAY_Y1.IDELAY_VALUE[2:0] = 3'b101
-LIOI3_X0Y25.IDELAY_Y1.ZIDELAY_VALUE[4:0] = 5'b11010
 LIOI3_X0Y25.ILOGIC_Y1.IDDR_OR_ISERDES.IN_USE
 LIOI3_X0Y25.ILOGIC_Y1.IFF.DDR_CLK_EDGE.OPPOSITE_EDGE
 LIOI3_X0Y25.ILOGIC_Y1.IFF.SRTYPE.SYNC
```

acomodi:
@litghost I am working on a fix. The issue is that I still did not add support for IDELAY in fasm2bels. The problem is that when instantiating IDELAYs, IDELAYCTRL is needed, but this will be added with PR https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1191.

litghost:
> @litghost I am working on a fix. The issue is that I still did not add support for IDELAY in fasm2bels. The problem is that when instantiating IDELAYs, IDELAYCTRL is needed, but this will be added with PR #1191.

Ah, that makes sense.  Ya, let's fix that before we merge this.  I recommended in #1191 that we split the BUFR and IDELAYCTRL stuff.  Could you move the IDELAYCTRL support into this PR?

acomodi:
Sure, to do so I need to add also all the support to instantiate the HCLK_IOI tile from https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1191

acomodi:
There is an additional issue. I got yosys to keep the IDELAYCTRL.

The problem now is that VPR doesn't like primitives that do not have any output pin.

acomodi:
When dealing with `fasm2bels`, I get an error when choosing the sink for the ISERDES, based on the presence of the IDELAY.

- Input `D` (from the IOPAD) must be chosen as sink of the ISERDES when there is *no* IDELAY
- Input `DDLY` (from the IDELAY) must be chosen as sink of the ISERDES whene there is IDELAY

The error I get is the following:

```
[100%] Generating serdes_basys3_8_ddr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/top_bit.v
  File "/home/alessandro/projects/symbiflow/symbiflow-arch-defs/build/env/conda/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/alessandro/projects/symbiflow/symbiflow-arch-defs/build/env/conda/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/alessandro/projects/symbiflow/symbiflow-arch-defs/xc7/fasm2bels/__main__.py", line 4, in <module>
    main()
  File "/home/alessandro/projects/symbiflow/symbiflow-arch-defs/xc7/fasm2bels/fasm2bels.py", line 302, in main
    for l in top.output_verilog():
  File "/home/alessandro/projects/symbiflow/symbiflow-arch-defs/xc7/fasm2bels/verilog_modeling.py", line 1443, in output_verilog
    assert len(rhs) == 1
AssertionError
```

And the code that produces the error is the following: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/b9c81c262c0ba92f94ef4d7aaf62fe6ac77a4c26/xc7/fasm2bels/verilog_modeling.py#L1442-L1446

Any idea on how to solve this?

litghost:
> The error I get is the following:

The error is indicting two wires are trying to assign to the same wire, which is obviously a bug.  A net can only have 1 source.

So the real question is why did you code generated a net with two sources?

acomodi:
> So the real question is why did your code generated a net with two sources?

This is very strange. When I add the `DDLY` sink, it should only have one source, which is `DATAOUT`

The problem is that `DDLY` appears to be connected also to the `IPAD` coming from the `IOB33` site. This is wrong, as the IPAD output can be connected only to port `D` of the ISERDES or port `IDATAIN` of IDELAY.

acomodi:
I think I may have found the issue: there are pips connecting `IDELAY_IDATAIN` to `IDELAY_DATAOUT`, which will let fasm2bels to create an additional connection from the `ISERDES_DDLY` pin to the `IOB33_O` pin

acomodi:
The last outstanding issue preventing this PR to go green on the `Vendor Tool Tests` is the IDELAYCTRL that gets ignored by VPR, in fact, nets without a sink get ignored.

I have tried to assign the output of the IDELAYCTRL (namely the `RDY` pin) to an output led, but I got the following unroutable error:
```
# Routing
  RR Graph Nodes: 1288326
  RR Graph Edges: 7938537
Confirming router algorithm: TIMING_DRIVEN.
Warning 186: No routing path for connection to sink_rr 245847, retrying with full device bounding box
Warning 187: No routing path for connection to sink_rr 551334, retrying with full device bounding box
Cannot route from BLK-TL-HCLK_IOI3.HCLK_IOI_IDELAYCTRL_RDY[0] (RR node: 501707 type: SOURCE location: (2,133) class: 3 capacity: 1) to BLK-TL-SLICEL.A2[0] (RR node: 551334 type: SINK location: (12,153) class: 1 capacity: 1) -- no possible path
Failed to route connection from 'serdes_test.idelayctrl' to '$auto$alumacc.cc:485:replace_alu$51.slice[0].carry4_1st_full' for net 'serdes_test.RDY' (#76)
Routing failed.
# Routing took 0.36 seconds (max_rss 2351.8 MiB, delta_rss +0.0 MiB)
Circuit is unroutable with a channel width factor of 500.
VPR failed to implement circuit

```

Any idea on how to solve this?

litghost:
> Any idea on how to solve this?

Did you verify where the RDY signal is connected to in Vivado?  Can it be routed to the general interconnect?

acomodi:
Yes, It goes to the bottom IOI tile interconnect and then it reaches the first INT tile on the right, therefore a valid route to any tile should be found. Maybe a missing segment definition?

litghost:
> Yes, It goes to the bottom IOI tile interconnect and then it reaches the first INT tile on the right, therefore a valid route to any tile should be found. Maybe a missing segment definition?

Walk the route and see if any of the wires are of the "unknown" segment type.  If so, that should be fixed.

litghost:
> Any idea on how to solve this?

The `route_diag` tool can be used to debug this issue.  Compile the tool with debugging prints enables, and you should be able to get additional information about what led to the unroutable error.  In addition, use mkurc's python based router to verify that the graph has the route in question.

acomodi:
```
./utils/rr_graph_walk/rr_graph_walk.py  --rr_graph /home/alessandro/projects/symbiflow/symbiflow-arch-defs/build/xc7/archs/artix7/devices/rr_graph_xc7a50t-bottom_test.rr_graph.real.xml -s 501707 -e 551334
Loading and parsing XML file...
| |                                                    #                                                   | 55655735 Elapsed Time: 0:02:20
1288326 nodes, 1107728 edges
Walking forward from 501707 to 551334...
No route to the target node!

```

Apparently there is no routing path in the rr_graph between the two nodes. I'll debug the rr_graph_import to see what is happening.

acomodi:
@litghost I have separated the serdes tests in two sets:
1. with IDELAYS: these one do not have the vivado target. The only outstanding issue is the IDELAYCTRL. I still need to debug what is happening as there is no connection between its output and logic cells in the rr_graph.
2. without IDELAYS: these have also vivado targets to compute the `diff_fasm` targets and should complete successfully.

This was done mainly because IMO, this PR introduces quite a lot of changes and it would be better to integrate it ASAP and correct the IDELAYCTRL issue with a second PR.

PR #1179, opened 2019-11-25
Increase congestion schedule

litghost:


litghost:
This is 10 minutes faster than #1178, but because #1180 hadn't been merged, we can't do a full comparisons.

PR #1180, opened 2019-11-25
Print QoR for multi-clock designs.

litghost:


utils/print_qor.py
```diff
@@ -42,6 +42,10 @@ def scan_critical(fname):
                         critical_path = float(parts[3])
                         fmax = float(parts[6])
                         return str(critical_path), str(fmax)
+                    elif len(parts) == 5 and parts[4].strip() == 'ns':
```
mithro: A comment which included an example of the string this was parsing would be good.

litghost: Sure

Issue #1182, opened 2019-11-26
PLL and BUFG need to be placed in the same clock row region


acomodi:
There are several situations for which the clock is being routed from the PLL output to the BUFG tile that are placed in incompatible locations.

Incompatible location example:
- PLL placed in TOP half (e.g. X1Y1)
- BUFG placed in BOTTOM half (e.g. X0Y0).

In such a situation, nor the BUFG output can be driven to the PLL clk input neither PLL output to BUFG input.

This situation causes fasm2bels to fail, as well a non-working output bitstream produced with the symbiflow toolchain.

acomodi:
@litghost FYI

litghost:
I've proposed #1231 as a solution to this issue.

acomodi:
Closing. Fixed with https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1237

Issue #1183, opened 2019-11-26
Add capability to import tiles with capacity


acomodi:
There are situation for which it is necessary to have `pb_type` definitions that have `capacity > 1`.

An example would be to have the `BUFG` type defined with a capacity of 12 instead of having 12 instantiation of the same pb_type.

This would potentially solve also issue https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1182, as the two BUFG tiles (with 16 BUFGCTRL sites each) will be able to be placed both in the TOP or BOTTOM half of the device.

acomodi:
@litghost FYI

litghost:
I've started looking into this, and it might be hard :(

The problem as I see it is how capacity is defined.  VPR defines capacity by completely duplicating the tile N times, when what I really want to 1 tile with N placement sites inside.  I want the pin definitions to be explicit, etc.

I'll open an issue upstream.

acomodi:
The up-side of this though is that actually the CLK_BUFG tiles currently have the same pin names repeated 16 times. This is good because the tile does not have extra pins that do not get duplicated, so, for VPR, it could be enough to have the pins stated only once for the tile.

I think the problem could be to link the pins when building the rr_graph.

litghost:
> I think the problem could be to link the pins when building the rr_graph.

Yes.  That would work, but I'm also thinking ahead to the the heterogenous capacity problem.

I think what I want to do is de-couple capacity from pin class definition (via an attribute), and then require an explicit physical type to logical tile mapping of the pins.  This will allow flexiblity in the pin assignments from the physical tile definition to the logical tile definition.  The capacity definition in this case will modify how many logical instances exist, but not modify how the physical ports are created.

This has a two fold benefit.  First it allows the existing routing import logic to work.  Second, when we add heterogeneous capacity, the requirement for uniform port definitions is relaxed.

litghost:
> The up-side of this though is that actually the CLK_BUFG tiles currently have the same pin names repeated 16 times. This is good because the tile does not have extra pins that do not get duplicated, so, for VPR, it could be enough to have the pins stated only once for the tile.
> 
> I think the problem could be to link the pins when building the rr_graph.

Given that I'm going on vacation soon, it might be worthwhile for you to try to implement capacity using the existing VPR version.  That may be faster, and then in the new year when I can finish the flexible physical tile <-> logical tile mapping, we can switch over.  Thoughts?

[...]

PR #1186, opened 2019-11-27
Add placement constraints using LOC

acomodi:
This PR is to add the capability to add design constraints through the `(* LOC="<site_location>" *)` attribute.

Given that VPR only admits user defined placements only for IOs, VPR needs to be modified to accept also non-io-related placement blocks.

Right now the placement fixed locations are alongside with the ones of `top_io.place`. They should probably be separated in two different files (placement constraints and iopad locations).
VPR needs also to be changed accordingly.

This addresses issue https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1182


xc7/utils/prjxray_create_place_constraints.py
```diff
@@ -0,0 +1,94 @@
...
+        args.output.write(line)
+
+    place_constraints = vpr_place_constraints.PlaceConstraints()
+    place_constraints.load_loc_sites_from_net_file(args.net)
+
+    with sqlite3.connect(args.connection_database) as conn:
+        for block, loc in place_constraints.get_loc_sites():
+            vpr_loc = get_vpr_coords_from_site_name(conn, loc)
+
+            vpr_loc = (vpr_loc[0], vpr_loc[1], 0)
```
litghost: Need a TODO to handle Z as part of #1183

litghost:
> They should probably be separated in two different files (placement constraints and iopad locations).
> VPR needs also to be changed accordingly.

I'm not sure two files are required.  Good question for Vaughn and Kevin.

acomodi:
> I'm not sure two files are required. Good question for Vaughn and Kevin.

I think that, if we handle differently how VPR treats the `user_pad_file` and make it more generic in order not to only check the `io_types`, there will be no problem in having only one constraints file. But I do agree it is something to be discussed with Vaughn and Kevin

xc7/utils/prjxray_create_place_constraints.py
acomodi: Done

acomodi:
What version of VTR is the `VtR + SymbiFlow Architecture Definitions (Presubmit)` CI build using? Because I get a failed build output because of the `pin_mapping` attribute.

```
Error 1: /tmpfs/src/github/symbiflow-arch-defs/build/xc7/archs/artix7/devices/xc7a50t-basys3-roi-virt/arch.timing.xml:426 Unexpected attribute 'pin_mapping' found on node 'site'. Expected (possibly) one of: 'pb_type'.
```

master+wip VtR should be able to handle this

litghost:
What's left WIP about this PR?

acomodi:
> What's left WIP about this PR?

Still need to update to newest VtR version that has the placement constraints patch for non-IO tiles. I am going to update it now.

common/cmake/devices.cmake
```diff
@@ -1233,7 +1261,30 @@ function(ADD_FPGA_TARGET)
     add_output_to_fpga_target(${NAME} IO_PLACE ${OUT_IO_REL})
     append_file_dependency(VPR_DEPS ${OUT_IO_REL})
 
-    set(FIX_PINS_ARG --fix_pins ${OUT_IO})
+    if(NOT ${NO_PLACE_CONSTR})
+      append_file_dependency(CONSTR_DEPS ${OUT_IO_REL})
```
litghost: Need to initialize `CONSTR_DEPS` to empty string before appending to it (e.g. `set(CONSTR_DEPS "")`)

common/cmake/devices.cmake
```diff
@@ -1233,7 +1261,30 @@ function(ADD_FPGA_TARGET)
...
+      # Generate LOC constrains
+      string(CONFIGURE ${PLACE_CONSTR_TOOL_CMD} PLACE_CONSTR_TOOL_CMD_FOR_TARGET)
+      separate_arguments(
+        PLACE_CONSTR_TOOL_CMD_FOR_TARGET_LIST UNIX_COMMAND ${PLACE_CONSTR_TOOL_CMD_FOR_TARGET}
+      )
+
+      add_custom_command(
+        OUTPUT ${OUT_CONSTR}
+        DEPENDS ${CONSTR_DEPS}
+        COMMAND ${PLACE_CONSTR_TOOL_CMD_FOR_TARGET_LIST} --input ${OUT_IO} --output ${OUT_CONSTR}
```
litghost: Rather than requiring the place constraint tool to take the `--input` and `--output` arguments, simply have them reference `OUT_IO` and `OUT_CONSTR`, which is how the other tools work.  Alternatively, use stdin/stdout redirection to pipe OUT_IO to to the tool, and pipe OUT_CONSTR from the tool.

litghost: Either way, update the DEFINE_ARCH with how the interface works.

common/cmake/devices.cmake
```diff
@@ -79,6 +82,14 @@ function(DEFINE_ARCH)
   # * OUT_EBLIF - Input path to EBLIF file.
   # * INPUT_IO_FILE - Path to input io file, as specified by ADD_FPGA_TARGET.
   #
+  # PLACE_TOOL_CONSTR_CMD variables:
+  #
+  # * PLACE_TOOL_CONSTR - Value of PLACE_TOOL_CONSTR property of <arch>.
+  #
+  # This command enables the possibility to add an additional step consisting
+  # on the addition of extra placement constraints through the usage of the chosen
+  # script.
```
litghost: How is data passed to script?  How is data retrieved from the script?

xc7/make/arch_define.cmake
```diff
@@ -56,6 +56,15 @@ function(ADD_XC7_ARCH_DEFINE)
...
         --net \${OUT_NET}"
+    PLACE_CONSTR_TOOL
+      ${symbiflow-arch-defs_SOURCE_DIR}/xc7/utils/prjxray_create_place_constraints.py
+    PLACE_CONSTR_TOOL_CMD "${CMAKE_COMMAND} -E env \
+    PYTHONPATH=${symbiflow-arch-defs_SOURCE_DIR}/utils \
+    \${PYTHON3} \${PLACE_CONSTR_TOOL} \
+        --net \${OUT_NET} \
+        --input /dev/stdin \
+        --output /dev/stdout \
+        --connection_database ${CMAKE_CURRENT_BINARY_DIR}/channels.db"
```
litghost: This path to the connection database won't work.  See how `FASM_TO_BIT_EXTRA_ARGS` work.

acomodi:
I get the following CI error, probably related to the CI itself:
```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build && /usr/bin/cmake -E echo Taking\ /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/bin/conda.lock && flock /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/bin/conda.lock /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/bin/conda install --force iverilog && /usr/bin/cmake -E touch_nocreate /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/bin/iverilog && /usr/bin/cmake -E touch_nocreate /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/bin/vvp
Taking /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/bin/conda.lock
Collecting package metadata (current_repodata.json): ...working... failed


WARNING: The --force flag will be removed in a future conda release.
         See 'conda install --help' for details about the --force-reinstall
         and --clobber flags.



CondaHTTPError: HTTP 520 ORIGIN ERROR for url <https://conda.anaconda.org/pkgw-forge/linux-64/repodata.json>
Elapsed: 00:03.774845
CF-RAY: 540583a398b8770a-LAX

A remote server error occurred when trying to retrieve this URL.

A 500-type error (e.g. 500, 501, 502, 503, etc.) indicates the server failed to
fulfill a valid request.  The problem may be spurious, and will resolve itself if you
try your request again.  If the problem persists, consider notifying the maintainer
of the remote server.

```

xc7/make/arch_define.cmake
acomodi: Done

common/cmake/devices.cmake

common/cmake/devices.cmake

acomodi:
> can you add an example to test that this is working?

Done, I have added a test for the `simple_ff` actually and verified that the CLB that is being used is the one containing  the specified SLICE LOC attribute in the design.

litghost:
> > can you add an example to test that this is working?
> 
> Done, I have added a test for the `simple_ff` actually and verified that the CLB that is being used is the one containing the specified SLICE LOC attribute in the design.

I assume you verify this manually?  Can you please add a `add_custom_target` that passes or failes based on whether the LOC constraints works?  And the add that target to the target `all_xc7_tests`?

Thanks

acomodi:
@litghost Sure. I am thinking on the easiest, yet sound method to do it. What about checking the `place_constraints` file and the final placement file? 

If the block assigned in the `top_constraints.place` gets assigned to another location we assume it has failed.

We assume of course that the `top_constraints.place` file is correctly created, with the right placement location of the specified LOCed site. What do you think about this solution?

litghost:
> @litghost Sure. I am thinking on the easiest, yet sound method to do it. What about checking the `place_constraints` file and the final placement file?
> 
> If the block assigned in the `top_constraints.place` gets assigned to another location we assume it has failed.
> 
> We assume of course that the `top_constraints.place` file is correctly created, with the right placement location of the specified LOCed site. What do you think about this solution?

I was thinking something like:
"grep <correct tile>.<correct slice> <output fasm> | <check that a FF was enabled>"

acomodi:
> 
> I was thinking something like:
> `grep "<correct tile>\.<correct slice>" <output fasm> | <check that a FF was enabled>`

Done

PR #1103, opened 2019-10-22
[WIP] BUFR support

mkurc-ant:
This PR adds support to the BUFR primitive.

Not sure whether the "bypass" mode of the BUFR is modeled correctly (as a separate primitive). The "bypass" and "regular" mode differ in timings (so does the manual say).

TODO:
- [x] Fix routing issues
- [ ] Make sure that correct fasm features are emitted
- [ ] Add timings
- [ ] Add simulation model for the BUFR_VPR

PR #1191, opened 2019-12-02
Add BUFR support

acomodi:
This PR is to add support for both BUFR (from https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1103) and IDELAYCTRL sites.

BUFR still needs to be fixed as it does not work on HW yet:
- [ ] fasm2bels support

IDELAYCTRL needs to have a test. To do so https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1174 needs to be merged first.


acomodi:
I have noticed that there are no FASM features for IDELAYCTRL (e.g. IN_USE). Moreover, the litex minitest instantiates only one of those, with no output signal:

```
IDELAYCTRL IDELAYCTRL(
	.REFCLK(clk200_clk),
	.RST(main_ethernetsoc_ic_reset)
);
```

litghost:
> I have noticed that there are no FASM features for IDELAYCTRL (e.g. IN_USE). Moreover, the litex minitest instantiates only one of those, with no output signal:
> 
> ```
> IDELAYCTRL IDELAYCTRL(
> 	.REFCLK(clk200_clk),
> 	.RST(main_ethernetsoc_ic_reset)
> );
> ```

I believe this is correct.  We simply need to route the relevant nets.

acomodi:
@litghost Ok, I think it is not a problem. The only think is to ensure that it does not get optimized and removed by Yosys, as I have seen that the IDELAYCTRL is absent from the `top.eblif`

litghost:
> @litghost Ok, I think it is not a problem. The only think is to ensure that it does not get optimized and removed by Yosys, as I have seen that the IDELAYCTRL is absent from the `top.eblif`

Sure.  There should be a way to do that.

xc7/primitives/bufr/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_file_target(FILE ntemplate.bufrN.model.xml SCANNER_TYPE xml)
```
litghost: Can we add capacity support rather than continuing to use this method?

litghost:
Can we split the BUFR (and capacity) changes into a PR seperate from the IDELAYCTRL changes?  I believe the capacity changes might be larger, and getting IDELAYCTRL in sooner seems to be a higher priority?

acomodi:
Sure, makes perfect sense, I'll split those

litghost:
Should probably update this PR to be BUFR only?

Issue #1189, opened 2019-11-29
Kokoro CI target "VtR + SymbiFlow Architecture Definitions" uses wrong VPR version


mkurc-ant:
I've observed some CI tests failing recently for a not obvious reason. I've also noticed that the Kokoro target "VtR + SymbiFlow Architecture Definitions (Presubmit)" uses the upstream VPR instead of the conda package pointed at by CMakeLists.txt file.

Is the use of upstream VPR intentional there?

litghost:
I believe I see the issue, it is using master instead of master+wip.

litghost:
@mithro

litghost:
Issue has been resolved!

mithro:
@litghost Have you confirmed Kokoro is now running on the right branch?

litghost:
> @litghost Have you confirmed Kokoro is now running on the right branch?

Yes, https://source.cloud.google.com/results/invocations/154bb11d-4c63-413c-83de-7ab80fbaaeb2/details

mithro:
Awesome!

PR #1192, opened 2019-12-02
Use length of pip names without trailing numbers.

litghost:
LV#.LV## is the same as LV#.LV#, so it sorts lexigraphically.
SW#BEG.SE#END0_S# is greater than SW#BEG.SE#END#.

### DIFF
--- a/xc7/utils/prjxray_edge_library.py
+++ b/xc7/utils/prjxray_edge_library.py
@@ -12,6 +12,7 @@ from prjxray.site_type import SitePinDirection
 from prjxray_constant_site_pins import yield_ties_to_wire
 from lib.connection_database import get_track_model, get_wire_in_tile_from_pin_name
 from lib.rr_graph.graph2 import NodeType
+import re
 import math
 import numpy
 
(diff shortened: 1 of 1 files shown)


mithro:
I solved this here -> https://github.com/SymbiFlow/prjxray/blob/e328dbdc146089b6d041c273278419ca9f315374/utils/xjson.py#L8-L21
```python
def extract_numbers(s):
    """
    >>> extract_numbers("CLK_HROW_WR10END2_3")
    ('CLK_HROW_WR', 10, 'END', 2, '_', 3)
    >>> extract_numbers("VBRK_WR1END2")
    ('VBRK_WR', 1, 'END', 2)
    """
    bits = []
    for m in re.finditer("([^0-9]*)([0-9]*)", s):
        if m.group(1):
            bits.append(m.group(1))
        if m.group(2):
            bits.append(int(m.group(2)))
    return tuple(bits)
```

There also seems to be -> https://github.com/SymbiFlow/prjxray/blob/e328dbdc146089b6d041c273278419ca9f315374/utils/sort_db.py#L91-L110 and https://github.com/SymbiFlow/prjxray/blob/e328dbdc146089b6d041c273278419ca9f315374/utils/sort_db.py#L199-L213
```python
def extract_num(i):
    """Extract number from a string to be sorted.

    >>> extract_num('BLAH123')
    ('BLAH', 123)
    >>> extract_num('123')
    123
    >>> extract_num('BLAH')
    'BLAH'
    >>> extract_num('')
    ''
    """
    g = NUM_REGEX.match(i).groups()
    if len(g[-1]) == 0:
        return i
    i = int(g[-1])
    if len(g[0]) == 0:
        return i
    else:
        return (g[0], i)

def sortable_tag(t):
    """
    >>> sortable_tag("CLBLL_L.CLBLL_L_A.CLBLL_L_A1")
    ('CLBLL', 'L', 'CLBLL', 'L', 'A', 'CLBLL', 'L', ('A', 1))

    >>> sortable_tag("CLBLL_L.CLBLL_LOGIC_OUTS23.CLBLL_LL_DMUX")
    ('CLBLL', 'L', 'CLBLL', 'LOGIC', ('OUTS', 23), 'CLBLL', 'LL', 'DMUX')

    >>> sortable_tag("BRAM_L.RAMB18_Y0.INIT_B[9]")
    ('BRAM', 'L', ('RAMB', 18), ('Y', 0), 'INIT', 'B', 9)

    >>> sortable_tag("BRAM_L.RAMB18_Y0.READ_WIDTH_A_18")
    ('BRAM', 'L', ('RAMB', 18), ('Y', 0), 'READ', 'WIDTH', 'A', 18)
    """
    return tuple(extract_num(i) for i in split_all(t, '_.[]') if i != '')
```

acomodi:
There is also an issue with `VtR + arch-defs` CI (https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1189). Apparently arch-defs is run with a wrong version of VtR, causing CI to fail when parsing the architecture file because of the `pin_mapping` attribute of `sites`

PR #1195, opened 2019-12-03
Update to latest upstream yosys.

### LINKED ISSUE #1161: Inference of 3-state IOBUFs in Yosys

litghost:
Fixes #1161

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ setup_env(
 add_conda_package(
   NAME yosys
   PROVIDES yosys
-  PACKAGE_SPEC "yosys 0.8_3555_g39a9dc5c 20190930_213940"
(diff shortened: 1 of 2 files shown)


litghost:
> LGTM if CI goes green.

Vendor test failure are pre-existing failures.

PR #1204, opened 2019-12-05
Import of ROI-less graph for Zynq

mkurc-ant:
This PR fixes some issues with prjxray import scripts that occur when importing ROI-less Zynq graph. Currently it adds the device target `xc7z010` which is the full ROI-less version for that part. It also adds the buttons test for that device (currently only that one).

The most important modification to the import scripts is filtering out `IOPAD` sites along with wires and PIPs for `PSS*` tiles. Those elements are not relevant to the PL (FPGA fabric) and are controlled solely by the PS (ARM core).

Right now we can't have both `xc7z010` and `xc7z020` because both parts contain different tile sets (020 has "L" IOBs and "R" CMT tiles). We would need separate architectures for those two.


xc7/utils/prjxray_arch_import.py
```diff
@@ -487,10 +487,20 @@ def insert_constant_tiles(conn, model_xml, complexblocklist_xml, tiles_xml):
     c.execute('SELECT pkey FROM tile_type WHERE name = "NULL";')
     null_tile_type_pkey = c.fetchone()[0]
 
+    # Get all 'NULL' tile locations
+    c.execute(
+        'SELECT grid_x, grid_y FROM phy_tile WHERE tile_type_pkey = ?',
+        (null_tile_type_pkey, )
+    )
+    locs = sorted(list(c.fetchall()))
```
litghost: Instead of fetchall, sort and limit in the SQL call.

xc7/utils/prjxray_form_channels.py
```diff
@@ -197,19 +197,72 @@ def get_switch(pip, pip_timing):
...
+            for p in tile_type.get_pips():
+                if p.net_from == site_pin.wire:
+                    masked_pips.append(p.name)
+                    masked_wires.append(p.net_to)
+                if p.net_to == site_pin.wire:
+                    masked_pips.append(p.name)
+                    masked_wires.append(p.net_from)
+
+    # Masked sites names
+    masked_sites = ["{}_{}".format(s.prefix, s.name) for s in iopad_sites]
```
litghost: Instead of making a string, why not just make a tuple of `(prefix, name)`?

xc7/tests/buttons/CMakeLists.txt
```diff
@@ -37,3 +30,25 @@ add_vivado_target(
...
+add_file_target(FILE zybo.pcf)
+add_file_target(FILE zyboz7.pcf)
+
+add_fpga_target(
+  NAME buttons_zybo_full
+  BOARD zybo-full
+  SOURCES buttons_zybo.v
+  INPUT_IO_FILE zybo.pcf
+  EXPLICIT_ADD_FILE_TARGET
+  )
```
litghost: Can you add a `add_vivado_target`?

litghost:
Can you add a test that includes a clock (e.g. blinky)?  And add a `add_vivado_target`?

mkurc-ant:
I addressed the comments and added zybo target for the `counter` test. I also added vivado targets for both `counter` and `buttons` but left fasm diff check disabled.

litghost:
> left fasm diff check disabled.

Why?

xc7/tests/counter/CMakeLists.txt
```diff
@@ -1,10 +1,18 @@
...
+  SOURCES counter_bufg.v
   INPUT_IO_FILE zybo.pcf
   )
 
+add_vivado_target(
+    NAME counter_zybo_vivado
+    PARENT_NAME counter_zybo
+    CLOCK_PINS clk
+    CLOCK_PERIODS 10.0
+    #DISABLE_DIFF_TEST
```
litghost: I assume you meant to disable this test?

mkurc-ant:
@litghost For the buttons test these features differ
```
@@ -1,3 +1,7 @@
+CFG_CENTER_MID_X67Y32.ALWAYS_ON_PROP1
+CFG_CENTER_MID_X67Y32.ALWAYS_ON_PROP2
+CFG_CENTER_MID_X67Y32.ALWAYS_ON_PROP3
```
Looks like that we are not emitting them now in SymbiFlow.

And for the counter test I get Vivado error:
```
ERROR: [Place 30-574] Poor placement for routing between an IO pin and BUFG. If this sub optimal condition is acceptable for this design, you may use the CLOCK_DEDICATED_ROUTE constraint in the .xdc file to demote this message to a WARNING. However, the use of this override is highly discouraged. These examples can be used directly in the .xdc file to override this clock rule.
        < set_property CLOCK_DEDICATED_ROUTE FALSE [get_nets RIOB33_X31Y75_IOB_X0Y76_I] >

        RIOB33_X31Y75_IOB_X0Y76_IBUF (IBUF.O) is locked to IOB_X0Y76
        bufgCLK_BUFG_BOT_R_X82Y48_BUFGCTRL_X0Y0_BUFGCTRL (BUFGCTRL.I0) is locked to BUFGCTRL_X0Y0
```
When I suppress that error it shows that the clock is routed to a BUFG in different top/bottom portion of the fabric than the clock IOB is located in. And the clock tree is not used for that route.

litghost:
> @litghost For the buttons test these features differ
> 
> ```
> @@ -1,3 +1,7 @@
> +CFG_CENTER_MID_X67Y32.ALWAYS_ON_PROP1
> +CFG_CENTER_MID_X67Y32.ALWAYS_ON_PROP2
> +CFG_CENTER_MID_X67Y32.ALWAYS_ON_PROP3
> ```
> 
> Looks like that we are not emitting them now in SymbiFlow.

We likely need a fasm2frames fix for the ALWAYS_ON_X.  We never understood them, but they still need to be emitted.

> 
> And for the counter test I get Vivado error:
> 
> ```
> ERROR: [Place 30-574] Poor placement for routing between an IO pin and BUFG. If this sub optimal condition is acceptable for this design, you may use the CLOCK_DEDICATED_ROUTE constraint in the .xdc file to demote this message to a WARNING. However, the use of this override is highly discouraged. These examples can be used directly in the .xdc file to override this clock rule.
>         < set_property CLOCK_DEDICATED_ROUTE FALSE [get_nets RIOB33_X31Y75_IOB_X0Y76_I] >
> 
>         RIOB33_X31Y75_IOB_X0Y76_IBUF (IBUF.O) is locked to IOB_X0Y76
>         bufgCLK_BUFG_BOT_R_X82Y48_BUFGCTRL_X0Y0_BUFGCTRL (BUFGCTRL.I0) is locked to BUFGCTRL_X0Y0
> ```
> 
> When I suppress that error it shows that the clock is routed to a BUFG in different top/bottom portion of the fabric than the clock IOB is located in. And the clock tree is not used for that route.

This should be solvable with fixing the site location of the BUFG via LOC once capacity is available.  FYI @acomodi .

mkurc-ant:
@litghost 
> We likely need a fasm2frames fix for the ALWAYS_ON_X. We never understood them, but they still need to be emitted.

What is the correct way of detecting what part is the fasm2frames run for? It takes the database path either from an env. vairable or an argument. Shall I assume that if that path contains `zynq7` than it is for Zynq ?

litghost:
> @litghost
> 
> > We likely need a fasm2frames fix for the ALWAYS_ON_X. We never understood them, but they still need to be emitted.
> 
> What is the correct way of detecting what part is the fasm2frames run for? It takes the database path either from an env. vairable or an argument. Shall I assume that if that path contains `zynq7` than it is for Zynq ?

Good question.  I suggest we make a new database file that is simply called "required_features" or some such.  fasm2frames can query it and handle it.

mkurc-ant:
https://github.com/SymbiFlow/prjxray/pull/1169

litghost:
> > And for the counter test I get Vivado error:
> > ```
> > ERROR: [Place 30-574] Poor placement for routing between an IO pin and BUFG. If this sub optimal condition is acceptable for this design, you may use the CLOCK_DEDICATED_ROUTE constraint in the .xdc file to demote this message to a WARNING. However, the use of this override is highly discouraged. These examples can be used directly in the .xdc file to override this clock rule.
> >         < set_property CLOCK_DEDICATED_ROUTE FALSE [get_nets RIOB33_X31Y75_IOB_X0Y76_I] >
> > 
> >         RIOB33_X31Y75_IOB_X0Y76_IBUF (IBUF.O) is locked to IOB_X0Y76
> >         bufgCLK_BUFG_BOT_R_X82Y48_BUFGCTRL_X0Y0_BUFGCTRL (BUFGCTRL.I0) is locked to BUFGCTRL_X0Y0
> > ```
> > 
> > 
> > When I suppress that error it shows that the clock is routed to a BUFG in different top/bottom portion of the fabric than the clock IOB is located in. And the clock tree is not used for that route.
> 
> This should be solvable with fixing the site location of the BUFG via LOC once capacity is available. FYI @acomodi .

Let's disable the DIFF FASM test and add a TODO pointing to https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1183 .  It will be a bit before the capacity issue is resolved.

mkurc-ant:
Needs database build from b3fea9c07b7ac37516d0a9cd0fc1b4d4e4a58be5 or later for the CI to pass.

litghost:
Latest sync PR's are merged in prjxray, just waiting for CI to computing on the commit.  After that a new prjrxray-db can be pushed.

PR #1216, opened 2019-12-10
Fix site pin switch when connecting pip is in another tile.

litghost:
This fixes the `IDELAYCTRL.RDY` and `IDELAYCTRL.RST`graph connections.


litghost:
> Have you double-checked whether the routing error still appears when connecting the RDY or RST pins?

I have not.

litghost:
> LGTM. Have you double-checked whether the routing error still appears when connecting the RDY or RST pins?

Worked!  I'll cook up a PR to enable diff fasm testing on IDELAY!

PR #1220, opened 2019-12-11
Mini Litex with DDR test

tmichalak:
This PR adds a minimal LiteX configuration for Arty and is featuring a Vexriscv CPU in the lite variant, DDR memory controller and uart.
In order to minimize the timing violations impact the sys_clk has been decreased to 50MHz.


litghost:
Looks like an OSERDES packing failure?

acomodi:
@litghost yes, `IOBUFTDS_M/S` modes lack of the correct pack_pattern. I think by modifying the https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/utils/add_pack_patterns.py script and adding the right patterns, this issue should be solved.

tmichalak:
Yes, I talked to @acomodi about it, I will add the pack pattern and see what else comes up.

xc7/tests/soc/litex/mini_ddr/scripts/min_ddr_arty.py
```diff
@@ -0,0 +1,103 @@
...
+            "BUFG", i_I=platform.request("clk100"), o_O=pll_clkin
+        )
+        self.submodules.pll = pll = S7PLL(speedgrade=-1)
+        self.comb += pll.reset.eq(~platform.request("cpu_reset"))
+        pll.register_clkin(pll_clkin, 100e6)
+        pll.create_clkout(self.cd_sys, sys_clk_freq)
+        pll.create_clkout(self.cd_sys4x, 4 * sys_clk_freq)
+        pll.create_clkout(self.cd_sys4x_dqs, 4 * sys_clk_freq, phase=90)
+        pll.create_clkout(self.cd_clk200, 200e6)
+        self.submodules.idelayctrl = S7IDELAYCTRL(self.cd_clk200)
```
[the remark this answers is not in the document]
acomodi: An additional thing would be to add the output to the idelayctrl (RDY signal), as currently VPR still removes it.

It could simply be directed to a led

mithro:
Just FYI - DDR generally has a "minimum" speed it has to operate at. It might not be possible to operate the DDR at only 200 MHz.

mkurc-ant:
After my recent attempts to nail down the DDR problem the design ceased to work at all. It looks like its skipping bytes sent through the UART:

```
s
1:X --= ==-  0M SRB
                     RB

                       AM...
izing SDwcoli00000000000000| delays00000000000000| delays:0000000000000| delays:0000000000000| delays:0000000000000| delays:0000000000000| delays: 0000000000000 delays: 0000000000000 dela b0 dela0000000000000000| delay000000000000000| delay000000000000000| delay000000000000000| delays00000000000000| delays:0000000000000| delays:0000000000000| delays: 0000000000000 delayb0 deowcontrol
rdwae56/256 errors
25m:rrors
```

While the same verilog file when used with vendor tools works flawlessly.

xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+  INPUT_XDC_FILE minilitex_ddr_arty.xdc
+  SDC_FILE arty.sdc
+)
+
+# TODO: Re-enable the diff_fasm test once the following issue is done
+# https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1183
+add_vivado_target(
+    NAME minilitex_ddr_arty_vivado
+    PARENT_NAME minilitex_ddr_arty
+    DISABLE_DIFF_TEST
```
litghost: Please uncomment `DISABLE_DIFF_TEST`, as I expect it should be passing now.

mkurc-ant: I will, but there are at least 2 issues with fasm2bels that will make it fail. One is IOI_SING wire naming, another is related to VREF and IN_TERM decoding. That PR is attempt to fix that https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1257

mkurc-ant:
@litghost Yes, It should. But for now there is no mechanism for that.

`IOSTANDARD`, `DRIVE` and `SLEW` can be defined either in verilog or XDC. There is the Yosys plugin that reads the XDC and sets these parameters on IO buffer cells within Yosys. One option I've thought of it to have another Yosys plugin that would extract this information from IO buffers prior to techmapping and store this information in the JSON file. Another way is to parse XDC in fasm2bels but that would not allow to get parameters defined in verilog. We can also keep original parameters during techmapping and pass them to the BLIF file (or convert to cell attributes). Then fasm2bels will be able to read them. Other ideas?

litghost:
> One option I've thought of it to have another Yosys plugin that would extract this information from IO buffers prior to techmapping and store this information in the JSON file

Why not just have the techmap's pass through the full parameter to the eblif?  That way we can simply read the eblif or convert the eblif to JSON.  Either change should be pretty straight forward.

mithro:
From @litghost;
> Why not just have the techmap's pass through the full parameter to the eblif?

This sounds like the best idea to me, any reason we should do it this way?

acomodi:
By rebasing on top of master, the feasibility filter can be set to `on`. There were some missing pack_patterns that prevented the packer to correctly choose the correct sites combination.

By enabling the filter, the packer now takes ~10x less time.

acomodi:
@litghost I could get Vivado to arrive at the end of the routing phase. It ended up with two partial antennas. Moreover I discovered a bug in fasm2bels for which, in some cases, IDELAYS were not instantiated in the resulting `top_bit.v`.

Here there is a link to a `.dcp`, `top_bit.v` and `top_bit.v.tcl` files tarball: https://drive.google.com/open?id=16uPVqpJdGvx0DXWXs_IrkyiK8_wdoSrq

acomodi:
@litghost By twitching the sdc file for arty, I have been able to complete fasm2bels, generating a bistream. The antenna nets issue has disappeared. 

I noticed that the IPAD pin for the 100 MHz clock is being routed through general interconnect instead of going through the CCIO pip, reaching the dedicated route.
This could be an additional reason for the DDR not properly working, as this situation would most probably mess with timing information.

xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    minilitex_ddr_arty.v
+    VexRiscv_Lite.v
+  INPUT_IO_FILE arty.pcf
+  INPUT_XDC_FILE minilitex_ddr_arty.xdc
+  SDC_FILE arty.sdc
+)
+
+add_vivado_target(
+    NAME minilitex_ddr_arty_vivado
+    PARENT_NAME minilitex_ddr_arty
```
litghost: Need XDC constraints to get valid timing analysis out of Vivado, see https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1344

xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+add_file_target(FILE mem.init)
+add_file_target(FILE mem_1.init)
+
+add_file_target(FILE arty.sdc)
+add_file_target(FILE arty.json) # IOSTANDARD defs for fasm2bels
```
litghost: I don't believe the arty.json file is required anymore, delete it?

xc7/tests/soc/litex/mini_ddr/arty.sdc
```diff
@@ -0,0 +1,13 @@
+create_clock -period 4.166 sys4x_clk__main_clkout_buf1 -waveform {0.000 2.083}
```
litghost: Can we modify the input verilog to avoid net aliases and these strange net names for SDC constraints?

acomodi: Done.

The only clock that cannot be changed is the system clock. This because it gets renamed as it enters in submodules. I need to see how to overcome this issue.

I have also added the clk200 which wasn't there and is necessary to correctly place&route the FDPE flipflops for the reset signal.

xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
acomodi: Done

xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
acomodi: Done

xc7/tests/soc/litex/mini_ddr/arty.sdc
litghost: The solution may have to be fixing VPR.  Do have time to take a crack at that?

acomodi: Sure, I can take a look into that. We should be able to use aliases or sth like this for clock net names, to avoid this situation.

PR #1229, opened 2019-12-14
Implement pb_array element to allow moving num_pb out of pb_type

HackerFoo:
Uses XSLT to transform
```xml
<pb_array num_pb="N">
   <pb_type ...>
```
to
```xml
<pb_type num_pb="N" ...>
```
to address https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/272

### DIFF
--- a/common/xml/convert-pb_type-attributes.xsl
+++ b/common/xml/convert-pb_type-attributes.xsl
@@ -14,6 +14,10 @@
       <xsl:if test="blif_model">
         <xsl:attribute name="blif_model"><xsl:value-of select="blif_model/text()"/></xsl:attribute>
       </xsl:if>
+      <!-- Inherit 'num_pb' attribute from pb_array elements -->
+      <xsl:if test="parent::pb_array/@num_pb">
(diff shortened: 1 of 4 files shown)


HackerFoo:
See the test for examples. It works with includes.

mithro:
These transforms are in the process of being extracted to https://github.com/SymbiFlow/vtr-xml-utils

@kgugala + @glatosinski how close are we to finishing that?

mithro:
You will also need to update the schema I think?

glatosinski:
All XSLT transforms are extracted to https://github.com/SymbiFlow/vtr-xml-utils

mithro:
@glatosinski they need to be removed from the original location.

HackerFoo:
@mithro @glatosinski Should I move this PR then? `vtr-xml-utils` isn't in https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/.gitmodules yet.

glatosinski:
@HackerFoo I will remove the content of vtr-xml-utils from symbiflow-arch-defs and add this as submodule, so you can move this PR.

HackerFoo:
Moved PR to https://github.com/SymbiFlow/vtr-xml-utils/pull/8

PR #1240, opened 2019-12-17
Zynq PS7 support

mkurc-ant:
Adds support for PS7 to SymbiFlow. The PS7 does not have any config bits. It can be directly instantiated in Verilog.


acomodi:
@mkurc-ant I believe this is good to merge?

mkurc-ant:
@acomodi I believe so.

PR #1234, opened 2019-12-16
xc7: tests: reformat tests to use common pcf files

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

This PR is to clean all the xc7 tests, by using the common pcf files.

Some of the tests need a custom `.pcf` file, but the majority of them can adopt the one found in `xc7/tests/common`

PR #1248, opened 2019-12-20
soc tests: added sdc file to picosoc and murax on 50T

acomodi:

This makes picosoc route in ~30 seconds and murax in ~18 seconds

### DIFF
--- a/tests/9-soc/murax/CMakeLists.txt
+++ b/tests/9-soc/murax/CMakeLists.txt
@@ -42,6 +42,7 @@ add_fpga_target(
     basys3_toplevel_no_roi.v
     Murax.v
   INPUT_IO_FILE basys3.pcf
+  SDC_FILE basys3.sdc
   EXPLICIT_ADD_FILE_TARGET
 )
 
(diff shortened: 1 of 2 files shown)


mithro:
Should the arty also get an sdc file too?

acomodi:
@mithro the `arty.sdc` is not there yet, but I think right now it is not necessary yet, as for now, all arty tests (counter and buttons) still use the `ROI`.

With https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1234 I am adding all the constraints file in a common folder and I will also add the `arty.sdc`.

If this goes green it would be good to merge it as it will most probably speed-up CI builds

mithro:
The vendor tool tests failed with;
```
	Clock Rule: rule_bufh_bufr_ramb
	Status: PASS
	Rule Description: Reginal buffers in the same clock region must drive a total number of brams less
	than the capacity of the region
	CLK_HROW_TOP_R_X60Y78_BUFHCE_X1Y20_BUFHCE (BUFHCE.O) is locked to BUFHCE_X1Y20

ERROR: [Place 30-99] Placer failed with error: 'IO Clock Placer failed'
Please review all ERROR, CRITICAL WARNING, and WARNING messages during placement to understand the cause for failure.
ERROR: [Common 17-69] Command failed: Placer could not place all instances
```

PR #1252, opened 2019-12-27
Add make install target

kgugala:
This PR adds make install target allowing to generate binary version of the Symbiflow toolchain.

Currently it emits all the required binaries and scripts. Architecture definitions files are missing (thus WIP status)


kgugala:
kokoro CI fails because we use quite old cmake (3.7). In the install function I use install(TARGETS) for target defined in a subdirectory. This seems to be supported from cmake 3.13 (https://stackoverflow.com/questions/34443128/cmake-install-targets-in-subdirectories). 

@mithro can we update cmake in kokoro? Or should I look for workaround?
(I have tested the arch-defs with cmake 3.13.4 on Debian 10)

mithro:
@kgugala Does the cmake file specify a minimum version?

kgugala:
> @kgugala Does the cmake file specify a minimum version?
right now it says 3.3, but it's too old for the install target. 

In kokoro we install cmake with apt. Switching to pip may help - pip has way newer version. If we can update cmake in kokoro I'll bump the minimal required version in cmake file

mithro:
Should be fixed in https://github.com/SymbiFlow/prjxray/pull/1191 ?

kgugala:
Install target emits all the required files, but `pinmap.csv`. The problem here is that we generate this file per board, not per device type. Since we switched to non-roi devices we can generate `pins.csv` file for every device type instead of board.

kgugala:
all the required files seem to be installed correctly. I think this PR is ready for review.

mithro:
@kgugala Can we test this on CI somehow?

litghost:
> @kgugala Can we test this on CI somehow?

As a concrete example of what we should do:
1. Run the make install target to a folder
2. Run a simple Makefile that generates a simple bitstream (e.g. counter) using the outputs
3. (optional) Convert the bitstream back to a verilog file and run a testbench against the design.

litghost:
> Install target emits all the required files, but `pinmap.csv`. The problem here is that we generate this file per board, not per device type. Since we switched to non-roi devices we can generate `pins.csv` file for every device type instead of board.

The current strategy is correct (ish).  The 50T family of designs share the arch/rrgraph, but need a different `pinmap.csv`, as only the package is different.  If you move the `pinmap.csv` to the device, you would generate multiple rrgraphs per 50T class device (I believe).

Can you detail what you suggestion is?

xc7/make/install.cmake
```diff
@@ -0,0 +1,131 @@
...
+  set(VPR_COMMON_TEMPLATE "${symbiflow-arch-defs_SOURCE_DIR}/xc7/toolchain_wrappers/vpr_common")
+  set(VPR_COMMON "${CMAKE_CURRENT_BINARY_DIR}/vpr_common")
+  configure_file(${VPR_COMMON_TEMPLATE} "${VPR_COMMON}" @ONLY)
+
+  install(FILES ${TOOLCHAIN_WRAPPERS} ${VPR_COMMON}
+          DESTINATION bin
+          PERMISSIONS WORLD_EXECUTE WORLD_READ OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE)
+
+  # install binaries
+  install(FILES ${YOSYS_BINS} ${VPR} ${GENFASM}
```
litghost: Yosys also needs to share directory, the binary isn't the complete thing

xc7/make/install.cmake
```diff
@@ -0,0 +1,131 @@
...
+  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/utils/lib/parse_pcf.py
+          DESTINATION bin/python/lib)
+
+  # install yosys data
+  install(DIRECTORY ${YOSYS_DATADIR}
+          DESTINATION share)
+
+  # install prjxray techmap
+  install(DIRECTORY ${symbiflow-arch-defs_SOURCE_DIR}/xc7/techmap
+          DESTINATION share/prjxray)
```
litghost: The techmap should be kept seperate from the the rest of prjxray

xc7/yosys/install_synth.tcl
```diff
@@ -0,0 +1,43 @@
+yosys -import
```
litghost: Copying and pasting the xc7/synth.tcl is not accepted.  Please find a way to have one script

kgugala: fixed

kgugala:
> How about having TECHMAP_PATH be a property of the device? And then setting the TECHMAP_PATH in the xc7 add_device location?

That's a good idea. Do we have to set it per device? Maybe we should define it in the ARCH? We actually do it for cells sim https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc7/make/arch_define.cmake#L22

litghost:
> Maybe we should define it in the ARCH?

Sounds good to me.

xc7/toolchain_wrappers/synth
```diff
@@ -0,0 +1,65 @@
...
+			;;
+	esac
+done
+
+if [ ${#VERILOG_FILES[@]} -eq 0 ]; then
+	echo "Please at least one Verilog file"
+	exit 1
+fi
+
+export INPUT_XDC_FILE=${XDC_FILES[*]}
```
litghost: Missing export of YOSYS_TECHMAP?  This is why adding a test case is important.

kgugala: the script exports `TECHMAP_PATH` here https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1252/files#diff-060a92a8b914fe0d8c57eef57705f9daR6

this variable is used in the tcl https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1252/files#diff-3a4d31cef2659b1011b54cd37fa3edc0R10

> This is why adding a test case is important.

I agree, I'm adding the tests

.github/kokoro/xc7.sh
```diff
@@ -30,4 +30,17 @@ echo "----------------------------------------"
 	popd
 )
 echo "----------------------------------------"
+
+echo
+echo "========================================"
+echo "Running installed toolchain tests"
+echo "----------------------------------------"
+(
+	sudo pip install git+https://github.com/SymbiFlow/fasm.git
```
litghost: There is never a justification for using ``sudo`` here.  Either depend on installing it to the local python cache, or make an explicit python virtualenv

xc7/tests/install_test/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	route -e ${TOP}.eblif -d ${DEVICE}
+
+${TOP}.fasm: ${TOP}.route
+	write_fasm -e ${TOP}.eblif -d ${DEVICE}
+
+${TOP}.bit: ${TOP}.fasm
+	write_bitstream -d ${BITSTREAM_DEVICE} -f ${TOP}.fasm -b ${TOP}.bit
+
+clean:
+	rm -f *.rpt
```
litghost: Rather than removing lots of files, can you put the files in a build directory, and then just delete that?

.github/kokoro/xc7.sh
kgugala: fixed

xc7/tests/install_test/Makefile
kgugala: fixed

xc7/make/install.cmake
kgugala: @litghost where do you think this should be placed? Those are xc7/VPR specific techmaps. Shall we move them to `share/yosys/xc7_vpr` or maybe to `share/xc7_vpr`? Or maybe sth else? What do you think will be better?

litghost: How about `share/techmaps/xc7`?

kgugala:
Install fails with `yosys: error while loading shared libraries: libreadline.so.7: cannot open shared object file: No such file or directory`

This can be fixed with setting `LD_LIBRARY_PATH` to `build/env/conda/lib`, but it doesn't feel a right thing to do. Maybe we should build Yosys in the CI? Or include the libraries in installation folder and set `LD_LIBRARY_PATH` in scripts?

@litghost @mithro what do you think

litghost:
> Install test fails with `yosys: error while loading shared libraries: libreadline.so.7: cannot open shared object file: No such file or directory`
> 
> This can be fixed with setting `LD_LIBRARY_PATH` to `build/env/conda/lib`, but it doesn't feel a right thing to do. Maybe we should build Yosys in the CI? Or include the libraries in installation folder and set `LD_LIBRARY_PATH` in scripts?
> 
> @litghost @mithro what do you think

Grrr, readline has been a bane for a while.  I think we should modify the yosys conda package to statically link readline.

kgugala:
> > Install test fails with `yosys: error while loading shared libraries: libreadline.so.7: cannot open shared object file: No such file or directory`
> > This can be fixed with setting `LD_LIBRARY_PATH` to `build/env/conda/lib`, but it doesn't feel a right thing to do. Maybe we should build Yosys in the CI? Or include the libraries in installation folder and set `LD_LIBRARY_PATH` in scripts?
> > @litghost @mithro what do you think
> 
> Grrr, readline has been a bane for a while. I think we should modify the yosys conda package to statically link readline.

I like this option

mithro:
This is probably caused by libreadline being picked up from the system being compiled on rather than from conda. I'll take another look at it.

mithro:
@kgugala It looks like you are installing Yosys as part of `make install`?
```
[534/534] Install the project...
-- Install configuration: "Release"
-- Installing: /tmpfs/src/install/bin/env
-- Installing: /tmpfs/src/install/bin/generate_constrains
-- Installing: /tmpfs/src/install/bin/pack
-- Installing: /tmpfs/src/install/bin/place
-- Installing: /tmpfs/src/install/bin/route
-- Installing: /tmpfs/src/install/bin/synth
-- Installing: /tmpfs/src/install/bin/write_bitstream
-- Installing: /tmpfs/src/install/bin/write_fasm
-- Installing: /tmpfs/src/install/bin/vpr_common
-- Installing: /tmpfs/src/install/bin/yosys
-- Installing: /tmpfs/src/install/bin/yosys-abc
-- Installing: /tmpfs/src/install/bin/yosys-smtbmc
-- Installing: /tmpfs/src/install/bin/yosys-filterlib
-- Installing: /tmpfs/src/install/bin/yosys-config
```

mithro:
@kgugala Where does the Yosys you are installing come from?

.github/kokoro/xc7.sh
```diff
@@ -2,8 +2,9 @@
 
 SCRIPT_SRC="$(realpath ${BASH_SOURCE[0]})"
 SCRIPT_DIR="$(dirname "${SCRIPT_SRC}")"
+INSTALL_DIR="`pwd`/install"
```
mithro:
```suggestion
INSTALL_DIR="$(pwd)/install"
```

xc7/toolchain_wrappers/generate_constrains
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
mithro: Please use `$(XXX)`

kgugala:
> @kgugala Where does the Yosys you are installing come from?

When invoking `make install` yosys (and all it's files) used in current CMake configuration will be installed. By default it will be Yosys from conda.

.github/kokoro/xc7.sh
kgugala: fixed

xc7/toolchain_wrappers/generate_constrains
kgugala: done

mithro:
You shouldn't be installing anything from conda? The installed files should be what ends up inside a conda package which depends on the Yosys + VtR + other related packages....

kgugala:
> The installed files should be what ends up inside a conda package which depends on the Yosys + VtR + other related packages....

This is exactly what is installed

mithro:
Nothing should be copied from the `env` or `conda` directories. The `make install` target should only install the files created by `symbiflow-arch-defs`.

kgugala:
> Nothing should be copied from the `env` or `conda` directories. The `make install` target should only install the files created by `symbiflow-arch-defs`.

Should then `make install` build vtr and yosys separately?

kgugala:
> > Nothing should be copied from the `env` or `conda` directories. The `make install` target should only install the files created by `symbiflow-arch-defs`.
> 
> Should then `make install` build vtr and yosys separately?

actually this may be a little bit problematic. Arch definition from project which is being installed may not work with the HEAD of `master+wip` (either in Yosys and VtR) and we don't have info which commit of the tools repos should be used with installed arch-defs. The only known-to-work version of the tools is the one from conda.

kgugala:
I defined ctest target for `make install` test. It should now set all the required env, so libs should no longer be a problem in kokoro.

However running `make test` also runs tests defined in https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/cmake/devices.cmake#L1768 and https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/cmake/devices.cmake#L1777. There is a lot of them (621) and they seem to be really long. @litghost did we run them before in CI?

kgugala:
TBH it looks like the `0-const_dummy_ice40_hx1k_tq144_check` test hanged on my local machine

mithro:
> > Nothing should be copied from the `env` or `conda` directories. The `make install` target should only install the files created by `symbiflow-arch-defs`.
> 
> Should then `make install` build vtr and yosys separately?

Yes. vtr and yosys should be compiled and installed using their normal flows.

litghost:
> However running `make test` also runs tests defined in

For now invoke "make test" on the test we care about, and let's file an issue to fix the other tests.

common/cmake/install.cmake
```diff
@@ -0,0 +1,72 @@
...
+  get_target_property_required(HAS_PLACE_DELAY ${DEVICE} "${PACKAGE}_HAS_PLACE_DELAY_CACHE")
+  if(${HAS_LOOKAHEAD})
+    get_target_property(LOOKAHEAD_FILE ${DEVICE} ${PACKAGE}_LOOKAHEAD_FILE)
+    list(APPEND INSTALL_FILES ${LOOKAHEAD_FILE})
+  endif()
+  if(${HAS_PLACE_DELAY})
+    get_target_property(PLACE_DELAY_FILE ${DEVICE} ${PACKAGE}_PLACE_DELAY_FILE)
+    list(APPEND INSTALL_FILES ${PLACE_DELAY_FILE})
+  endif()
+  get_target_property_required(RR_GRAPH_FILE ${DEVICE} ${PACKAGE}_OUT_RRXML_REAL)
```
litghost: This should probably be `${PACKAGE}_OUT_RRBIN_REAL` now
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/cmake/devices.cmake#L661

kgugala:
> > Install target emits all the required files, but `pinmap.csv`. The problem here is that we generate this file per board, not per device type. Since we switched to non-roi devices we can generate `pins.csv` file for every device type instead of board.
> 
> The current strategy is correct (ish). The 50T family of designs share the arch/rrgraph, but need a different `pinmap.csv`, as only the package is different. If you move the `pinmap.csv` to the device, you would generate multiple rrgraphs per 50T class device (I believe).
> 
> Can you detail what you suggestion is?

here is the tree of the install folder for 50t family
```
xc7a50t_test/
├── arch.timing.xml
├── channels.db
├── rr_graph_xc7a50t_test.lookahead.bin
├── rr_graph_xc7a50t_test.place_delay.bin
├── rr_graph_xc7a50t_test.rr_graph.real.patched.bin
├── xc7a35tcpg236-1
│   └── pinmap.csv
└── xc7a35tcsg324-1
    └── pinmap.csv
```

xc7/make/install.cmake
kgugala: done

litghost:
> rr_graph_xc7a50t_test.rr_graph.real.patched.bin

It is better to take `rr_graph_xc7a50t_test.rr_graph.real.bin` as that capnp ensures that edges are sorted correctly.

kgugala:
> > rr_graph_xc7a50t_test.rr_graph.real.patched.bin
> 
> It is better to take `rr_graph_xc7a50t_test.rr_graph.real.bin` as that capnp ensures that edges are sorted correctly.

fixed

kgugala:
CI is green and I believe all the comments were addressed. @mithro @acomodi can we merge this one?

acomodi:
LGTM. Merging now

PR #1242, opened 2019-12-18
xc7: added RL_IOI_SING tile pips correct fasm_feature

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

This PR solves the issue for which the pips related to the IOI_SING tiles are incorrect, as they should refer to the correct half of a full IOI tile.

With these changes, the IOI_SING pip features get modified according to the correct IOI half they belong to.

Example of fasm feature modification:

```
RIOI3_SING_X43Y49.IOI_OCLKM_0.IOI_SING_LEAF_GCLK5 --> RIOI3_SING_X43Y49.IOI_OCLKM_1.IOI_LEAF_GCLK5 
```

PR #1257, opened 2020-01-02
Fixes for fasm2bels

mkurc-ant:
This PR fixes handling of some pips/wires that are named differently for `[LR]IOI3_SING` tiles. Basically undoes wire renaming in fasm2bels flow that was introduced by https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1242

Also fixes bugs with setting of the `IN_TERM` parameter on top-level ports.

### DIFF
--- a/xc7/fasm2bels/connection_db_utils.py
+++ b/xc7/fasm2bels/connection_db_utils.py
@@ -1,4 +1,24 @@
+import re
 import functools
+
+# A map of wires that require "SING" in their name for [LR]IOI3_SING tiles.
+IOI_SING_WIRES = {
+    "IOI_IOCLK0": "IOI_SING_IOCLK0",
+    "IOI_IOCLK1": "IOI_SING_IOCLK1",
+    "IOI_IOCLK2": "IOI_SING_IOCLK2",
+    "IOI_IOCLK3": "IOI_SING_IOCLK3",
+    "IOI_LEAF_GCLK0": "IOI_SING_LEAF_GCLK0",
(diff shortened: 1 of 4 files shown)


xc7/fasm2bels/connection_db_utils.py
```diff
@@ -14,6 +33,22 @@ def get_tile_type_pkey(tile):
 
     @functools.lru_cache(maxsize=None)
     def maybe_get_wire(tile, wire):
```
litghost: This seems weird.  What is generating the wrong wire names?

mkurc-ant: With the https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1242 some wire renaming was introduced. It is done during graph edge annotation with fasm features. The thing with `IOI_SING` tiles is that the "top" and "bottom" SING tile of a column uses different subsets of IOI tile bits and that is why it is needed. Moreover, some wires equivalent to those in "normal" IOIs have "SING" in their names. But since only the fasm features are renamed, not wires in the database, in order to successfully decode `IOI_SING` routing those wires need to be renamed back.

In order to avoid all of that we would probably need to create different tile types for "top" and "bottom" SINGs.

litghost: So the bad names are coming from the FASM features?  Let's fix this at the FASM level then, rather than here.

mkurc-ant: Ok, that makes sense. Done.

Issue #1239, opened 2019-12-17
Feature CLK_HROW_R_CK_GCLKx_ACTIVE not emitted


mkurc-ant:
I noticed that we emit these features only when a particular output of a `BUFHCE` is in use. But we do not emit them for CLK_HROW tiles in which no `BUFHCE` is used and there is an active clock wire passing through them.

Consider a design for xc7a50t where the `CLK_HROW_TOP_R_X60Y130` tile has active `BUFHCE`s in it. But the clock signal is routed from a `BUFG` in `CLK_BUFG_BOT_R_X60Y48` tile through `CLK_HROW_TOP_R_X60Y78`. The "bottom" CLK_HROW tile does not have anything active inside.

In such a design Vivado emits a `CLK_HROW_R_CK_GCLKx_ACTIVE` feature for the `CLK_HROW_TOP_R_X60Y78` tile while SymbiFlow don't. Anyways, the design works correctly in hardware.

PR #1237, opened 2019-12-17
[xc7] Add clock placer

acomodi:
This PR builds on top of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1232 as a fix for the python linter was needed.

With prjxray ad prjxray-db updated to latest version, containing the clock_region addition, this PR should now go green.

PR #1262, opened 2020-01-10
fasm2bels IO settings from EBLIF

mkurc-ant:
This PR adds the functionality to fasm2bels that reads `IOSTANDARD` and `DRIVE` parameters from EBLIF file and applies them to IO buffers.

The mapping between EBLIF cells and IO buffers decoded from fasm is done through top-level port names and locations read from a PCF file. Therefore both EBLIF and PCF files need to be provided.


mithro:
fasm has support for annotations -> https://github.com/SymbiFlow/fasm/blob/master/docs/specification.rst#annotations -- these type of things are probably a good use of them?

mkurc-ant:
@mithro But that won't help much. The problem is that the actual IOSTANDARD and DRIVE cannot be unambiguously decoded from the fasm / bitstream. And since the fasm is created from the bitstream there is nothing reliable to annotate it with. What we need is a way to pass the information from the design to fasm2bels so it can set these parameters accordingly.

Right now it is a separate JSON file. But as fasm2bels can use eblif for other things we can pass the parameters in there.

utils/eblif.py
```diff
@@ -8,8 +8,8 @@
 ]
 
 sub_level = [
-    "cname",
     "attr",
+    "param",
```
litghost: Can you explain this change?  Maybe add a comment to global explaining what is going on here?  This change seems suspicious, but presumably there is a good reason for it.

xc7/fasm2bels/iob_models.py
```diff
@@ -94,53 +94,52 @@ def append_obuf_iostandard_params(
...
-            for i, d, s in possible_iostandards:
-                eprint(
-                    " {}| {}| {}|".format(
-                        i.ljust(18),
-                        str(d).ljust(7), s.ljust(5)
-                    )
-                )
-            eprint("")
+        # Demote NSTD-1 to warning
+        top.disable_drc("NSTD-1")
```
litghost: Under what circumstances should the code ever reach this path?  Why is this not a failure?

xc7/fasm2bels/iob_models.py
```diff
@@ -159,38 +158,42 @@ def append_ibuf_iostandard_params(
...
-        # Valid
-        else:
-            bel.parameters["IOSTANDARD"] = '"{}"'.format(iostandard)
+        eprint("Candidates are:")
+        for i in possible_iostandards:
+            eprint(" {}".format(i.ljust(15)))
+        eprint("")
+
+        # Demote NSTD-1 to warning
+        top.disable_drc("NSTD-1")
```
litghost: Under what circumstances should the code ever reach this path? Why is this not a failure?

xc7/fasm2bels/verilog_modeling.py
```diff
@@ -1245,12 +1248,110 @@ def __init__(self, db, grid, conn, name="top"):
...
+    def make_iosettings_map(self, parsed_eblif):
+        """
+        Builds map of top-level port names to cell names and cell names to IO
+        # settings map using data from a parsed EBLIF file.
+        """
+
+        # Map of EBLIF cell types of IO buffers to their ports connecting them
+        # with IO pads.
+        IOBUF_PORTS = {
+            "IBUF_VPR": ["I"],
```
litghost: There should **not** be an assumption that the input bitstream came from VPR.  This map should not be contained with `fasm2bels` at all, but a tool in front of it.  That tool can assume VPR.  That tool can then be adapted for nextpnr or Vivado itself as needed.

xc7/tests/iosettings/generate.py
```diff
@@ -68,10 +67,10 @@
                     ("P18", "R18"),
                 ]
         },
-    "arty":
+    "arty-full":
         {
             "BUFG":
-                "BUFGCTRL_X0Y0",
+                "BUFGCTRL_X0Y16",
```
litghost: There should no longer be a need to explicitly specify BUFG locations now that https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1237 is merged

xc7/fasm2bels/verilog_modeling.py
mkurc-ant: The assumption is about the eblif, not the bitstream. Since Vivado cannot write eblifs its reasonable. But now when I gave some thoughts to it I have a better idea.

xc7/fasm2bels/iob_models.py
mkurc-ant: Generally this happens when the IOSTANDARD and DRIVE parameters provided externally (via eblif or cmd line parameters) do not match possibilities that can be inferred from the fasm.

The condition will also be triggered when there is no external IOSTANDARD and DRIVE information provided.

I'm not sure whether this should be treated as an error. After all the design can be imported to Vivado for inspection even with some IOs do not have any IO standard assigned.

utils/eblif.py
mkurc-ant: First, the variable `sub_level` was not used at all in the code. Now it lists keywords that define parameters / attributes of a cell, which values are to be stored in a dictionary. I've added a comment to the code.

xc7/tests/iosettings/generate.py
mkurc-ant: Ok, fixed that.

mkurc-ant:
I've updated the PR and fixed some bugs as well. These are the IOSTANDARD test results I'm getting on my machine:

For `diff_output_arty-full_DIFF_SSTL135`, `output_arty-full_LVCMOS15`, `output_arty-full_LVCMOS25` and `output_arty-full_LVCMOS33`.
- Failure related to https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1239

For `inout_arty-full_LVCMOS18` and `output_arty-full_LVCMOS18`
- Incorrect handling of the `PUDC_B` special pin site. I've created an issue for that: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1263

The test for single ended SSTL135 results in unknown bits. See https://github.com/SymbiFlow/prjxray/issues/1182

The rest of the tests passes correctly with the IOSTANDARD and DRIVE passed from Yosys to EBLIF and decoded from it by fasm2bels.

utils/eblif.py
litghost: > the variable `sub_level` was not used at all in the code

That answers my questions :)

xc7/fasm2bels/iob_models.py
litghost: Still wondering about this?

xc7/fasm2bels/iob_models.py
litghost: In all cases, the IOSTANDARD needs to be supplied or explicit from the bitstream, yes?  In the event that an input from Vivado is supplied, it should have the right data?

xc7/fasm2bels/verilog_modeling.py
litghost: Ya, the new code looks much better

litghost:
I'm unclear, are the failing cases present and disabled, or simply not present?  Should we add the failing cases and comment them out?

mkurc-ant:
@litghost All of them are enabled. When the CI finishes I hope to see the same failures as on my local machine.

litghost:
Looks like some of the DRAM tests started failing, unclear why.  Can you replicate that?

mkurc-ant:
@litghost Yes. These failures happened for https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1234 in which PCFs were unified. I guess that there are some unused I/O pins in ROI targets. That causes broken routes in fasm2bels.

I'll investigate which targets are failing and why.

mkurc-ant:
It looks like the vivado targets for DRAM tests were never ran until #1234 was merged. I've checked randomly picked "stable" (merge) commits from between 6 months ago and now and those tests always failed.

xc7/fasm2bels/iob_models.py
litghost: Still don't understand why this is not an error.

litghost:
Can we disable the test that is failing because of  #1263 and re-enable it once that issue is fixed?

xc7/fasm2bels/iob_models.py
mkurc-ant: This is not an error (for fasm2bels) because if it was then the lack of external IOSTANDARD data would cause the fasm2bels flow to fail. Very often one wants to import a design to Vivado just to look at placement and routing, not necessarily generate a bitstream. Therefore I think that fasm2bels should print a warning that it's missing IOSTANDARD data, but not fail.

From the Vivado bitstream generation it is an error, correct IOSTANDARD has to be specified. If the information is not available to fasm2bels (no EBLIF file of the design given), then one has to supply it externally to Vivado somehow (via XDC ?).

I agree that maybe disabling `NSTD-1` in case of missing IOSTANDARD data is not a good idea. Thoughts?

Now that you mentioned it: When decoding a Vivado bitstream with fasm2bels there is no EBLIF to take IOSTANDARD from. I'm not sure how to supply that data to fasm2bels in this case. Through the XDC file of the original design? Do you have any other ideas?

mkurc-ant:
@litghost I've disabled the LVCMOS18 test which was failing due to #1263. The rest of failing ones should pass once https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1271 is merged.

xc7/fasm2bels/iob_models.py
litghost: But all bitstreams have an iostandard?  Therefore if the iostandard cannot be determined, that **is** a fatal error for fasm2bels

litghost: But all bitstreams have an iostandard?  Therefore if the iostandard cannot be determined, that **is** a fatal error for fasm2bels

litghost:
> @litghost I've disabled the LVCMOS18 test which was failing due to #1263. The rest of failing ones should pass once #1271 is merged.

Ok, so rebase on master and things should be green except for the previously failing vendor tests?  Please do so, and we'll see.

xc7/fasm2bels/iob_models.py
litghost: > IOSTANDARD data would cause the fasm2bels flow to fail. Very often one wants to import a design to Vivado just to look at placement and routing

But how can you be confident in fasm2bels unless it generates the same bitstream?

litghost: > Now that you mentioned it: When decoding a Vivado bitstream with fasm2bels there is no EBLIF to take IOSTANDARD from. I'm not sure how to supply that data to fasm2bels in this case. Through the XDC file of the original design? Do you have any other ideas?

This is a totally valid question, which we can file a feature request for and fix in a future version.

mkurc-ant:
Rebased.

mkurc-ant:
@litghost So now fasm diff fails only for the `murax` test. This is what we expect.
```
ERROR: [DRC RTSTAT-5] Partial antennas: 3 net(s) have a partial antenna. The problem bus(es) and/or net(s) are 
```

PR #1271, opened 2020-01-21
Fix incorrect emission of CLK_HROW_R_CK_GCLKx_ACTIVE features

mkurc-ant:
This PR targets to fix https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1239

A `CLK_HROW_R_CK_GCLKx_ACTIVE` feature has to be emitted even when a `CLK_HROW` tile does not have any active bels and a clock signal is just routed through it.

This is done by attaching `CLK_HROW_R_CK_GCLKx_ACTIVE` features to edges that trigger `*_ENABLE_ABOVE` and `_*ENABLE_BELOW` features. The additional annotation takes into account neighborhood of `CLK_BUFG_REBUF` tiles in such way that `CLK_HROW_R_CK_GCLKx_ACTIVE` is emitted only if there is a `CLK_HROW` tile in the nearest vicinity.


xc7/utils/prjxray_routing_import.py
```diff
@@ -111,7 +111,49 @@ def populate_bufg_rebuf_map(conn):
...
+    for i, tile_name in enumerate(rebuf_and_hrow_tiles):
+        if tile_name is not None and tile_name.startswith("CLK_BUFG_REBUF"):
+
+            rebuf_to_hrow_map[tile_name] = {
+                "above": maybe_get_clk_hrow(i - 1),
+                "below": maybe_get_clk_hrow(i + 1),
+            }
+
+
+# Find nodes touching rebuf wires.
```
litghost: Comment indent is wrong?

xc7/utils/prjxray_routing_import.py
```diff
@@ -111,7 +111,49 @@ def populate_bufg_rebuf_map(conn):
 
     cur = conn.cursor()
 
-    # Find nodes touching rebuf wires.
+    # Find CLK_HROW_TOP_R, CLK_HROW_TOP_R and REBUF tiles.
```
litghost: Please move the map construction to a function.  Consider writing a pytest for the function?

xc7/tests/gclk_active/gclk_through.v
```diff
@@ -0,0 +1,15 @@
+module top (
+  input  wire clk,
+  output wire out
+);
+
+// ============================================================================
+
+wire clk_g;
+
+(* LOC="BUFGCTRL_X0Y0" *)
```
litghost: Is this LOC really needed?  I suspect not?

mkurc-ant: It is to ensure that the test design triggers the condition that is meant to be testsed. You're right that in 99.9% of cases the placer will use correct BUFG site, but I want to eliminate that 0.1% in which it may not.

xc7/utils/prjxray_routing_import.py
mkurc-ant: Fixed.

xc7/utils/prjxray_routing_import.py
mkurc-ant: I moved it to a function. However developing a test is problematic as the function relies on data from the `channels.db`

PR #1267, opened 2020-01-17
Add penalty cost switch to avoid wrong routes

acomodi:
This PR adds a `delayfull` switch that has an intrinsic `fake` delay that forces the router to take other paths.
This should solve https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1153 and https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1019

PR #1279, opened 2020-01-24
Update arch-defs to use new db format for part-independent files

acomodi:

This PR is a follow-up of https://github.com/SymbiFlow/prjxray/pull/1196.

As the location of tilegrid and tileconn has changed, the same changes should reflect in arch-defs.


common/cmake/devices.cmake
```diff
@@ -12,6 +12,7 @@ function(DEFINE_ARCH)
   # ~~~
   # DEFINE_ARCH(
   #    ARCH <arch>
+  #    PART <part>
```
litghost: I don't agree that PART is part of ARCH?

common/cmake/devices.cmake
```diff
@@ -427,6 +434,7 @@ function(DEFINE_DEVICE)
   # DEFINE_DEVICE(
   #   DEVICE <device>
   #   ARCH <arch>
+  #   PART <part>
```
litghost: This is fine

xc7/archs/artix7/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
 project_xray_prepare_database(
-  PART artix7
+  ARCH artix7
+  PART xc7a35tcpg236-1
```
litghost: So while this does accommodate the 200T, this causes multiple copies of the arch/rrgraph/lookahead to be computed for graphs that are identical.  Keep in mind that the parts in the 35T share a graph, but have different pinouts (e.g. packaging).  It seems like a mistake for us to no re-use the arch/rrgraph/lookahead for parts that share a graph, but have different pinouts.  @mithro / @kgugala  How much do we care?

common/cmake/devices.cmake
acomodi: So, the main issue is that right now, most of the scripts do require the PART.
Right now, the PART is passed only when creating the board. The reason I assigned the PART to DEFINE_ARCH is that, for instance, artix 50T and 200T are actually different architectures with different structures (even though they share the db info).

This is untrue for the 35T which do actually share tilegrid and tileconn. My plan is to define only one arch for each one of these families, that will share rr_graph, arch.xml and all that is equal. While the pins and can be retrieved in the corresponding location in the db

litghost: So as a concrete example, there are 4 pinouts for the 200T, but likely only 1 RR graph.  It does not make sense to generate 4 lookaheads for the same graph, as this is very expensive.

Maybe `project_xray_prepare_database` should have the following arguments:

`PARTS <list of parts>`

and then the database generation picks a part when it needs too.  This also enables verification that each graph is in fact the same, as an optional follow on step.

In additional maybe an argument like
`DEFAULT_PART` or `EXAMPLE_PART`, where the documentation lists this part as the prototypical part for the family of parts with the same graph.

acomodi: With the `list of parts` you mean parts independent one to another (50T and 200T) or parts that are of the same "parts family"?

In addition I need to figure out how to pass the PART that needs to be fed as argument to the various scripts.

litghost: list of parts would be all parts with the same RR graph (e.g. all 50T).

> In addition I need to figure out how to pass the PART that needs to be fed as argument to the various scripts.

So I think there are two cases.  When generating the connection database, the `PROTOTYPICAL_PART` is sufficient.  For things like `fasm2frames`, IO placement, and `fasm2bels`, you probably want the part defined in `add_xc7_device`.

acomodi: I have adopted a slightly different approach.
Given that, for instance, the tile_import scripts require the PART, I have passed to `define_arch` a prototypical PART, which is common to each part of the family (all the artix7).

For the `prepare database`, instead, a list of sub-families are passed as a list (50T sub-family, 200T sub-family) and the generic channels.db is created for each of them.

This way we can define one device for each of these families, and a single rr_graph (and company) are going to be produced for that sub-family only.

When defining the boards, instead, we can use the available devices and they will use the correct package_pins files for the `after FASM` scripts.

acomodi:
@mithro This PR would need an update to the db after the changes merged with https://github.com/SymbiFlow/prjxray/pull/1196

common/cmake/devices.cmake
```diff
@@ -12,6 +12,7 @@ function(DEFINE_ARCH)
   # ~~~
   # DEFINE_ARCH(
   #    ARCH <arch>
+  #    PROTOTYPE_PART <rpototype_part>
```
litghost: Small spelling error here, `rpototype_part` vs `prototype_part`.

xc7/archs/artix7/devices/xc7a50t-bottom-virt/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 add_xc7_device_define_type(
   ARCH artix7
-  DEVICE xc7a50t-bottom
+ DEVICE xc7a50t-bottom
```
litghost: Missing a space?

xc7/make/project_xray.cmake
```diff
@@ -377,53 +213,233 @@ function(PROJECT_XRAY_PREPARE_DATABASE)
...
+  foreach(PART ${PROJECT_XRAY_PREPARE_DATABASE_PARTS})
+    file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/channels/${PART})
+    set(CHANNELS channels/${PART}/channels.db)
+    add_custom_command(
+      OUTPUT ${CHANNELS}
+      COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${PRJXRAY_DIR}:${symbiflow-arch-defs_SOURCE_DIR}/utils
+      ${PYTHON3} ${FORM_CHANNELS}
+      --db_root ${PRJXRAY_DB_DIR}/${ARCH}/
+      --part ${PROTOTYPE_PART}
+      --connection_database ${CMAKE_CURRENT_BINARY_DIR}/${CHANNELS}
```
litghost: Why are we generating a database per part?

acomodi: So, `PARTS` variable contains a list of parts sub-families (50T and 200T). As of my understanding, each sub-family needs its own channels.

litghost: Ok.

xc7/archs/artix7/devices/xc7a50t-bottom-virt/CMakeLists.txt
acomodi: Fixed

common/cmake/devices.cmake
acomodi: Fixed

acomodi:
Vendor tools finished, with all known failures:
- murax
- clock through general interconnect (still need https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1267 to be merged)

I'll proceed with the merge.

PR #1284, opened 2020-01-28
A script for the PS7 cell

mkurc-ant:
This is a script that reads the `ps7_ports.json` (meant to be in the prjxray database https://github.com/SymbiFlow/prjxray/pull/1170) and basing on it generates:

- XML model of the PS7 for the VPR
- XML pb_type of the PS7 for the VPR
- The PS7 cell verilog blackbox
- A techmap that converts `PS7` to `PS7_VPR` and ties all unconnected input ports to GND in the same way as the vendor tools do.

This script was used to generate all of the above that are present in arch defs right now. We might consider generating these instead of keeping it hard-coded in the future.

### DIFF
--- a/xc7/utils/prjxray_make_ps7_models.py
+++ b/xc7/utils/prjxray_make_ps7_models.py
@@ -15,33 +15,11 @@ import os
 # =============================================================================
 
 
-def main():
-
-    # Parse arguments
-    parser = argparse.ArgumentParser(
-        description=__doc__,
-        formatter_class=argparse.RawDescriptionHelpFormatter
-    )
-
-    parser.add_argument(
-        "json",
-        type=str,
-        help="Input JSON file with PS7 pins grouped into ports"
-    )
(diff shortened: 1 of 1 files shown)


xc7/utils/prjxray_make_ps7_models.py
```diff
@@ -0,0 +1,247 @@
...
+    args = parser.parse_args()
+
+    # Load ports
+    with open(args.json, "r") as fp:
+        ports = json.load(fp)
+
+    # .....................................................
+    # Generate XML model
+    pb_name = "PS7"
+    blif_model = "PS7_VPR"
```
litghost: Please, can we do a function per file?

mkurc-ant: Ok, I'll work on that

mkurc-ant: Done.

PR #1286, opened 2020-01-29
Convert from rrgraph XML to capnp during lookahead computation.

litghost:
The incremental XML writer we use during rr graph patching is actually
pretty memory efficent, so for now I don't see a reason to do anything
other than to write XML first, then convert to capnp using VPR.


litghost:
@acomodi / @HackerFoo This is green, please review?

litghost:
> Just to know, how much time is required to convert the rr_graph to capnp?

Because we do it during the lookahead/place delay computation, only the cost to write it out.

PR #1288, opened 2020-01-30
output reports after fasm2bels

acomodi:

This PR adds a change to output reports after routing is successful during fasm2bels.

### DIFF
--- a/xc7/utils/vivado_create_runme.py
+++ b/xc7/utils/vivado_create_runme.py
@@ -11,7 +11,6 @@ create_project -force -part {part} design design
 
 read_verilog {bit_v}
 synth_design -top {top}
-write_checkpoint -force design_{name}_pre_route.dcp
 source {bit_tcl}
 """.format(
             name=args.name,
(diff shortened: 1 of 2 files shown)


acomodi:
Apparently the clock through general interconnect is still randomly happening.
This could be due to the absence of constraint, I will try and see whether adding an sdc constraints file could solve the issue.
Probably a different and more robust solution than the penalty_cost could be setting max criticality to clock nets.

mithro:
I put my suggestion for solving the IO to BUFG issue in @acomodi's doc at https://docs.google.com/document/d/1D5uhrk4p0VDyY7lZthLR83TOv44rKOJNRC97f61Og3Q/edit?ts=5e3433ef#

acomodi:
@litghost Given that the `create_clock` is still missing in the constraint files, the timing_summary will not contain valid data. I would merge this one. 

We need to get information on which pin is the clock and at what frequency it is working. I think that this can be done when https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1291 gets solved.

For now I think it would be ok to merge this PR.

xc7/utils/vivado_create_runme.py
```diff
@@ -11,7 +11,6 @@ def create_runme(f_out, args):
 
 read_verilog {bit_v}
 synth_design -top {top}
-write_checkpoint -force design_{name}_pre_route.dcp
```
litghost: This was put here for a reason.  Something sourcing the tcl generates an error when either setting LOC/BEL and/or FIXED_ROUTE.

acomodi: Ok, I'll add another dcp creation than. Mainly because, for large designs, sourcing the tcl in Vivado takes quite a high amount of time, and it would be nice to have a dcp generated, in case P&R fails.

PR #1290, opened 2020-01-31
add sdc file to iobuf_i2c test

acomodi:

This PR is to check whether the clock to general interconnect issue can be solved also by adding SDC files to constrain the clock.


litghost:
CI is red?

acomodi:
@litghost Just checked, the clock net has a different name. As soon as https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1286 (as it contains the VtR update) I can add the correct sdc constraint.

litghost:
> @litghost Just checked, the clock net has a different name. As soon as #1286 (as it contains the VtR update) I can add the correct sdc constraint.

So something @tmichalak  and @acomodi should think about is designing a Yosys XDC plugin that takes timing constraints per synthesis, and outputs a SDC suitable for the output blif for VPR.

acomodi:
@litghost Exactly, I was thinking at a way to extract the clock nets names, with their relative clock periods and other information, so that they can be used to automatically generate SDC files.

litghost:
> @litghost Exactly, I was thinking at a way to extract the clock nets names, with their relative clock periods and other information, so that they can be used to automatically generate SDC files.

Filed https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1291

PR #1293, opened 2020-02-03
Added optional loading rr graph edges to the graph library

mkurc-ant:
This PR adds optional loading of VPR rr graph edges. By default they are not loaded. It also adds loading segment_id for rr graph nodes.

### DIFF
--- a/utils/lib/rr_graph/graph2.py
+++ b/utils/lib/rr_graph/graph2.py
@@ -227,6 +227,7 @@ class Graph(object):
             block_types,
             grid,
             nodes,
+            edges=None,
             build_pin_edges=True
     ):
         self.switches = switches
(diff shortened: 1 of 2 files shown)


litghost:
Why would this be used?

mkurc-ant:
@litghost So from the point of view of the Symbiflow it is not needed. But that code is a kind of library for manipulating the routing graph and I wanted to improve its functionality. I found it to be useful to access the rr graph data for debugging purposes.

litghost:
> @litghost So from the point of view of the Symbiflow it is not needed. But that code is a kind of library for manipulating the routing graph and I wanted to improve its functionality. I found it to be useful to access the rr graph data for debugging purposes.

Ah, this is optional loading, not optional writing. Got it.

PR #1294, opened 2020-02-03
Add ddr test

acomodi:
This PR adds the minimal ddr test corresponding to the one in https://github.com/SymbiFlow/prjxray/pull/1218.


acomodi:
@litghost @mithro As for the minilitex_ddr_test (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1220), this one, that does not include the CPU, has the same timing issues.

In fact, by adding the clock to the vivado output dcp, and printing the timing summary, there are plenty of paths that do not meet HOLD/SETUP constraints.

Here is the `timing_report_summary` of the last local run I had: https://pastebin.com/AfLvBXam

xc7/utils/prjxray_create_place_constraints.py
```diff
@@ -159,9 +159,12 @@ def __init__(self, conn, io_locs, blif_data):
 
                     clock['sink_nets'].append(sink_net)
 
-                    assert sink_net in self.input_pins or sink_net in self.clock_sources, (
-                        sink_net, self.input_pins, self.clock_sources.keys()
-                    )
+                    # TODO: Some sinks for clock networks could come from logic cells.
+                    # For instance, a negated clock does not have a clock as source, but
+                    # the output of a logic cell.
```
litghost: This is (practically) never an acceptable solution for inverting a clock.  This should remain an error.  If a design is inverting a clock through logic, that is a **bug** in the synthesis.

xc7/tests/ddr/arty.json
```diff
@@ -0,0 +1,234 @@
+{
```
litghost: Why can this not be generated automatically right now?  @mkurc-ant ?

xc7/tests/ddr/README.md
```diff
@@ -0,0 +1,16 @@
...
+- UART-DDR design, which is the minimum design possible to test the correct functioning of the DDR module;
+- Testing script: under the `scripts` directory, there is a script that performs memory calibration, and is used to test the correctness of the design.
+
+## Testing
+
+To test the design, do the following:
+
+```
+cd scripts
+make test_dram
```
litghost: I assume these instructions are from the prjxray version?  They don't apply here?

xc7/tests/ddr/arty.sdc
```diff
@@ -0,0 +1,13 @@
+create_clock -period 5 sys4x_clk__main_clkout_buf1 -waveform {0.000 2.500}
```
litghost: Some of these clocks are related?  Do they need a clock group or some such?  I'm not totally clear.  @mithro ?

xc7/tests/ddr/scripts/Makefile
```diff
@@ -0,0 +1,5 @@
+litex/litex/tools/litex_client.py:
```
litghost: Not sure we need this script?

litghost: If we do need it, should be convert to `CMakeLists.txt`?

xc7/tests/ddr/scripts/arty.py
```diff
@@ -0,0 +1,123 @@
...
+
+import argparse
+
+from migen import *
+
+from litex_boards.platforms import arty
+from litex.build.xilinx import VivadoProgrammer
+from litex.build.xilinx.vivado import vivado_build_args, vivado_build_argdict
+
+from litex.soc.cores.clock import *
```
litghost: No `*` imports allowed

xc7/tests/ddr/scripts/test_sdram.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
```
litghost: All the non-generated python files need root level file comments explaining what they are, and their purpose.

xc7/utils/prjxray_create_place_constraints.py
acomodi: The problem is that we need an inverted clock, and the only way of obtaining it are:
- negate through logic (which is done by Vivado in the minitest as well)
- phasing by 180 degrees in the PLL (which does not actually work correctly in the minitest)
- using the MMCM that has CLKB outputs as well

litghost: There is option 4:
 - Invert using dedicated inverters at the site

Where is this invert clock coming from and going to?

> * negate through logic (which is done by Vivado in the minitest as well)

Be more specific?  Is it inverted in logic the input netlist, or inverted in logic in the final placed/routed design.  If inverted in logic in the final P&R'd design, does that design actually met timing?

acomodi: > Where is this invert clock coming from and going to?

It comes from the fast clock (200MHz) clock buffer and it is routed to the CLKB pin of the ISERDESes

> Invert using dedicated inverters at the destination site

In this case this is not possible, at least now. I think that there is a `IS_CLKB_INVERTED` property, and if yes, we need to enhance the fuzzer to get support for it.

> Is it inverted in logic the input netlist, or inverted in logic in the final P&R'd design

I need to check this actually, Vivado automatically infers BUFGs for clock-related signals, in fact I could not find this clock inverted in logic in the top.edif file from the minitest.

litghost: > It comes from the fast clock (200MHz) clock buffer and it is routed to the CLKB pin of the ISERDESes

The CLKB pin to the ISERDES **has** a dedicated invert.  Using a logic-based inverter here is a synthesis **bug**.

xc7/tests/ddr/README.md
acomodi: Actually they do, but the Makefile is there to download the litex dependency needed by the test_sdram script.

I think that we could just avoid using the Makefile and call the script directly, and add the information in the README to get the litex dependency to let the test_sdram script to correctly run.

xc7/utils/prjxray_create_place_constraints.py
litghost: CLKB_inverter

acomodi: Then we need enhancement to the bufg fuzzer to be able to use that inverter. I'll also make sure it works in the minitest as well.

litghost: > In this case this is not possible, at least now. I think that there is a `IS_CLKB_INVERTED` property, and if yes, we need to enhance the fuzzer to get support for it.

I'm not confident that this feature actually exists (or it may be conflated with ISERDES in use).

No matter the conclusion, using a logic-based clock inverter instead of the dedicate hardware clock inverter is never the right answer.  The timing violations are likely a **result** of using the logic-based clock inverter.

litghost: > Then we need enhancement to the ILOGIC fuzzer to be able to use that inverter. I'll also make sure it works in the minitest as well.

Before going down this path, first make a minitest to explore what we want to do.  Given that mini-tests to date show no unknown bits, it is possible we've already encountered this bit.

xc7/tests/ddr/README.md
```diff
@@ -0,0 +1,16 @@
+# UART-DDR test
```
mithro: Does this test work with Yosys for synthesis into Vivado for place and route?

litghost: > Does this test work with Yosys for synthesis into Vivado for place and route?

It is worth noting that even if yosys emits a logic invert, Vivado may still remove it during the load of the input.

acomodi: @mithro yes, this work both with Vivado full flow and with Yosys synthesis + Vivado p&r.

For what regards the Yosys + Vivado test, the inverted clock still appears as logic-inverted in the final dcp file (after route).

litghost: > For what regards the Yosys + Vivado test, the inverted clock still appears as logic-inverted in the final dcp file (after route).

Does this design meet timing?  Comparing the Vivado full flow vs Yosys synthesis, do you see reduced setup and/or hold timing margins?

mithro: Does the Yosys->Vivado flow work on **hardware**?

acomodi: @mithro Yes it does work, I still need to get a comparison between setup/hold timings though

mithro: In the after place and route for the Yosys->Vivado flow, what is the pathway for this clock?

acomodi: - Full Vivado flow: the CLKB-related clock for ISERDESes is the same as the CLK one, but, as pointed by @litghost, the CLKB gets automatically inverted.
- Yosys->Vivado flow: the fast clock gets to a SLICE, gets inverted and enters a BUFG, from which it reaches the CLKB pin of the ISERDES.
Screenshot from 2020-02-04 12-42-37

The fact is that, by creating some smaller designs with the ISERDES only, and inverting the CLKB with the `IS_CLKB_INVERTED` parameter, nothing actually changes. I have two identical `top.bits` and `top.fasm` of two designs, which I have double checked differ only in the inversion of the CLKB

mithro: Can you add a picture for the Full Vivado flow too?

mithro: https://forums.xilinx.com/t5/Implementation/IS-CLK-INVERTED-attribute-of-ISERDES/td-p/991945

> The ISERDESE2 in XAPP524 has “INTERFACE_TYPE = NETWORKING".   Table 3-1 of UG471 says that CLKB should “Always connect to inverted CLK unless in MEMORY_QDR mode.” 
> 
> UG472 shows that outputs of the MMCM come in pairs (eg. CLKOUT1, CLKOUT1B), with CLKOUT1B being an inverted version of CLKOUT1.  So, for the ISERDESE2 in XAPP524, you could (for example) drive the CLK input from MMCM-CLKOUT1 and drive CLKB input from MMCM-CLKOUT1B.  For both CLK and CLKB, ensure that a BUFG is placed between the MMCM and the ISERDESE2 to “balance” the clock routing and reduce skew between the two clocks.  When doing this you should also drive CLKDIV of ISERDESE2 from another clock output of the MMCM (and through a BUFG) as described on page-153 of UG471(v1.10).
>

xc7/utils/prjxray_create_place_constraints.py
litghost: If we remove the change in this file, does the PR still go green?  I still believe this change is incorrect, and should be removed from this PR.

acomodi: Not yet, there still is the need to understand how to generate the inverted clock, or at least how to use the CLKB inverter in the ISERDESEs.

acomodi: @litghost this has been fixed. By using the same clock input for CLK and CLKB, the implementation works on HW.

xc7/tests/ddr/scripts/arty.py
acomodi: Fixed

xc7/tests/ddr/scripts/test_sdram.py
acomodi: Done

xc7/tests/ddr/arty.sdc
```diff
@@ -0,0 +1,12 @@
...
```
litghost: These clock net names are not found in the input net list.  @acomodi can you investigate how hard it is to fix VPR to accept the input net names?  Alternatively, can we modify the input verilog to avoid aliases on the clock nets so that the constraints can match the verilog?

PR #1308, opened 2020-02-12
Remove VPR flags set to default values, and set --quick_check_route

HackerFoo:
Clean up unnecessary flags, and enable `--quick_check_route`.


litghost:
What is the expected behavior here?

HackerFoo:
@litghost  No change, other than taking slightly longer because of `--quick_check_route`.

PR #1313, opened 2020-02-14
fasm2bels: fix LDCE is_c_inverted

acomodi:

Apparently the `IS_C_INVERTED` parameter for the LDCE does not exist. In fact, I get the following when building a fasm2bels generated verilog file:

```
ERROR: [Synth 8-3438] module 'LDCE' declared at '/opt/Xilinx/Vivado/2017.2/scripts/rt/data/unisim_comp.v:20239' does not have any parameter 'IS_C_INVERTED' used as named parameter override [symbiflow/prjxray/minitests/f2b_vivado/build/top_bit.v:89128]
```

### DIFF
--- a/xc7/fasm2bels/clb_models.py
+++ b/xc7/fasm2bels/clb_models.py
@@ -1073,7 +1073,11 @@ def process_slice(top, s):
 
             site.add_internal_source(ff5, 'Q', lut + '5Q')
             ff5.parameters['INIT'] = init
-            ff5.parameters['IS_C_INVERTED'] = IS_C_INVERTED
+
+            if name in ['LDCE', 'LDPE']:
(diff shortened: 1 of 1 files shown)


xc7/fasm2bels/clb_models.py
```diff
@@ -1073,7 +1073,9 @@ def connect_ce_sr(bel, ce, sr):
 
             site.add_internal_source(ff5, 'Q', lut + '5Q')
             ff5.parameters['INIT'] = init
-            ff5.parameters['IS_C_INVERTED'] = IS_C_INVERTED
```
litghost: I believe this should be "IS_G_INVERTED" for LDCE or LDPE.

PR #1316, opened 2020-02-17
Add Zephyr Litex test

kgugala:
This PR adds simple (no DDR) Litex/Vex SoC test. The system is capable to boot Zephyr.


litghost:
This test is missing an SDC file to constraint clocks.

Issue #1274, opened 2020-01-23
ReadTheDocs CI target randomly failing


acomodi:
In some cases, the ReadTheDocs CI build fails with the following message:

```
Traceback (most recent call last):
  File "/home/docs/.pyenv/versions/2.7.16/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/home/docs/.pyenv/versions/2.7.16/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/docs/checkouts/readthedocs.org/user_builds/symbiflow-arch-defs/envs/1267/lib/python2.7/site-packages/pip/__main__.py", line 16, in <module>
    from pip._internal import main as _main  # isort:skip # noqa
  File "/home/docs/checkouts/readthedocs.org/user_builds/symbiflow-arch-defs/envs/1267/lib/python2.7/site-packages/pip/_internal/__init__.py", line 40, in <module>
    from pip._internal.cli.autocompletion import autocomplete
  File "/home/docs/checkouts/readthedocs.org/user_builds/symbiflow-arch-defs/envs/1267/lib/python2.7/site-packages/pip/_internal/cli/autocompletion.py", line 8, in <module>
    from pip._internal.cli.main_parser import create_main_parser
  File "/home/docs/checkouts/readthedocs.org/user_builds/symbiflow-arch-defs/envs/1267/lib/python2.7/site-packages/pip/_internal/cli/main_parser.py", line 12, in <module>
    from pip._internal.commands import (
  File "/home/docs/checkouts/readthedocs.org/user_builds/symbiflow-arch-defs/envs/1267/lib/python2.7/site-packages/pip/_internal/commands/__init__.py", line 6, in <module>
    from pip._internal.commands.completion import CompletionCommand
  File "/home/docs/checkouts/readthedocs.org/user_builds/symbiflow-arch-defs/envs/1267/lib/python2.7/site-packages/pip/_internal/commands/completion.py", line 6, in <module>
    from pip._internal.cli.base_command import Command
  File "/home/docs/checkouts/readthedocs.org/user_builds/symbiflow-arch-defs/envs/1267/lib/python2.7/site-packages/pip/_internal/cli/base_command.py", line 25, in <module>
    from pip._internal.index import PackageFinder
ImportError: cannot import name PackageFinder
```

mithro:
[...]

PR #1318, opened 2020-02-18
oserdes: fix IS_T_INVERTED parameter

acomodi:

This PR fixes a bug related to the `IS_T{N}_INVERTED` parameter during techmap of OSERDESes.


acomodi:
Only thing that went red is the docs. Probably same issue as here: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1274.

Merging

PR #1321, opened 2020-02-19
Refactor SDC to avoid VPR net aliases.

litghost:
This is a workaround for the VPR bug between SDC constraints and net aliases for the DDR uart test, until VPR is fixed.


litghost:
This change somehow caused https://github.com/SymbiFlow/yosys-symbiflow-plugins/issues/8, unclear why.

PR #1324, opened 2020-02-20
Add workaround for LV wire R/C values.

litghost:
This workaround is required because of the workaround present in prjxray
around LV pip data.

Needs prjxray-db after 84b1457b885861fee221c5c6255d41cfb962ffc0


litghost:
> Can you log an issue about the FIXME?

Of course, done. https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1326

PR #1325, opened 2020-02-20
Suppress useless warnings from check_route.

litghost:
These warnings are now being generated since check route is running.

For context without this there are ~50k lines saying something to affect of:
```
Warning 17512: in check_route: found non-adjacent segments in traceback while checking net 6591:
  RR node: 863411 type: CHANY location: (96,14) track: 393 len: 0 longline: 0 seg_type: FAN_BOUNCE dir: BI_DIR capacity: 1
  RR node: 801604 type: CHANY location: (95,14) track: 61 len: 0 longline: 0 seg_type: INPINFEED dir: BI_DIR capacity: 1
```

Which is totally meaningless for our graph.


mithro:
LGTM!

PR #1332, opened 2020-02-25
Add A200T support

litghost:
@mithro @HackerFoo buttons for Nexys Video is working (with VTR @ http://github.com/litghost/vtr-verilog-to-routing/commits/85ea49173b99f3b8dde3c6d06573691e063ef5b9)

This does **not** enable the A200T kokoro runner, I plan to make a follow PR for that.


xc7/yosys/synth.tcl
```diff
@@ -7,16 +7,16 @@ yosys -import
 
 # Map (actually rename) explicitly instantiated IBUFs/OBUFs in order to
 # distinguish them from the inferred ones.
-techmap -map $::env(symbiflow-arch-defs_SOURCE_DIR)/xc7/techmap/iob_map.v
+#techmap -map $::env(symbiflow-arch-defs_SOURCE_DIR)/xc7/techmap/iob_map.v
```
HackerFoo: Why are these commented out? Should there be an issue to add them back?

litghost: That is actually a very subtle question.  The bottom line is: ROI based designs need these lines, and ROI-less design **cannot** have these lines.  

The reason things worked before on the basys3 is that all of the IO attributes could be left as defaults, and on the nexys video, that is not the case.  More tcl logic is required make this synth script work for both ROI and ROI-less designs, and I want to get a clocking design working before I go back and fix it.

I also think I want to split off the changes to the `synth.tcl` into a smaller PR anyways.  I'm working on splitting off the other changes now, like the `PULLTYPE` fix, so that reviews can focus on the changes.

litghost:
I've also tested counter, and it's working too.   So I think we have A200T up and running.  Now to split up and merge smaller changes, and reduce runtime to something more manageable.

mithro:
How is the runtime looking?

litghost:
> How is the runtime looking?

I'm working on collecting that data now.  There are roughly 4 large time sinks in the routing import:
1. Form channels
2. Create edges
3. Routing import (rr graph patching)
4. Lookahead computation

I've already cut step 4 in half, and I'm working on step 3 right now because it tends to be the one that gets re-invoked the most whenever VPR is recompiled or the arch is regenerated.  There is a caching step that kicks in when the output of step 3 is identical to skip step 4.

xc7/yosys/synth.tcl
litghost: Fix is going in for iob inference here: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1339

litghost:
1. Form channels

litghost:
A simple conversion of the routing import from xml based to capnp based drops the runtime by 15 minutes, at the cost of 10 GiB of memory.  This is mostly a side affect of capnp not having an incremental writing ability.

Routing import with capnp:
```
        User time (seconds): 1587.44                                                                     
        System time (seconds): 186.20                                                                    
        Percent of CPU this job got: 97%                                                                 
        Elapsed (wall clock) time (h:mm:ss or m:ss): 30:22.20
        Maximum resident set size (kbytes): 25105640
```

PR #1339, opened 2020-02-25
Fix iobuf inference

mkurc-ant:
This PR fixes IOBUF inference during synthesis for ROI and ROI-less targets. The USE_ROI is passed to the TCL synthesis script where it is used to determine whether to infer IOBs or not.


xc7/yosys/synth.tcl
```diff
@@ -1,22 +1,18 @@
+# Import the commands from the plugins to the tcl interpreter
 yosys -import
 
 plugin -i xdc
 plugin -i fasm
-#Import the commands from the plugins to the tcl interpreter
-yosys -import
```
litghost: Isn't this required?

mkurc-ant: But it's done in the first line of the script.

mkurc-ant:
I haven't checked my changes thoroughly. I'm waiting what the CI says.

xc7/yosys/synth.tcl
litghost: If you read the comment, the `yosys -import` imports new commands that come from the XDC and FASM plugins.

mkurc-ant: Ahh..., you're right.

litghost:
CI is red.  Worth noting that io buf inference isn't really ready for general use, see: https://github.com/YosysHQ/yosys/issues/1720

mkurc-ant:
I've read that. But unfortunately we have to infer IO bufs as many designs like eg. the Litex genrated ones do not instance them. And if there are no IO bufs then the XDC plugin has nothing to attach IOSTANDARD settings to.

litghost:
> I've read that. But unfortunately we have to infer IO bufs as many designs like eg. the Litex genrated ones do not instance them. And if there are no IO bufs then the XDC plugin has nothing to attach IOSTANDARD settings to.

I agree we need to have working IO buf inference, I'm just worried that the broken `iopadmap` implementation will cause many failures.  We may need to patch yosys before we enable inference in symbiflow-arch-defs.

litghost:
Superceded by #1349

PR #1344, opened 2020-02-26
Add clock constraints for Vivado timing analysis to DDR UART Test

litghost:


xc7/tests/ddr/arty_clocks.xdc
```diff
@@ -0,0 +1,7 @@
+create_clock -name clk100 -period 10.0 [get_nets clk100]
+
+set_false_path -quiet -to [get_nets -quiet -filter {mr_ff == TRUE}]
+
+set_false_path -quiet -to [get_pins -quiet -filter {REF_PIN_NAME == PRE} -of [get_cells -quiet -filter {ars_ff1 == TRUE || ars_ff2 == TRUE}]} ]
+
+set_max_delay 2 -quiet -from [get_pins -quiet -filter {REF_PIN_NAME == Q} -of [get_cells -quiet -filter {ars_ff1 == TRUE}]} ] -to [get_pins -quiet -filter {REF_PIN_NAME == D} -of [get_cells -quiet -filter {ars_ff2 == TRUE}]} ]
```
acomodi: I wonder whether we need these. I need to double check, but probably the `ars_ff1` and `mr_ff` properties are not set after fasm2bels.

I think the only relevant information in the xdc is the clock creation to get the timing summary data.
Could be worth to move this xdc to the `common` directory and name it something like `arty_ddr_clocks.xdc`, as all the tests regarding ddr and the SoCs in general use the same clock name (baselitex, mini_ddr, mini).

litghost: > Could be worth to move this xdc to the `common` directory and name it something like `arty_ddr_clocks.xdc`, as all the tests regarding ddr and the SoCs in general use the same clock name (baselitex, mini_ddr, mini).

I don't think this is a good idea, because small changes to the design would make the XDC different.

PR #1345, opened 2020-02-26
Add clean to update_tools.py to cleanup after some resets.

litghost:


mithro:
Could you add a comment about why you are cleaning `-fx`?

Issue #1346, opened 2020-02-26
Vivado errors on some valid clock placements


litghost:
In particular, Vivado errors on explicit BUFH -> PS7 placement.  I've confirmed that Vivado will take `axi_regs_zybo` design, and route the design the same way as current master symbiflow-arch-defs, but generate an error because of the BUFH -> PS7 placement.  The solution is to not emit explicit BUFH instances, but instead emit route through `FIXED_ROUTE`'s.  This is tricky, and is not the most important thing to work on right now.

For now, I recommend we disable the DIFF_FASM test for `axi_regs_zybo` until this is fixed.

litghost:
@mkurc-ant FYI

PR #1347, opened 2020-02-26
Disable axi_lite_reg DIFF_FASM test until workaround is implemented.

litghost:


litghost:
Issues filed to workaround #1346

PR #1348, opened 2020-02-26
Working initial capnp based rr graph patcher.

litghost:
- Converts fasm2bels to use capnp for faster rr graph reading.
 - Converts xc7 routing import to use capnp for faster rr graph reading
   and writing.

### DIFF
--- a/utils/lib/rr_graph_capnp/graph2.py
+++ b/utils/lib/rr_graph_capnp/graph2.py
@@ -2,8 +2,10 @@ import os.path
 import re
 from lib.rr_graph import graph2
 from lib.rr_graph import tracks
+import gc
 
 import capnp
+import capnp.lib.capnp
 capnp.remove_import_hook()
 
 CAMEL_CASE_CAPITALS = re.compile('([A-Z]+)')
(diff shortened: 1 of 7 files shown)


utils/lib/rr_graph_capnp/graph2.py
```diff
@@ -0,0 +1,548 @@
...
+        CAPNP_ENUM_CACHE[key] = enum_type.__dict__[camel_case_e]
+
+    return CAPNP_ENUM_CACHE[key]
+
+
+def read_switch(sw):
+    timing = sw.timing
+    sizing = sw.sizing
+
+    return graph2.Switch(
```
mithro: Why are the `graph2.Switch` object and the captnproto object are not compatible?

Could we make the attribute names for these objects in `graph2` just match the captnproto versions?

litghost: The whole point of graph2 objects is to be storage independent.  As for field names, the capnp naming is camelCase, which non of the rest of our python code is.

litghost:
> Only one real comment -- there does seem to be a lot of conversion from `camelCaseStyle` to `camel_case_style`? Could we make things easier by just changing one side? (We can add an exception to yapf if needed too?)

The camelCase is **required** by capnp, and we just keep it isolated from the rest of our code.

utils/lib/rr_graph_capnp/graph2.py
```diff
@@ -0,0 +1,548 @@
...
+            channels_obj,
+            connection_box_obj,
+            num_nodes,
+            nodes_obj,
+            num_edges,
+            edges_obj,
+            node_remap=lambda x: x
+    ):
+        """
+        Writes the routing graph to the XML file.
```
acomodi: This is not an XML file anymore right? I mean, it gets binarized by capnp, so maybe this comment could be misleading

litghost: Fixed

Issue #1275, opened 2020-01-23
[xc7] 128-bit and 256-bit LUT-RAM vendor tests are disabled


litghost:
The 128-bit and 256-bit LUT-RAM vendor tests are disabled (see #1268).  These are likely caused by bugs in how fasm2bels emits these structures.  Likely path forward is to have fasm2bels emit RAM64M + F7/F8 muxes rather than 2x RAM128X1S/RAM128X1D/RAM256X1S structures when all 4 LUT-RAM's are present in a SLICEM.

This issue is a split from #1269

PR #1320, opened 2020-02-19
Add Base Litex minitest capable of running Linux on Vexriscv

tmichalak:
This PR adds a test with the Linux on Vexriscv Litex design which features a Linux capable VexRiscv CPU.
The design start the BIOS and passes memtest but fails to download the binaries using TFTP or UART.

PR #1356, opened 2020-03-02
Fix DRAM vendor tools tests

acomodi:
This PR is to fix issues with DRAM128 (both single and dual port) and partially fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1275.

The fixes applied are of two kinds:

- **Architecture definition**: the DRAM128 mode for SLICEM did not provide features and mux connections for the CI, BI and AI input pin muxes. 
A question I couldn't actually answer is why the implemented designs did work on HW even though the top.fasm did not emit the `CLBLM_R_X3Y0.SLICEM_X0.CLUT.DI1MUX.DI_DMC31` for instance. The fact is that it actually appears in the `top.bit.fasm`.

- **fasm2bels**: there are some sinks which should have be routed through, but cannot appear in the resulting verilog file (more explanation [here](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1320#issuecomment-592605250)). The second patch adds the sinks, but then removes them based on the selected DRAM128 type.


acomodi:
The same fixes applied for DRAM256x1S resulted to be effective. This PR now completely solves https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1275

litghost:
> * A question I couldn't actually answer is why the implemented designs did work on HW even though the top.fasm did not emit the `CLBLM_R_X3Y0.SLICEM_X0.CLUT.DI1MUX.DI_DMC31` for instance. The fact is that it actually appears in the `top.bit.fasm`.

So the hardware works in both mux configurations, as long as the router routes to the correct LUT-RAM data in port.  Vivado is picky about which ports it uses, but the underlying hardware is more flexible than what Vivado accepts.

PR #1361, opened 2020-03-04
Integrate VPR with SDC alias fix.

litghost:
- SDC files that were previously using generated aliases now uses one of
  the net names.


litghost:
@acomodi Thanks, new VPR is working great!

acomodi:
CI went green on all targets. Merging.

PR #1363, opened 2020-03-07
Use latest VtR from conda and update flags.

litghost:
As part of upstream PRs being accepted some flags were renamed for
clarity.


mithro:
LGTM. Merge if green.

PR #1367, opened 2020-03-10
XC7 re-work

acomodi:
This PR is to make xc7 subtree more generic and allow the presence of other device families.

The idea is to have an "umbrella" directory, namely `xilinx`, to hold the various device families (e.g. xc7, spartan, etc).

Given that most of the code in the xc7 tree (utils, cmake files, etc.) could potentially be shared among the various families, it should reside in a common directory in the `xilinx` tree.

This has multiple benefits:
- generalize scripts and utilities;
- avoid copy pasting existing code for future families support;
- start a clean-up process of possible unused code.


xilinx/common/cmake/arch_define.cmake
```diff
@@ -14,23 +41,29 @@ function(ADD_XC7_ARCH_DEFINE)
...
     ARCH ${ARCH}
+    FAMILY ${FAMILY}
+    DOC_PRJ ${PRJRAY_DIR}
+    DOC_PRJ_DB ${PRJRAY_DB_DIR}
     PROTOTYPE_PART ${PROTOTYPE_PART}
     YOSYS_SYNTH_SCRIPT ${YOSYS_SYNTH_SCRIPT}
     YOSYS_CONV_SCRIPT ${YOSYS_CONV_SCRIPT}
     DEVICE_FULL_TEMPLATE \${DEVICE}-\${PACKAGE}
-    CELLS_SIM ${YOSYS_DATADIR}/xilinx/cells_sim.v ${symbiflow-arch-defs_SOURCE_DIR}/xc7/techmap/cells_sim.v
+    CELLS_SIM ${YOSYS_DATADIR}/xilinx/cells_sim.v ${symbiflow-arch-defs_SOURCE_DIR}/xilinx/common/techmap/cells_sim.v
```
mkurc-ant: That will have to be specialized per `PRJRAY_ARCH`

common/cmake/devices.cmake
```diff
@@ -116,6 +120,9 @@ function(DEFINE_ARCH)
   set(
     oneValueArgs
     ARCH
+    FAMILY
+    DOC_PRJ
```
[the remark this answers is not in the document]
acomodi: Sure, done.

xilinx/common/cmake/arch_define.cmake
acomodi: Right, done, moved all techmap stuff in `xc7` directory.

acomodi:
We need to wait for https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1252 to be merged before proceeding with this PR, as the two PRs will conflict and `make install` has higher priority than this one.

tests/9-scalable_proc/CMakeLists.txt
```diff
@@ -97,6 +98,7 @@ scalable_proc(
   TOP_FILE basys3_top.v
   BOARD basys3
   INPUT_IO_FILE basys3.pcf
+  INPUT_SDC_FILE ${symbiflow-arch-defs_SOURCE_DIR}/xilinx/xc7/tests/common/basys3.sdc
```
kgugala: why do we need absolute path here? Other files use relative paths

acomodi: This because the SDC file is not in the scalable proc tests, but I think it would actually be better to copy the basys3.sdc from the xc7 path. Because these set of tests should be detached from the particular architecture.

acomodi: Done

PR #1369, opened 2020-03-11
DNM: test if penalty cost can be avoided

acomodi:


litghost:
You should re-run this PR with different VPR seeds (`--seed`) to see if this is stable along multiple runs.

acomodi:
@litghost Ok, will do. One thing we cannot check though with this PR is the fact that we may not confirm that the weird BUFH -> BUFH connection is avoided.

litghost:
> @litghost Ok, will do. One thing we cannot check though with this PR is the fact that we may not confirm that the weird BUFH -> BUFH connection is avoided.

I think BUFH -> BUFH causes a vendor test failure.

Issue #1370, opened 2020-03-12
make env failing


jchidley:
Today, I followed the [Getting Started Instructions](https://symbiflow.readthedocs.io/en/latest/symbiflow-arch-defs/docs/source/getting-started.html)

the step `make env` failed.

Here is the output of "C:/Users/jackc/git/symbiflow-arch-defs/build/CMakeFiles/CMakeError.log"

```cmd
Determining if the include file pthread.h exists failed with the following output:
Change Dir: C:/Users/jackc/git/symbiflow-arch-defs/build/CMakeFiles/CMakeTmp

Run Build Command(s):C:/Program Files (x86)/Microsoft Visual Studio/2019/BuildTools/MSBuild/Current/Bin/MSBuild.exe cmTC_7c8f0.vcxproj /p:Configuration=Debug /p:Platform=x64 /p:VisualStudioVersion=16.0 /v:m
Microsoft (R) Build Engine version 16.4.0+e901037fe for .NET Framework

Copyright (C) Microsoft Corporation. All rights reserved.



  Microsoft (R) C/C++ Optimizing Compiler Version 19.24.28319 for x64

  Copyright (C) Microsoft Corporation.  All rights reserved.

  CheckIncludeFile.c

  cl /c /Zi /W3 /WX- /diagnostics:column /Od /Ob0 /D WIN32 /D _WINDOWS /D "CMAKE_INTDIR=\"Debug\"" /D _MBCS /Gm- /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /Fo"cmTC_7c8f0.dir\Debug\\" /Fd"cmTC_7c8f0.dir\Debug\vc142.pdb" /Gd /TC /errorReport:queue "C:\Users\jackc\git\symbiflow-arch-defs\build\CMakeFiles\CMakeTmp\CheckIncludeFile.c"

C:\Users\jackc\git\symbiflow-arch-defs\build\CMakeFiles\CMakeTmp\CheckIncludeFile.c(1,10): fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory [C:\Users\jackc\git\symbiflow-arch-defs\build\CMakeFiles\CMakeTmp\cmTC_7c8f0.vcxproj]



Performing C++ SOURCE FILE Test CXX_COMPILER_SUPPORTS_UNDEF_VAR failed with the following output:
[...]

PR #1379, opened 2020-03-23
cmake: update minimum version to 3.14.0

acomodi:

This PR updates the minimum CMake version required.  This should solve https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1370#issuecomment-602045210

### DIFF
--- a/.github/travis/script.sh
+++ b/.github/travis/script.sh
@@ -5,8 +5,6 @@ set -e
 
 $SPACER
 
-which cmake
-
 start_section "symbiflow.configure_cmake" "Configuring CMake (make env)"
 make env
 cd build
--- a/.github/travis/script.sh
+++ b/.github/travis/script.sh
@@ -5,6 +5,8 @@ set -e
 
 $SPACER
 
+which cmake
+
 start_section "symbiflow.configure_cmake" "Configuring CMake (make env)"
 make env
 cd build
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,9 @@ before_install:
 
 install:
  - stdbuf -i0 -o0 -e0 ./.github/travis/install.sh
-
+ - wget https://cmake.org/files/v3.14/cmake-3.14.0-Linux-x86_64.tar.gz
+ - tar -xf cmake-3.14.0-Linux-x86_64.tar.gz
+ - export PATH=`realpath cmake-3.14.0-Linux-x86_64/bin`:$PATH
 script:
   # ping stdout every 9 minutes or Travis kills build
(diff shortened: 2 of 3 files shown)


.github/travis/install.sh
```diff
@@ -3,5 +3,10 @@
 source .github/travis/common.sh
 set -e
 
+# Fix cmake version
```
mithro: You should use https://apt.kitware.com/

mithro: https://docs.travis-ci.com/user/languages/cpp/#cmake

mithro: https://github.com/sourceryinstitute/OpenCoarrays/blob/964360cf183baf07b846ea13818509d14711eba3/.travis.yml#L71-L81

acomodi: Ok. I am unsure whether I did that correctly. Travis CI still uses 3.12.4.

Would it be ok to install manually cmake in the `install` step following https://apt.kitware.com/ instructions?

mithro: It looks like it installed cmake 3.17 -- you should add a `which cmake` to see what is going on here....
```
Get:3 https://apt.kitware.com/ubuntu xenial/main amd64 cmake-data all 3.17.0-0kitware1 [1,683 kB]
Get:4 https://apt.kitware.com/ubuntu xenial/main amd64 cmake amd64 3.17.0-0kitware1 [7,422 kB]
```

acomodi: Turns out that the default cmake that is being used is the one shipped with xenial from Travis is 3.12.4, regardless of the kitware installation.

```
which cmake
/usr/local/cmake-3.12.4/bin/cmake
```

To solve this I have manually extracted the binaries from the cmake official releases

mithro: How does `/usr/local/cmake-3.12.4/bin/cmake` end up in the path?

acomodi: Unsure, I believe it gets there as a default when using xenial. Another way probably to solve this issue is to use the apt.kitware package, as previously done, and then update the PATH to use the kitware cmake in the install step.

mithro: @acomodi Let's use the apt.kitware package and figure out how to make sure the `/usr/local/cmake-3.12.4/bin/cmake` is removed / not used. Worse comes to worse we can do a `sudo rm -rf /usr/local/cmake-3.12.4`

acomodi: @mithro Done, what was missing was making the priority of the kitware cmake higher in the PATH env var. Travis CI should now turn green.

mithro:
Merge if green.

acomodi:
CI went gree. Merging

PR #1381, opened 2020-03-24
xc7: skip prohibited sites during physical tiles import

acomodi:

This PR adds the possibility to skip the physical tile import in the DB whenever the tiles contain prohibited sites.

This requires first https://github.com/SymbiFlow/prjxray/pull/1278

### DIFF
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -81,6 +81,7 @@ CREATE TABLE phy_tile(
   grid_x INT,
   grid_y INT,
   clock_region_pkey INT,
+  prohibited INT,
   FOREIGN KEY(tile_type_pkey) REFERENCES tile_type(pkey),
   FOREIGN KEY(clock_region_pkey) REFERENCES clock_region(pkey)
 );
--- a/utils/lib/connection_database.sql
+++ b/utils/lib/connection_database.sql
@@ -81,7 +81,6 @@ CREATE TABLE phy_tile(
   grid_x INT,
   grid_y INT,
   clock_region_pkey INT,
-  prohibited INT,
   FOREIGN KEY(tile_type_pkey) REFERENCES tile_type(pkey),
   FOREIGN KEY(clock_region_pkey) REFERENCES clock_region(pkey)
 );
@@ -118,6 +117,7 @@ CREATE TABLE site_instance(
     y_coord INT,
     site_pkey INT,
     phy_tile_pkey INT,
+    prohibited BOOLEAN,
     FOREIGN KEY(site_pkey) REFERENCES site(pkey),
     FOREIGN KEY(phy_tile_pkey) REFERENCES phy_tile(pkey)
 );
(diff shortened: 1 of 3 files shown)


litghost:
prjxray change has been merged, add submodule update?

utils/lib/connection_database.sql
```diff
@@ -81,6 +81,7 @@ CREATE TABLE phy_tile(
   grid_x INT,
   grid_y INT,
   clock_region_pkey INT,
+  prohibited INT,
```
litghost: Sites are prohibited, not tiles.  This is the wrong place for the flag.  Also this should be a BOOLEAN, not an INT.

acomodi: The only reason to set the prohibited to Tiles is that it was more straightforward to skip the tile itself during arch_import, but I guess it is more consistent as you suggested. Will fix that and bump prjxray

xc/common/utils/prjxray_arch_import.py
```diff
@@ -480,6 +480,18 @@ def get_tiles(
...
+        if phy_tile_pkey is not None:
+            c2.execute(
+                "SELECT prohibited FROM site_instance WHERE phy_tile_pkey = ?",
+                (phy_tile_pkey, )
+            )
+
+            prohibited_sites = c2.fetchall()
+
+            # Skip generation of prohibited tiles
+            if True in prohibited_sites:
```
litghost: This won't work as expected.  fetchall returns [(True), (True)] not [True, True].

litghost: What I've done is unpack the tuple at fetchall:
```
any_prohibited_sites = any(prohibited for (prohibited,) in c2.fetchall())
```

acomodi: Yep, much cleaner, thanks for the suggestion.

utils/lib/connection_database.sql
acomodi: Fixed

xc/common/utils/prjxray_arch_import.py
acomodi: Fixed

xc/common/utils/prjxray_form_channels.py
```diff
@@ -578,6 +582,15 @@ def import_phy_grid(db, grid, conn, get_switch, get_switch_timing):
                 )
             )
 
+            site_instance_pkey = write_cur.lastrowid
+
+            cur.execute(
```
litghost: This should be "write_cur" not "cur", and then line 490 is not required.

acomodi: Sure, forgot about that. Done

acomodi:
@litghost Seems that thye Doc and Testing Arch CIs are still running, but by opening them they show up to be complete (at least to me). Would be ok to merge?

PR #1386, opened 2020-03-30
Heterogeneous tiles implementation

acomodi:

This PR is an initial implementation of the heterogeneous tiles introduced with https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1208.

TODO:
- [ ] find a better way to use the `start_tile` implementation so that it gets used more generally by the various tile import scripts.
- [x] produce a new VtR master+wip with the heterogeneous tiles implementation

### DIFF
--- a/utils/lib/pb_type_xml.py
+++ b/utils/lib/pb_type_xml.py
@@ -97,7 +97,12 @@ def add_switchblock_locations(xml):
     })
 
 
-def start_sub_tile(sub_tile_name, pin_assignments, input_wires, output_wires):
+def start_sub_tile(sub_tile_name, input_wires, output_wires):
+    """This function returns the sub tile definition for a given tile.
+
+    The returned sub tile contains only input, output and clock ports.
+    The rest of the tags are going to be added by the caller.
+    """
     sub_tile_xml = ET.Element(
         'sub_tile', {
             'name': add_vpr_tile_prefix(sub_tile_name),
@@ -147,7 +152,39 @@ def start_tile(
         output_wires=None,
         sites=None,
 ):
-    """ Starts a pb_type by adding input, clock and output tags. """
+    """ Returns a new tile xml definition.
+
+    Input parameters are:
+        - tile_name: name to assign to the tile
+        - pin_assignments: location of the pins to correctly build pin locations
+        - input and output wires: needed to correctly assign input, output and clock ports
+        - sites: set of pb_types that are going to be related to this tile
+
+    Input/Output wires and sites are mutually exclusive.
+    Both of the parameters cannot be assigned
+
+    When sites is not None, it will contain a set of different kinds of sub tiles
+    to be added to the tile definition, each corresponding to one of the site types
+    present in the parameter.
+
+    As an example:
+        - HCLK_IOI has in total 9 sites:
+            - 4 BUFR
+            - 4 BUFIO
+            - 1 IDELAYCTRL
+
+        The resulting sites parameter is a dictionary that looks like this:
+            {
+                "BUFR": [bufr_instance_1, bufr_instance_2, ...],
+                "BUFIO": [bufio_instance_1, bufio_instance_2, ...],
+                "IDELAYCTRL": [idleayctrl_instance_1]
+            }
+
+    When Input/Output wires are set, there will be only one corresponding sub tile that
+    takes the same name as the parent tile.
+    """
+
+    # Check whether sites and Input/Output wires are mutually exclusive
     assert bool(input_wires) == bool(output_wires)
     assert bool(sites) != bool(input_wires)
 
@@ -161,8 +198,7 @@ def start_tile(
 
     if sites is not None:
         for site_type in sites.keys():
-            num_sub_tile = 0
-            for site in sites[site_type]:
+            for num_sub_tile, site in enumerate(sites[site_type]):
                 sub_tile_name = "{}_{}_{}".format(
                     tile_name, site_type, num_sub_tile
                 )
@@ -170,7 +206,7 @@ def start_tile(
                 site, input_wires, output_wires = site
 
                 sub_tile_xml = start_sub_tile(
-                    sub_tile_name, pin_assignments, input_wires, output_wires
+                    sub_tile_name, input_wires, output_wires
                 )
 
                 fc_xml = add_fc(sub_tile_xml)
@@ -210,8 +246,6 @@ def start_tile(
                 sub_tile_xml.append(equivalent_sites_xml)
 
                 tile_xml.append(sub_tile_xml)
-
-                num_sub_tile = num_sub_tile + 1
     else:
         sub_tile_xml = start_sub_tile(
             tile_name, pin_assignments, input_wires, output_wires
--- a/utils/lib/pb_type_xml.py
+++ b/utils/lib/pb_type_xml.py
@@ -28,7 +28,9 @@ def object_ref(pb_name, pin_name, pb_idx=None, pin_idx=None):
     return '{}{}.{}{}'.format(pb_name, pb_addr, pin_name, pin_addr)
 
 
-def add_pinlocations(tile_name, xml, fc_xml, pin_assignments, wires):
+def add_pinlocations(
+        tile_name, xml, fc_xml, pin_assignments, wires, sub_tile_name=None
+):
     """ Adds the pin locations.
 
     It requires the ports of the physical tile which are retrieved
@@ -46,7 +48,12 @@ def add_pinlocations(tile_name, xml, fc_xml, pin_assignments, wires):
             if side not in sides:
                 sides[side] = []
 
-            sides[side].append(object_ref(add_vpr_tile_prefix(tile_name), pin))
+            if sub_tile_name is not None:
+                name = sub_tile_name
+            else:
+                name = tile_name
+
+            sides[side].append(object_ref(add_vpr_tile_prefix(name), pin))
 
     for side, pins in sides.items():
         ET.SubElement(pinlocations_xml, 'loc', {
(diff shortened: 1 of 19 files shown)


xc/common/utils/prjxray_import_tile_capacity.py
```diff
@@ -30,70 +47,28 @@ def main():
...
-        'site', {
-            'pb_type': add_vpr_tile_prefix(site.type),
-            'pin_mapping': 'custom'
-        }
-    )
-    for site_idx, site in enumerate(sites):
-        if site.type != args.pb_type:
-            continue
-
-        for site_pin in site.site_pins:
```
litghost: This mapping is still needed, where did it go?

acomodi: It is a leftover actually, I tried to move that in the utiliy lib, to extract IO wires. It is in the `get_wires` function.

litghost: Get wires is the loop here https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1386/files#diff-3c687ed1bdcb39751e14fcd2591aaaadL43-L53, I'm talking about the loop here: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1386/files#diff-3c687ed1bdcb39751e14fcd2591aaaadL81-L93

acomodi: Ah right, I misunderstood. You mean the physical to logical pin mappings. It ended up in the `start_tile` function in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1386/files#diff-858a400e24b804d76914f3604da75835. I still need to get to better generalize that utility library.

acomodi:
Update: I am testing this PR alongside with the VtR heterogenous implementation integrated in master+wip and the litex-linux base test seems to be working, meaning that it passes Memtest.

I need to setup locally a tftp server to load linux binaries in DDR

litghost:
> Update: I am testing this PR alongside with the VtR heterogenous implementation integrated in master+wip and the litex-linux base test seems to be working, meaning that it passes Memtest.
> 
> I need to setup locally a tftp server to load linux binaries in DDR

Awesome!

acomodi:
Edit: Linux boots correctly. I used UART to load binaries as I have some troubles with the tftp server (on the local host, so unrelated to the bitstream).

utils/lib/pb_type_xml.py
```diff
@@ -90,19 +97,148 @@ def add_switchblock_locations(xml):
     })
 
 
-def start_pb_type(
+def start_sub_tile(sub_tile_name, pin_assignments, input_wires, output_wires):
```
litghost, line 100: Can we have a docstring please?

utils/lib/pb_type_xml.py
```diff
@@ -90,19 +97,148 @@ def add_switchblock_locations(xml):
...
+
+
+def start_tile(
         tile_name,
         pin_assignments,
+        input_wires=None,
+        output_wires=None,
+        sites=None,
+):
+    """ Starts a pb_type by adding input, clock and output tags. """
```
litghost, line 150: Please add docstring commands explains arguments

utils/lib/pb_type_xml.py
```diff
@@ -90,19 +97,148 @@ def add_switchblock_locations(xml):
...
+    assert bool(input_wires) == bool(output_wires)
```
litghost, line 151: Can you add a comment on what these asserts are checking?

utils/lib/pb_type_xml.py
```diff
@@ -90,19 +97,148 @@ def add_switchblock_locations(xml):
...
+        {
+            'name': add_vpr_tile_prefix(tile_name),
+        },
+        nsmap={'xi': XI_URL},
+    )
+
+    if sites is not None:
+        for site_type in sites.keys():
+            num_sub_tile = 0
+            for site in sites[site_type]:
```
litghost, line 165: Should be `for num_sub_tile, site in enumerate(sites[site_type]):`, yes?

utils/lib/pb_type_xml.py
litghost: This is probably clearer written as an if/else

xc/common/cmake/project_ray.cmake
```diff
@@ -607,7 +607,6 @@ function(PROJECT_RAY_TILE_CAPACITY)
   #   ARCH <arch>
   #   TILE <tile>
   #   SITE_TYPE <site type>
-  #   SITE_COORDS [X|Y|XY]
```
litghost, line 610: Comment needs to be updated to match behavior?

utils/lib/pb_type_xml.py
acomodi: Yep, much cleaner, thanks

xc/common/cmake/project_ray.cmake
acomodi: Done

utils/lib/pb_type_xml.py
acomodi: Done

utils/lib/pb_type_xml.py
acomodi: Done

utils/lib/pb_type_xml.py
acomodi: Done

utils/lib/pb_type_xml.py
acomodi: Done

acomodi:
For some reason the vendor tool test failed. I think though, by looking at the log, that all the tests acutally passed.

litghost:
> For some reason the vendor tool buildfailed. I think though, by looking at the log, that all the tests acutally passed.

There was a failure, but I don't think it is related.

acomodi:
@litghost Would this be ok to merge given that all vendor tool tests actually passed?

utils/lib/pb_type_xml.py
```diff
@@ -28,7 +28,9 @@ def object_ref(pb_name, pin_name, pb_idx=None, pin_idx=None):
     return '{}{}.{}{}'.format(pb_name, pb_addr, pin_name, pin_addr)
 
 
-def add_pinlocations(tile_name, xml, fc_xml, pin_assignments, wires):
+def add_pinlocations(
+        tile_name, xml, fc_xml, pin_assignments, wires, sub_tile_name=None
+):
     """ Adds the pin locations.
```
litghost, line 34: Miss documentation for new argument?

utils/lib/pb_type_xml.py
```diff
@@ -90,19 +97,180 @@ def add_switchblock_locations(xml):
     })
 
 
-def start_pb_type(
+def start_sub_tile(sub_tile_name, input_wires, output_wires):
```
litghost, line 100: Please provide documentation of expect argument types in some way.  Docstring and/or type annotations are acceptable.

utils/lib/pb_type_xml.py
```diff
@@ -90,19 +97,180 @@ def add_switchblock_locations(xml):
...
+    Input parameters are:
+        - tile_name: name to assign to the tile
+        - pin_assignments: location of the pins to correctly build pin locations
+        - input and output wires: needed to correctly assign input, output and clock ports
+        - sites: set of pb_types that are going to be related to this tile
+
+    Input/Output wires and sites are mutually exclusive.
+    Both of the parameters cannot be assigned
+
+    When sites is not None, it will contain a set of different kinds of sub tiles
```
litghost, line 166: Does it make sense to just make two functions?

xc/common/cmake/project_ray.cmake
```diff
@@ -606,18 +606,16 @@ function(PROJECT_RAY_TILE_CAPACITY)
...
-  #
-  # Future support: Emit tiles with capacity that contain more than 1 site type
-  # once VPR supports it.
+  # SITE_TYPES: contains the comma separated list of sites that are going to be used
+  #             as sub tiles for the specified tile. The total number of instances
+  #             of a site type appears as the capacity of the sub tile
+
   set(options)
-  set(oneValueArgs ARCH TILE SITE_TYPE SITE_COORDS)
+  set(oneValueArgs ARCH TILE SITE_TYPES)
```
litghost, line 618: Should SITE_TYPES be a multiValueArgs?

acomodi: Done

utils/lib/pb_type_xml.py
acomodi: Yeah, probably having them split is better. I have added a new `start_heterogeneous_tile` func.

utils/lib/pb_type_xml.py
acomodi: Done

utils/lib/pb_type_xml.py
acomodi: Done

acomodi:
I have found the vendor tools error:

```
[828/1771] cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test && diff -u /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test/bram_test_36/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test/bram_test_36/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_bram_test_36_vivado.bit.fasm
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test && diff -u /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test/bram_test_36/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test/bram_test_36/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_bram_test_36_vivado.bit.fasm
--- /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test/bram_test_36/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit.fasm	2020-04-05 12:30:47.957112578 -0700
+++ /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/bram_test/bram_test_36/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_bram_test_36_vivado.bit.fasm	2020-04-05 12:37:09.500094783 -0700
@@ -4088,7 +4088,6 @@
 INT_L_X6Y110.EE4BEG2.LOGIC_OUTS_L10
 INT_L_X6Y110.EL1BEG0.NR1END1
 INT_L_X6Y110.ER1BEG2.LOGIC_OUTS_L5
-INT_L_X6Y110.FAN_ALT5.FAN_BOUNCE1
 INT_L_X6Y110.GFAN0.GND_WIRE
 INT_L_X6Y110.GFAN1.GND_WIRE
 INT_L_X6Y110.IMUX_L25.GFAN0
```

This is an ROI-related test, I will investigate what is happening here.
This did not happen for the [test PR](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1391), therefore I suppose a change introduced with this PR is generating this issue.

acomodi:
I think I do understand what is the reason, less sure about the reasons though.
By inspecting two route files (one with master and one with this PR for `bram_test_36`) and the respective fasm files, there is a difference in how the `BRAM_REGCLKB` is being routed to `$true`.

In the failing case, the `$true` VCC wire is not directly connected to the `FAN_ALT5` (which then is routed to the `BRAM_REGCLKB` pin), but there is a jump through the `FAN_BOUNCE1`.

I am unsure why the router chooses this path, instead of connecting `FAN_ALT5` directly to the VCC wire, and, more importantly, it's hard to say whether this is an issue of the heterogeneous tiles feature or other changes happened in VtR master.

acomodi:
I have run the `bram_test_36` on test on symbiflow master against a local master+wip version with the updated master and `explicit_ports` and it seems to correctly route the constant net.

This indicates that there might be something not correctly working with:
- The heterogeneous tiles VtR implementation: I do not think this is the case, because there is nothing in the heterogeneous tiles that changes routing behaviour. One thing to check though is the correctness of the post-placement synchronization of the net terminals, which should be fine though as, if they weren't, there would most probably be a VTR assertion failure. In addition, the BRAM tile has only one equivalent site as well as only one sub tile;
- Changes introduced with this PR

acomodi:
@litghost I realized one other thing that is different w.r.t. symbiflow master is that we are now using the bucket router_heap. I have tried to remove it and use the default one and, at least locally, the failing bram_36_diff_fasm test passes correctly.

I think that might be the reason for the vendor tools to fail. Anyhow, I am unsure whether this is a strong failure, meaning that it could potentially be solved in fasm2bels, as probably, for this weird route through two FAN pips, the tcl won't include the fixed route to the REGCLKB site pin of the BRAM 36.

litghost:
> we are now using the bucket router_heap

We were using the bucket heap before, the difference is the default router heap was set to the bucket heap, and now it matches upstream, and the default router heap is binary again.  See https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/462/files#diff-f021ce5c0df49c7bf6020d0a4d94db66L1720

utils/lib/pb_type_xml.py
```diff
@@ -90,19 +100,209 @@ def add_switchblock_locations(xml):
...
+
+            equivalent_sites_xml.append(site_xml)
+            sub_tile_xml.append(equivalent_sites_xml)
+
+            tile_xml.append(sub_tile_xml)
+
+    return tile_xml
+
+
+def start_tile(
```
litghost, line 257: Thanks for splitting the functions, I think it looks much clearer.

xc/common/cmake/project_ray.cmake
```diff
@@ -606,19 +606,17 @@ function(PROJECT_RAY_TILE_CAPACITY)
...
+  #   SITE_TYPES <site types>
   #   )
   # ~~~
   #
-  # Import a tile from project xray that contains exactly 1 site type, and emit
-  # a tile with capacity for each instance of that site type.
-  #
-  # Future support: Emit tiles with capacity that contain more than 1 site type
-  # once VPR supports it.
+  # SITE_TYPES: contains the comma separated list of sites that are going to be used
```
litghost, line 613: Because SITE_TYPES is now a mutliValueArgs, this is simply a CMake list

acomodi:
@litghost Indeed, what I meant was that by leaving the default heap (which is the binary with the new master+wip) the vendor tools pass.

What remains to understand is why using the bucketed one (still to verify with CI though) the bram_test_36 fails.

litghost:
> @litghost Indeed, what I meant was that by leaving the default heap (which is the binary with the new master+wip) the vendor tools pass.

I expect randomness?  I doubt the default heap always passes.  It is worth making sure the bucket heap is completing faster than the binary heap, and by a significant amount.  If the binary heap is 1% slower, but always has a green CI, that is not the same situation as a 10-15% slower, right?

acomodi:
@litghost I would say we let this CI run finish and extract run-time data. As I suppose, the bucket heap should give us greater performance, in that case we may consider solving the weird jump on the FAN_BOUNCE in fasm2bels?
What I have experienced is that, for several runs, the bucket heap always failed that test, instead the binary one did succeed everytime.

litghost:
> in that case we may consider solving the weird jump on the FAN_BOUNCE in fasm2bels?

I've considered it.  But I wasn't off hand able to come up with a good closed form algorithm for doing so.

acomodi:
@litghost All right, then is it worth restoring the bucket heap (if it results to be required to maintain the great improvement w.r.t. the binary one), removing the diff fasm test from the bram_test_36 and opening an issue to keep track of this?

litghost:
> @litghost All right, then is it worth restoring the bucket heap (if it results to be required to maintain the great improvement w.r.t. the binary one), removing the diff fasm test from the bram_test_36 and opening an issue to keep track of this?

To be clear, I don't believe switching from the binary to bucket or vise-versa actually fixes the vendor tools test.  However if the vendor tools test is consistently green on binary, then let's do that.

litghost:
If CI is green, merge

acomodi:
`xc7 vendor tools` went green even though GH does report it as still running. Merging.

PR #1397, opened 2020-04-06
V2X as a submodule

mkurc-ant:
This PR makes the V2X a submodule. It also removes all V2X tests from arch-defs. They are part of the V2X repository.


mkurc-ant:
Need https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1386 to pass the CI

PR #1406, opened 2020-04-08
install: create tar of the installed toolchain

acomodi:

This PR compresses the installed toolchain directories to be uploaded on a cloud storage

### DIFF
--- a/.github/kokoro/continuous-install.cfg
+++ b/.github/kokoro/continuous-install.cfg
@@ -11,13 +11,7 @@ timeout_mins: 4320
 action {
   define_artifacts {
     # File types
-    regex: "**/*result*.xml"
-    regex: "**/*sponge_log.xml"
-    regex: "**/.ninja_log"
-    regex: "**/pack.log"
-    regex: "**/place.log"
-    regex: "**/route.log"
-    regex: "**/*_qor.csv"
+    regex: "**/symbiflow-arch-defs-install*.tar.gz"
     strip_prefix: "github/symbiflow-arch-defs-continuous-install/"
   }
 }
--- a/.github/kokoro/continuous-install.cfg
+++ b/.github/kokoro/continuous-install.cfg
@@ -11,7 +11,7 @@ timeout_mins: 4320
 action {
   define_artifacts {
     # File types
-    regex: "**/symbiflow-arch-defs-install*.tar.gz"
+    regex: "**/symbiflow-arch-defs-install*.tar.xz"
     strip_prefix: "github/symbiflow-arch-defs-continuous-install/"
   }
 }
--- a/.github/kokoro/continuous-install.cfg
+++ b/.github/kokoro/continuous-install.cfg
@@ -1,0 +1,38 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Generated from .github/kokoro/kokoro-cfg.py
+# To regenerate run:
+# cd .github/kokoro/ && python3 kokoro-cfg.py
+
+build_file: "symbiflow-arch-defs-continuous-install/.github/kokoro/install.sh"
+
+timeout_mins: 4320
+
+action {
+  define_artifacts {
+    # File types
+    regex: "**/*result*.xml"
+    regex: "**/*sponge_log.xml"
+    regex: "**/.ninja_log"
+    regex: "**/pack.log"
+    regex: "**/place.log"
+    regex: "**/route.log"
+    regex: "**/*_qor.csv"
+    strip_prefix: "github/symbiflow-arch-defs-continuous-install/"
+  }
+}
+
+env_vars {
+  key: "KOKORO_TYPE"
+  value: "continuous"
+}
+
+env_vars {
+  key: "KOKORO_DIR"
+  value: "symbiflow-arch-defs-continuous-install"
+}
+
+env_vars {
+  key: "SYMBIFLOW_ARCH"
+  value: "install"
+}
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -27,7 +27,9 @@ echo "========================================"
 echo "Compressing and uploading install dir"
 echo "----------------------------------------"
 (
-	tar -cf - ${INSTALL_DIR} | xz -9 -T`nproc` > ${ROOT_DIR}/symbiflow-arch-defs-install-$(git describe).tar.xz
+	pushd ${ROOT_DIR}
+	tar -cf install | xz -9 -T`nproc` > ${ROOT_DIR}/symbiflow-arch-defs-install-$(git describe).tar.xz
+	popd
 )
 echo "----------------------------------------"
 
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -3,6 +3,7 @@
 SCRIPT_SRC="$(realpath ${BASH_SOURCE[0]})"
 SCRIPT_DIR="$(dirname "${SCRIPT_SRC}")"
 INSTALL_DIR="$(pwd)/install"
+ROOT_DIR="$(pwd)"
 GIT_DESCRIBE=$(git describe)
 
 export CMAKE_FLAGS="-GNinja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR}"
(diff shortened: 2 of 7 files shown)


acomodi:
@mithro @litghost FYI, where and how the tarball should be uploaded?

.github/kokoro/continuous-install.cfg
```diff
@@ -0,0 +1,38 @@
...
+# To regenerate run:
+# cd .github/kokoro/ && python3 kokoro-cfg.py
+
+build_file: "symbiflow-arch-defs-continuous-install/.github/kokoro/install.sh"
+
+timeout_mins: 4320
+
+action {
+  define_artifacts {
+    # File types
```
mithro, line 13: Add the tar.gz to the artifacts here and it will be uploaded to GCS.

.github/kokoro/presubmit-install.cfg
```diff
@@ -0,0 +1,38 @@
+# Format: //devtools/kokoro/config/proto/build.proto
```
mithro, line 1: I don't think we want a presubmit install?

.github/kokoro/steps/hostsetup.sh
```diff
@@ -64,6 +64,7 @@ sudo apt-get install -y \
         python3-yaml \
         virtualenv \
         ninja-build \
+        pigz \
```
mithro, line 67: pigz?

.github/kokoro/install.sh
```diff
@@ -0,0 +1,45 @@
...
+	popd
+)
+echo "----------------------------------------"
+
+echo
+echo "========================================"
+echo "Compressing and uploading install dir"
+echo "----------------------------------------"
+(
+	tar -c --use-compress-program="pigz" -f install-${GIT_DESCRIBE}.tar.gz ${INSTALL_DIR}
```
mithro, line 28: The output binary should probably have symbiflow-arch-defs or symbiflow in the name somewhere?

.github/kokoro/steps/hostsetup.sh
acomodi: I found this to be compressing library that can run on multiple cores, unlike gzip. The files to compress can become pretty huge with bigger parts, hence the choice to use this instead of gzip

.github/kokoro/presubmit-install.cfg
acomodi: Problem is that having the make install target in xc7 is not entirely correct.
We may want to have in the install directory the various databases/rr_graphs/etc. of potentially other families, such as ice40 for instance.
It would not be consistent to have the install target in the xc7 directory.

acomodi: @mithro actually I am confused, what is the difference between presubmit and continuous?

mithro:
Presubmit -- runs on pull requests.
Continuous - runs after merged to master.

acomodi:
@mithro Ok, so IMO it would be best to have the install target in the presubmit actually, as there is also a test that verifies that everything is working with the installed toolchain.

.github/kokoro/presubmit-install.cfg
mithro: Presubmit -- run in pull requests.
Continuous -- run after the stuff has been merged to master branch.

.github/kokoro/steps/hostsetup.sh
mithro: SGTM!

Maybe you want to use lzma, 7z or bzip2 instead? (Or maybe we should publish them all in the same way that the Linux kernel does?)

.github/kokoro/install.sh
```diff
@@ -0,0 +1,45 @@
...
+	popd
+)
+echo "----------------------------------------"
+
+echo
+echo "========================================"
+echo "Compressing and uploading install dir"
+echo "----------------------------------------"
+(
+	tar -c --use-compress-program="pigz" -f symbiflow-arch-defs-install-${GIT_DESCRIBE}.tar.gz ${INSTALL_DIR}
```
litghost, line 29: Why not bz2 or lzma?

.github/kokoro/install.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+SCRIPT_SRC="$(realpath ${BASH_SOURCE[0]})"
+SCRIPT_DIR="$(dirname "${SCRIPT_SRC}")"
+INSTALL_DIR="$(pwd)/install"
+GIT_DESCRIBE=$(git describe)
+
+export CMAKE_FLAGS="-GNinja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR}"
```
litghost, line 8: Please use the explicit A200T disable option here

acomodi: If we are going to perform the install step in a separate ci build, IMO we could avoid having the disable/enable options and just run the make install for all the supported architectures.

Disabling the a200t was necessary as previously, make install was running on the xc7 target.

litghost: Ok

.github/kokoro/install.sh
acomodi: I have found pigz to be the fastest one, with a good compression ratio. 

The local install directory I have is ~12 Gb, while the compressed archive is ~2.5 Gb

litghost: Can you post your results table?  I'm willing to wait an hour to gain a large compression ratio drop.  Speed is not important here.

acomodi: Sure, I am capturing exact data now.

acomodi: @litghost This is what I have for now:

```
total uncompressed size: 12.2G

pigz:
    real    0m31.171s
    user    9m9.889s
    sys     0m11.734s

    compressed size: 2.5G
    compression ratio: 4.88
    space saving: 79.51%

bz2:
    real    15m16.055s
    user    15m12.854s
    sys     0m21.321s

    compressed size: 1.9G
    compression ratio: 6.4211
    space saving: 84.43%

lzma:
    real    46m39.288s
    user    46m43.604s
    sys     0m52.514s

    compressed size: 852M
    compression ratio: 14.6714
    space saving: 93.18%

```

I am running also `xz` compression which has the multithread implementation as well. Will update as soon as it finishes.

litghost: LZMA is definitely the way to go here.

acomodi: Actually found xz to be much faster with even better compression ratio than LZMA:

```
xz:
    real    5m47.481s
    user    91m35.688s
    sys     0m20.094s

    compressed size: 850M
    compression ratio: 14.7059
    space saving: 93.2%
```

litghost: Works for me!

litghost: > Question: how does the install kokoro cfg get triggered? I cannot see it running

Working on it

litghost: > > Question: how does the install kokoro cfg get triggered? I cannot see it running
> 
> Working on it

Should be working in 20 minutes, hopefully.

acomodi:
There appear to be issues with packagin the installed directory. I have removed the a200t build to speed things up and get to a stable state.

Once install is working and stable we could add back the 200T

kgugala:
> There appear to be issues with packagin the installed directory. I have removed the a200t build to speed things up and get to a stable state.
> 
> Once install is working and stable we could add back the 200T

the install target passed, but I don't see any install artifacts

acomodi:
@kgugala https://console.cloud.google.com/storage/browser/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/presubmit/install/19/20200414-132809

There is something wrong actually with the naming of the tarball. I think `git describe` needs to be called just before compressing the install dir

acomodi:
Fixed all the issues with the install CI. I think a review pass can be made now.

Here is the installed package: https://console.cloud.google.com/storage/browser/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/presubmit/install/24/20200415-081225

litghost:
I see that xc7 install job has complete, did that compile without A200T?

acomodi:
@litghost yes, I disabled it for now. IMO if this PR goes green we should merge it and enable the 200T after https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1415 lands as well.

The main reason was because there were issues with packaging the tarball and the 200t did take too long, slowing down the whole process.

I could re-add the 200T once again if you prefer

mithro:
Let's land without 200T and enable it separately.

.github/kokoro/presubmit-install.cfg
litghost: We want both to ensure that PR's don't break it once submitted.

acomodi:
CI green, merging

mithro:
First output! -> https://console.cloud.google.com/storage/browser/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/continuous/install/4/20200416-002215/

PR #1407, opened 2020-04-08
A VPR placement visualization utility.

mkurc-ant:
This is a small utility script I wrote. It allows to visualize VPR placement by parsing the `.place` file only.

An example visualization (this is for Quicklogic EOS S3):
placement

### DIFF
--- a/utils/vpr_place_view.py
+++ b/utils/vpr_place_view.py
@@ -98,9 +98,6 @@ def generate_image(grid_size, placement, block_size=8, colormap=None):
 
 
 def main():
-    """
-    Main.
-    """
 
     # Parse arguments
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


utils/vpr_place_view.py
```diff
@@ -0,0 +1,140 @@
...
+
+    return image
+
+
+# =============================================================================
+
+
+def main():
+    """
+    Main.
```
mithro, line 102: Improve docstring or delete it.

utils/vpr_place_view.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python3
+"""
```
mithro, line 2: Would be nice to add an output image to the pull request.

utils/vpr_place_view.py
mkurc-ant: Fixed.

utils/vpr_place_view.py
mkurc-ant: Added.

PR #1417, opened 2020-04-14
Add techmap for LDCE

rw1nkler:
This commit adds missing LDCE latch techmap.
The Xilinx documentation of the primitive is available in  [[HERE]](https://www.xilinx.com/support/documentation/sw_manuals/xilinx13_2/7series_scm.pdf#page=285)

### DIFF
--- a/xc/xc7/techmap/cells_map.v
+++ b/xc/xc7/techmap/cells_map.v
@@ -180,6 +180,21 @@ FDPE_ZINI #(.ZINI(!|INIT), .IS_C_INVERTED(|1))
 endmodule
 
 // ============================================================================
+// Latches
+
+module LDCE (
+   output reg Q,
+   input CLR, D, G, GE
+);
+
+parameter [0:0] INIT = 1'b0;
+
+LDCE_ZINI #(.ZINI(!INIT))
+  _TECHMAP_REPLACE_ (.Q(Q), .CLR(CLR), .D(D), .G(G), .GE(GE));
+
+endmodule
+
+// ============================================================================
 // LUTs
 
 module LUT1(output O, input I0);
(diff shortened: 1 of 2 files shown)


andrewb1999:
@litghost is there any reason this pull request hasn't been merged other than a rebase? I have recently run into the issue with LDCEs being generated instead of LDCE_ZINIs and this seems to be the solution.

litghost:
> @litghost is there any reason this pull request hasn't been merged other than a rebase? I have recently run into the issue with LDCEs being generated instead of LDCE_ZINIs and this seems to be the solution.

I believe remember, but I believe this was not enough for LDCE support.  In general, latches are not something that should be present in pure FPGA designs, so we haven't prioritized implementing support for them.  Are you confident that your design should contain latches?

andrewb1999:
> I don't remember, but I believe this was not enough for LDCE support. In general, latches are not something that should be present in pure FPGA designs, so we haven't prioritized implementing support for them. Are you confident that your design should contain latches?

I am definitely not convinced that the design should have latches, but they are inferred by yosys. Is there something I should be doing to prevent yosys from inferring them?

litghost:
> > I don't remember, but I believe this was not enough for LDCE support. In general, latches are not something that should be present in pure FPGA designs, so we haven't prioritized implementing support for them. Are you confident that your design should contain latches?
> 
> I am definitely not convinced that the design should have latches, but they are inferred by yosys. Is there something I should be doing to prevent yosys from inferring them?

Latches occur when you improperly define the process/state transition.  I'm not an expert in HDL, here is an article that speaks to this kind of thing: https://www.nandland.com/articles/how-to-avoid-transparent-latches-in-vhdl-and-verlog.html

Issue #1172, opened 2019-11-25
Make a Yosys plugin that does the inout port split


mkurc-ant:
Since BLIF file cannot handle inout ports, these are represented as a single net connected to both input and output port with the same name. This creates an issue when something drives that net as there are multiple drivers there.

Currently to solve this we use the following flow:
- Synthesize the design using Yosys but write it to JSON instead of BLIF.
- Process the JSON using a python script that splits inout ports to separate input and output ports.
- Load the JSON to Yosys again and write the final BLIF (this is essentially a format conversion).

The "as it should be done" solution would be to have a plugin for Yosys that does the inout port split. That would allow to get rid of the need for writing JSONs and processing them.

mithro:
We already have eBLIF -- could we just extend that?

mkurc-ant:
One possibility is to add let's say an `.inout` model to the BLIF file. But that's only one thing. We would then need to make VPR support inout ports which it does not right now. I can't tell how much work would that require.

PR #1461, opened 2020-05-02
Avoid calling yosys twice from synth script

olofk:
For Edalize, it would be preferred to call yosys directly with an extra tcl file and potentially some env vars set instead of calling the synth script. This patch makes it easier to accomplish that by calling the convertion script from inside yosys instead of relying on bash doing that


olofk:
I have absolutely no clue what these failing tests mean, so could you let me know if there is anything else I should do with the PR? Like update some test cases or so

xc/xc7/yosys/conv.tcl
```diff
@@ -1,20 +0,0 @@
-yosys -import
```
acomodi, line 1: @olofk The failure is due to the absence of the conv.tcl.
I see you have moved it to the synth.tcl so to have everything computed in one call only. The problem with this is the current handling of dependencies in cmake, as the conv.tcl script cannot be found anymore, the build fails.
[devices.cmake](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/cmake/devices.cmake) holds all the cmake functions to create the devices targets, and there are the two separate commands to call the synth.tcl and conv.tcl.


IIRC the original reason to split synth and conv scripts was to apply the `split_inouts` script between the two, as the inouts signals need to be preprocessed to continue with the further stage of the build, @mkurc-ant can confirm that.

olofk: ok, so thinking about it again, the python script+conv.tcl together basically forms a json -> eblif bridge. I can see the case for having this separate.

We can drop this change and I will take care of it on the edalize side instead

mkurc-ant: I agree, that's the reason for that. Ultimately we would want the split process to be part of the Yosys flow. This could be done via a plugin: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1172

PR #1466, opened 2020-05-06
PicoSoC and Murax for Nexys Video

mkurc-ant:
This PR adds targets for PicoSoC and Murax to run on Nexys Video board.

I've also cleaned up cmakes for these two SoCs and added separate targets for running them with 50MHz clock and 100MHz clock. I've also added separate firmwares for PicoSoC for 50MHz and 100MHz to have the uart baudrate correct. Didn't do that for Murax as generation of its ROM content is more problematic and it always had the baudrate dependent on the clock.

All have been checked on actual HW and work. Even though that VPR reports max frequency to be ~50MHz both work fine on 100MHz.


mkurc-ant:
With this PR I added designs that intend to run these SoCs at 50MHz. I used logic to divide the input clock. It looks like this makes fasm2bels create antenna nets. I've tried to run the divided clock through a BUFG but it makes the clock placement script crash.

What's the best course of action: fix the placement script to allow logic -> BUFG, use a PLL to get 50MHz clock? Or maybe leave only the design that runs on 100MHz ??

litghost:
> a BUFG but it makes the clock placement script crash.

Fix the clock placement script?  These large fabrics likely have some clock placement cases that were not considered before?

tcal-x:
I've been seeing this too.
The problem seems to be that Vivado clock tree optimization adds the BUFG,
which then conflicts with the placement constraints in top_bit.v.tcl.
I've been working around this issue by disabling optimization in Vivado so
that the BUFG is not added.
I know how to disable the optimization in the GUI (I'm doing this flow
manually)
but haven't tried to replicate this in the Vivado script we use.
Tim

litghost:
Whenever using yosys + symbiflow, BUFG's need to explicitly instanced for now.  It is worth noting that yosys has BUFG insertion logic, but I don't believe we should enable that logic until yosys used in symbiflow-arch-defs has been sync'd with the upstream yosys.

To be clear, if Vivado is inserting a BUFG's during vivado diff fasm processing (e.g. output from fasm2bels back into Vivado), the solution is to add the missing BUFG's into the input to yosys before the VPR flow happens.  Do not disable Vivado's BUFG's insertion behavior.

mkurc-ant:
So the problem here was that the script that places clock resources didn't allow a generic (non-clock) net to be source for a BUFG. I've changed it so now it allows for that. Let's wait for the CI.

litghost:
You mean interconnect?  Why would you want an interconnect based clock?

mkurc-ant:
> You mean interconnect? Why would you want an interconnect based clock?

I've made a clock divider using logic to get from 100MHz to 50MHz. Then to get from logic to the clock network a BUFG is needed. Didn't want to use a PLL not to overcomplicate these examples.

litghost:
> > You mean interconnect? Why would you want an interconnect based clock?
> 
> I've made a clock divider using logic to get from 100MHz to 50MHz. Then to get from logic to the clock network a BUFG is needed. Didn't want to use a PLL not to overcomplicate these examples.

There is never a reason to rely on logic to divide the clock.  Use a BUFR (which can divide) or a PLL/MMCM.  A BUFGCTRL is also a supported clock divider.

litghost:
@mkurc-ant Please rebase this PR on master

mkurc-ant:
I've made the allowance of BUFG being driven by a clock generated using logic optional. Also added a techmap for BUFGCE and used it to divide 100MHz clock by 2 as described in https://forums.xilinx.com/t5/Other-FPGA-Architecture/How-to-divide-a-clock-by-2-with-a-simple-primitive-without-Clock/m-p/784079

tests/9-soc/murax/basys3-full_toplevel_50.v
```diff
@@ -0,0 +1,48 @@
...
+  wire clk100;
+  BUFG bufg(.I(clk), .O(clk100));
+
+  // BUFGCE as divide by 2
+  reg clk50_ce;
+  always @(posedge clk100)
+      clk50_ce <= !clk50_ce;
+
+  wire clk50;
+  BUFGCE bufg50 (.I(clk), .CE(clk50_ce), .O(clk50));
```
litghost, line 20: Thanks!

tests/9-soc/picosoc/CMakeLists.txt
```diff
@@ -7,28 +9,28 @@ add_file_target(FILE spimemio.v SCANNER_TYPE verilog)
...
-add_file_target(FILE hx8k-b-evn.pcf)
+get_target_property_required(PYTHON3 env PYTHON3)
+get_target_property(PYTHON3_TARGET env PYTHON3_TARGET)
 
-add_file_target(FILE basys3_demo.v SCANNER_TYPE verilog)
-add_file_target(FILE basys3_demo_no_roi.v SCANNER_TYPE verilog)
-add_file_target(FILE basys3_demo_tb.v SCANNER_TYPE verilog)
-add_file_target(FILE basys3.pcf)
-add_file_target(FILE basys3.sdc)
+file(GLOB HEX_FILES *.hex)
```
litghost, line 15: Do not use GLOB

tests/9-soc/picosoc/basys3-full_demo_100.v
```diff
@@ -27,16 +27,18 @@ module basys3_demo (
 	output [15:0] led
 );
 
-	wire clk_bufg;
-	BUFG bufg (.I(clk), .O(clk_bufg));
+    // Input 100MHz clock through a BUFG
+    wire clk100;
+    BUFG bufg100 (.I(clk), .O(clk100));
```
litghost, line 32: Mixed tabs and spaces here?

xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -322,9 +333,18 @@ def __init__(self, vpr_grid, io_locs, blif_data, roi):
...
-                    )
+                    if sink_net not in self.input_pins and sink_net not in self.clock_sources:
+
+                        # Allow BUFGs to be driven by generic sources but only
+                        # when enabled.
+                        if bel == "BUFGCTRL_VPR" and allow_bufg_logic_sources:
+                            continue
+
+                        # The clock source comes from logic, disallow that
+                        eprint(
```
litghost, line 344: This is not actually an error, and it doesn't disallow it?

tests/9-soc/picosoc/basys3-full_demo_100.v
mkurc-ant: Fixed

tests/9-soc/picosoc/CMakeLists.txt
mkurc-ant: What's wrong about using GLOB? I didn't wanted to hard code names of HEX files.

xc/common/utils/prjxray_create_place_constraints.py
mkurc-ant: I added an exit statement below.

When we talked about it we wanted to generally disallow logic driving clock sinks. There is an exception though for a `BUFGCTRL_VPR` but only enabled when the flag `allow_bufg_logic_sources` is set.

tests/9-soc/picosoc/basys3-full_demo_100.v
mkurc-ant: @litghost ping?

xc/common/utils/prjxray_create_place_constraints.py
litghost: There is still no exit?

tests/9-soc/picosoc/CMakeLists.txt
litghost: We briefly discussed why GLOB should not be used.  Please remove it's usage.

mkurc-ant: Ok, I've fixed that.

xc/common/utils/prjxray_create_place_constraints.py
mkurc-ant: Unclear what happened before but the exit statement should be there now.

tests/9-soc/picosoc/CMakeLists.txt
```diff
@@ -7,28 +9,34 @@ add_file_target(FILE spimemio.v SCANNER_TYPE verilog)
...
-    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/hex2progmem.py ${PYTHON3} ${PYTHON3_TARGET}
+set(HEX_FILES
+    firmware.hex
+    firmware_noflash_100.hex
+    firmware_noflash_50.hex
+    firmware_noflash_25.hex
 )
 
-add_file_target(FILE progmem.v GENERATED)
+foreach(HEX_FILE ${HEX_FILES})
```
litghost, line 22: Perfect, thanks!

xc/xc7/techmap/cells_map.v
```diff
@@ -2985,6 +2985,25 @@ module BUFGCE (
   );
 endmodule
 
+module BUFGCE (
```
litghost, line 2988: Is this in the cells_map.v twice now?

mkurc-ant: Right, fixed.

mkurc-ant:
@litghost A200T is green

Issue #1442, opened 2020-04-24
OpenTitan support


rw1nkler:
Here is the branch with the current progress on OpenTitan support:
https://github.com/antmicro/symbiflow-arch-defs/tree/opentitan_earlgray

Currently, I'm struggling with the error in `vpr_io_place.py`:
```
Generating earlgray_nexys_video/artix7_200t-xc7a200t-virt-xc7a200t-test/top_io.place
Traceback (most recent call last):
  File "/home/build/rwinkler/google-symbiflow-arch-defs/xc/common/utils/prjxray_create_ioplace.py", line 159, in <module>
    main()
  File "/home/build/rwinkler/google-symbiflow-arch-defs/xc/common/utils/prjxray_create_ioplace.py", line 73, in main
    io_place.load_block_names_from_net_file(args.net)
  File "/home/build/rwinkler/google-symbiflow-arch-defs/utils/vpr_io_place.py", line 60, in load_block_names_from_net_file
    "//block[@instance='inpad[0]'] | //block[@instance='outpad[0]']"
  File "src/lxml/etree.pyx", line 1581, in lxml.etree._Element.xpath
  File "src/lxml/xpath.pxi", line 305, in lxml.etree.XPathElementEvaluator.__call__
  File "src/lxml/xpath.pxi", line 225, in lxml.etree._XPathEvaluatorBase._handle_result
lxml.etree.XPathEvalError: Error in xpath expression
xc/xc7/tests/soc/earlgray/CMakeFiles/file_xc_xc7_tests_soc_earlgray_earlgray_nexys_video_artix7_200t-xc7a200t-virt-xc7a200t-test_top_io.place.dir/build.make:71: recipe for target 'xc/xc7/tests/soc/earlgray/earlgray_nexys_video/artix7_200t-xc7a200t-virt-xc7a200t-test/top_io.place' failed
make[3]: *** [xc/xc7/tests/soc/earlgray/earlgray_nexys_video/artix7_200t-xc7a200t-virt-xc7a200t-test/top_io.place] Error 1
make[3]: *** Deleting file 'xc/xc7/tests/soc/earlgray/earlgray_nexys_video/artix7_200t-xc7a200t-virt-xc7a200t-test/top_io.place'
[...]

PR #1483, opened 2020-05-12
Allow import scripts to parse huge .net files

rw1nkler:
This commit changes the architecture importing scripts and allows them to parse huge .net files. Without these changes, libxml fails on using XPath expressions due to the internal memory allocation limit. In that situation, the python wrapper on the libxml library throws misleading error message about incorrect XPath expression.

This changes are required for OpenTitan and are related to https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1442#issuecomment-619055675

### DIFF
--- a/utils/vpr_io_place.py
+++ b/utils/vpr_io_place.py
@@ -56,9 +56,11 @@ class IoPlace(object):
         net_root = net_xml.getroot()
         self.net_to_block = {}
 
-        for block in net_root.xpath(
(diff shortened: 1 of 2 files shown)


litghost:
The description of the commit and the PR is rather poor.  Please re-write it.  These changes are required for OpenTitan, but have nothing to do with OpenTitan per say.

litghost:
@HackerFoo Any additional thoughts?

PR #1484, opened 2020-05-13
xc7: avoid splitting CLBs

acomodi:

This PR is an initial attempt to avoid the split of CLBL tiles into SLICEL and SLICEM tiles, but have the m defined as sub tiles.

This requires https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1298 to correctly work, as there are the fixes for some bugs found during developement of this PR.


acomodi:
@litghost Indeed, I think that a major reorganization can be done here, so we can have, for instance, one script to generate the physical tiles and one to generate the logical blocks (to mimic what happens currently in VPR).

The only thing I might have some doubts are the IOPADs. With the current implementation (correct me if I am wrong), the IOB and IOI canonical tiles are merged to form one placement location instead of two. I think this was originally designed to have the IOPADs packed all together in the same cluster. How about we move to use the canonical locations and having something similar with the carry chains that are packed in macros, so that the IOB and IOI placement gets to be satisfied.

acomodi:
A200T failed with the following:

```
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t/xc/common/utils/prjxray_routing_import.py", line 1410, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t/xc/common/utils/prjxray_routing_import.py", line 1397, in main
    node_remap=node_remap,
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t/utils/lib/rr_graph_capnp/graph2.py", line 587, in serialize_to_capnp
    rr_graph.write(f)
  File "capnp/lib/capnp.pyx", line 1178, in capnp.lib.capnp._DynamicStructBuilder.write
  File "capnp/lib/capnp.pyx", line 3897, in capnp.lib.capnp._write_message_to_fd
capnp.lib.capnp.KjException: kj/io.c++:362: overloaded: ::writev(fd, current, iov.end() - current): No space left on device; fd = 5
```

Apparently it consumed all space on the device. I need to check whether this is due to this PR or is a general issue

utils/lib/pb_type_xml.py
```diff
@@ -100,6 +100,19 @@ def add_switchblock_locations(xml):
     })
 
 
+def get_site_pin_wire(equivalent_site_pin, site_pins):
+    no_prefix_equivalent_site_pin_wire = equivalent_site_pin.wire.split('_'
```
litghost, line 104: This needs a docstring

acomodi: Done, actually I noticed I had a way to complex logic to check the site pin names, I have simplified that alongside with the docstrings addition.

utils/lib/pb_type_xml.py
```diff
@@ -100,6 +100,27 @@ def add_switchblock_locations(xml):
...
+    We make the assumption that the name of the pin (prior to stripping the tile name prefix)
+    is the same, both for the original and the equivalent sites:
+    """
+
+    for site_pin in site_pins:
+
+        if site_pin.name == equivalent_site_pin.name:
+            return site_pin.wire
+
+    assert True, "No equivalent pin found!"
```
litghost, line 121: Should be "assert False"?

acomodi: Indeed, fixed. Thanks for spotting this

acomodi:
@litghost I have checked only the baselitex test, but given that it combines almost all of the resources supported (IOSERDESes, PLL, BUFG, SLICEM/L, etc.), I think it is enough.

Linux boots correctly via ethernet as well.

acomodi:
On a side note, P&R runtimes are in line with the master's ones, so no neat improvement/regression with baseline (at least for the baselitex design).

litghost:
> On a side note, P&R runtimes are in line with the master's ones, so no neat improvement/regression with baseline (at least for the baselitex design).

We aren't expecting runtime improvements yet, we are just simplifying the import and removing code and complexity (once cleaned up).

PR #1497, opened 2020-05-20
channels: fix vpr grid map generation

acomodi:

This PR fixes the vpr grid map from vpr locations to canonical locations.


acomodi:
Vendor tools failed for the VCC_WIRE incorrect route as always, all the rest passed. Merging

PR #1502, opened 2020-05-25
Support for IDDR and ODDR

mkurc-ant:
This PR adds support for IDDR and ODDR primitives.
- [x] IDDR architecture
- [x] IDDR techmap(s)
- [x] IDDR fasm2bels
- [x] IDDR tests
- [x] ODDR architecture
- [x] ODDR techmaps
- [x] ODDR fasm2bels
- [x] ODDR tests
- [x] Timings for both
- [x] How to map `IDDR` to `IDDR_2CLK`
- [x] More pack patterns
- [x] Add a test that can be run on HW

The `IDDR` is added as another mode of the ILOGICE3 pb_type. The connection `IDELAY`->`IDDR` is modeled as a `mux` - no need for separate pb_types.

An `ODDR` can be used to drive either `O` or `T` of an `(I)OBUFT(DS)`. Two independent `ODDR`s can be packed into the same `OLOGIC` site to drive `O` and `T` independently.

For that the `OLOGICE3` pb_type has two modes: `OSERDESE2` and `OLOGIC`. In the second mode there are two almost identical child pb_types - one for `ODDR` for `OQ` and one for `ODDR` for `TQ`. These pb_types can operate independently either in the `ODDR` mode or a pass-through mode. This allows to have all combinations of `ODDR` uses.

Depending on where an `ODDR` is used, different fasm features have to be emitted. Fortunately for `O` and `T` they have one-to-one correspondences. The same blif models for both could be used. They are however nested i different pb_types that provide necessary parameters to fasm features mappings.

For the updated `OLOGICE3` model to work the `T` connection for `OBUF` had to be made explicit. The IOB architecture has been modified to have the `OBUFT` by default. That allowed 'O' and 'T' paths to be split and made fully independent.

The issue with Vivado routing OBUFT.T to const1 + enabling inversion while it is logically connected to const0 has been fixed according to the suggestion in the issue: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1514


mkurc-ant:
I didn't see any CI errors besides https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1400 so I rebased it on top of master. Now the CI should pass.

xc/xc7/fasm2bels/iob_models.py
```diff
@@ -465,9 +468,67 @@ def process_single_ended_iob(top, iob):
...
     top.add_site(site)
 
 
+def cleanup_single_ended_iob(top, site):
+    """
+    Cleans up a single-ended IOB
+    - Detects whether there is an OBUFT with T input routed to const1. If so
+      then checks whether the ZINV_T1 inverter is enabled in the neighboring
+      OLOGIC site. If both cases are true then the OBUFT is replaced with an
+      OBUF.
```
litghost, line 481: Good comment, thanks!

xc/xc7/fasm2bels/iob_models.py
```diff
@@ -465,9 +468,67 @@ def process_single_ended_iob(top, iob):
...
+    if bel is None:
+        return
+
+    # Check if its T input is routed to const1
+    src = top.find_source_from_sink(site, "T")
+    if src != 1:
+        return
+
+    # Get the neighboring IOI3 tile prefix and loc
+    ioi_tile_loc = site.tile.rsplit("_", maxsplit=1)[1]
```
litghost, line 495: Rather than doing a string replace here, why not walk 1 tile to the left or right?  That would be more robust?

xc/xc7/fasm2bels/ioi_models.py
```diff
@@ -23,7 +25,60 @@ def get_ioi_site(db, grid, tile, site):
...
+
+    # Check if we have an IDDR
+    bel = site.maybe_get_bel("IDDR")
+    if bel is not None:
+
+        # Check if the SR is const
+        source = top.find_source_from_sink(site, 'SR')
+
+        # If so then connect both 'S' and 'R' of the BEL to the same const.
+        if source in [0, 1]:
```
litghost, line 48: Should this also have a ` and len(source) == 2`?

xc/xc7/fasm2bels/ioi_models.py
```diff
@@ -87,7 +142,183 @@ def process_idelay(top, features):
...
+    bel = Bel('ISERDESE2')
+
+    data_rate = None
+    if site.has_feature("ISERDES.DATA_RATE.SDR"):
+        data_rate = '"SDR"'
+    else:
+        data_rate = '"DDR"'
+    bel.parameters['DATA_RATE'] = data_rate
+
+    # TODO: There shouldn't be mixed width in FASM features.
```
litghost, line 160: This feels like it could crop up in a new fasm2bels failure in the future.  Is it possible to add a failing test now to demostrate the issue you are mentioning here?

xc/xc7/techmap/cells_map.v
```diff
@@ -2496,33 +2566,33 @@ module OSERDESE2 (
...
   parameter DATA_WIDTH = 4;
   parameter SERDES_MODE = "MASTER";
-  parameter TRISTATE_WIDTH = 4;
+  parameter TRISTATE_WIDTH = 1;
 
   if (DATA_RATE_OQ == "DDR" &&
       !(DATA_WIDTH == 2 || DATA_WIDTH == 4 ||
         DATA_WIDTH == 6 || DATA_WIDTH == 8)) begin
-    wire _TECHMAP_FAIL_ = 1'b1;
+    wire _TECHMAP_FAIL_;
```
litghost, line 2577: Why did you change the _TECHMAP_FAIL_ here?  Has yosys's behavior around this feature changed?

xc/xc7/techmap/cells_map.v
```diff
@@ -2715,64 +2788,114 @@ module OSERDESE2 (
     wire t4 = T4;
   end endgenerate
 
-  OSERDESE2_VPR #(
```
litghost, line 2718: Can you explain in words why we need both `OSERDESE2_VPR` and `OSERDESE2_NO_TQ_VPR`

litghost:
This PR needs to be rebased.

xc/xc7/techmap/cells_map.v
mkurc-ant: Yosys behavior hasn't changed. The thing is when you declare the `_TECHMAP_FAIL_` wire and assign it `1` then the tehcmap fails silently and the un-mapped cell remains. But when you don't assign it anything then the Yosys throws an error. The former is a way of telling Yosys that a techmap is not applicable. I think that the latter is more what we want here (and probably everywhere).

xc/xc7/techmap/cells_map.v
mkurc-ant: So as far as I understand when having an `OSERDES+OBUFT` the `T` connection in the design does not have to pass through the `OSERDES` (can be driven by something else). This implies that the `OSERDES` works in the "T bypass mode". That also happens when the `T` is implicitly or explicitly connected to a constant.

I have changed the arch so the 'T' connection is always explicitly present and routed by VPR. Since the connection in the netlist does not pass through the `OSERDES` cell I had to make a variant that allows bypassing the "T" section.

xc/xc7/fasm2bels/ioi_models.py
mkurc-ant: @litghost You mean add it and then disable? (we want green CI). I can think of a test with data width of eg. 4 that cannot be decoded now.

xc/xc7/fasm2bels/ioi_models.py
mkurc-ant: No, that is not the way it works. Here I check whether the `SR` sink is routed to a constant and if so then I assign it to both to `S` and `R` ports of the `IDDR` cell.

`IDDR` cell has `S` and `R` ports but physically they correspond to a single `SR` input and an appropriate bit controlling whether it is set or reset. I've found in the documentation that when neither reset not set are used, both should be connected to a const (logically).

xc/xc7/fasm2bels/iob_models.py
mkurc-ant: Fixed.

mkurc-ant:
@litghost I've discovered an issue with handling of `IDDR` vs. `IDDR_2CLK`.

In the techmap I map `IDDR` to the two clock version by connecting both `C` and `CB` to the same signal and setting the `IS_CB_INVERTER` parameter. Since there are no bits for that I wonder how should I handle this. For an `IDDR` Vivado always show that both `CLK` and `CLKB` of the site are connected to the same net and that the CLKB inverter is enabled.

Similarly, there is an issue in fasm2bels - whenever I decode a `IDDR_2CLK` with `C` and `CB` connected to the same net, Vivado complains that this is illegal.

acomodi:
@mkurc, as far as I understood with the ISERDES, the CLKB inverter is not enabled by any feature, with, or without it, the bitstream remains the same. What I have done in fasm2bels, at the time, was to add the IS_CLKB_INVERTED parameter to the ISERDES, but still assign the same net to CLK and CLKB. This avoids Vivado complaining about the clock signals being the same.

mkurc-ant:
@acomodi Thanks, I'll try that.

mkurc-ant:
@acomodi That did the trick! I now have IDDR and ODDR working in hardware as well as correct decoding in fasm2bels.

Although there is a fasm difference:
```
 HCLK_CMT_X8Y26.HCLK_CMT_CK_BUFHCLK9_USED

 HCLK_CMT_L_X106Y26.HCLK_CMT_CCIO0_ACTIVE
-HCLK_CMT_L_X106Y26.HCLK_CMT_CCIO0_USED
 HCLK_CMT_L_X106Y26.HCLK_CMT_CK_BUFHCLK6_USED
 HCLK_CMT_L_X106Y26.HCLK_CMT_CK_IN10.HCLK_CMT_MUX_CLK_PLL0
```

But I probably know the cause, we shouldn't emit `CCIOn_USED` feature for routes through `HCLK_CMT_MUX_OUT_FREQ_REFn`.

mkurc-ant:
Ok, fixed that too. Let's see what the CI has to say.

xc/xc7/techmap/cells_map.v
litghost: I think you are depending on undocumented yosys behavior.  If you want to generate an error, I believe yosys has now implemented `$error`

xc/xc7/techmap/cells_map.v
litghost: I'm not sure I follow yet.  Do you mind make diagrams of the various cases you are talking about?  I'll keep reading the verilog to see if I can follow what you are saying.

xc/common/primitives/ologice3/ologice3.pb_type.xml
```diff
@@ -32,24 +32,183 @@
...
+    <pb_type name="OLOGIC_TFF" num_pb="1">
+      <input  name="TCE" num_pins="1"/>
+      <input  name="T1"  num_pins="1"/>
+      <input  name="T2"  num_pins="1"/>
+      <input  name="SR"  num_pins="1"/>
+      <input  name="CLK" num_pins="1"/>
+      <output name="TQ"  num_pins="1"/>
+      <output name="TFB" num_pins="1"/>
+
+      <mode name="PASSTHROUGH">
```
litghost, line 49: Rather than model T1 -> TQ as a mode, couldn't this be modelled as a mux?  Because that is likely what is happening underneath, yes?

xc/xc7/fasm2bels/ioi_models.py
litghost: Your right, resolving.

xc/xc7/fasm2bels/ioi_models.py
litghost: > @litghost You mean add it and then disable? (we want green CI). I can think of a test with data width of eg. 4 that cannot be decoded now.

That is one way to go.  I think the valid options are:
 - Create and issue, and attach a failing test case so that it can be fixed in the future
 - Create the failing test case and fix it in this PR

The TODO by itself is insufficient.

xc/xc7/techmap/cells_map.v
litghost: I think I understand what you are describing now, and believe the better approach here is to express the `OSERDES.DATA_RATE_TQ.BUF` vs  `OSERDES.DATA_RATE_TQ.DDR` as a VPR routing mux.  Can you explain why you took a mode based approach, rather than a VPR routing mux?

mkurc-ant: Apart from your question, I was mistaken about the `T` connection bypassing `OSERDES` When you do something like that then Vivado complains so it is not legal:

```
ERROR: [DRC THRU-1] OSERDES Route-thru Availability: Output buffer obuf_5 is driven by an OSERDES cell on the I pin and from fabric on the T pin. The current netlist does not have a legal routing. The OSERDES is used and no concurrent route-thru from fabric is supported. To resolve this violation, modify the design so that the output buffer pins are driven from the OSERDES (e.g. connect the OSERDES TQ port to the OBUF* T port).
```
I need to rethink that again. Maybe a routing mux will be applicable here.

xc/xc7/techmap/cells_map.v
mkurc-ant: This is not undocumented behavior, see manual for the `techmap` command:
```
    _TECHMAP_FAIL_
        When this wire is set to a non-zero constant value, techmap will not
        use this module and instead try the next module with a matching
        'techmap_celltype' attribute.

        When such a wire exists but does not have a constant value after all
        _TECHMAP_DO_* commands have been executed, an error is generated.
```

xc/xc7/fasm2bels/ioi_models.py
mkurc-ant: I'll go with an issue then. Solving it probably requires adjustment to prjxray fuzzers, and I think that this PR is already kind of big.

xc/common/primitives/ologice3/ologice3.pb_type.xml
mkurc-ant: So since Vivado actually does not allow to bypass T on an OSERDES then I should probably use a mux there. But I need to rethink it:

```
ERROR: [DRC THRU-1] OSERDES Route-thru Availability: Output buffer obuf_5 is driven by an OSERDES cell on the I pin and from fabric on the T pin. The current netlist does not have a legal routing. The OSERDES is used and no concurrent route-thru from fabric is supported. To resolve this violation, modify the design so that the output buffer pins are driven from the OSERDES (e.g. connect the OSERDES TQ port to the OBUF* T port).
```

xc/xc7/techmap/cells_map.v
litghost: You didn't set a value at all.  Yosys generating an error when `_TECHMAP_FAIL_` has no value is an undocumented behavior.  The previous code set `_TECHMAP_FAIL_` to 1, which is non-zero, which what the documentation called for.  You are depending on undocumented behavior.

mkurc-ant:
@litghost I've removed the `OSERDES` "without T" and used a routing mux to make the `OLIGICE3.TQ` output select from `OSERDES.TQ` and the synthetic T inverter `T_INV`.

xc/common/primitives/ologice3/ologice3.pb_type.xml
```diff
@@ -180,6 +339,12 @@
         </meta>
       </metadata>
     </pb_type>
+
+    <pb_type name="T_INV" blif_model=".subckt T_INV" num_pb="1">
```
litghost, line 343: You probably need a combinatorial delay = "0.0" throught this thing to prevent breaking timing arcs.

xc/xc7/techmap/cells_map.v
```diff
@@ -1979,7 +1979,7 @@ module IBUF (
 endmodule
 
 module OBUF (
-  input I,
+  input  I,
```
litghost, line 1982: Extra space?

xc/common/primitives/ologice3/ologice3.pb_type.xml
```diff
@@ -206,23 +371,14 @@
       <direct name="TBYTEOUT" input="OSERDESE2.TBYTEOUT" output="OLOGICE3.TBYTEOUT" />
       <direct name="TCE" input="OLOGICE3.TCE" output="OSERDESE2.TCE" />
       <direct name="TFB" input="OSERDESE2.TFB" output="OLOGICE3.TFB" />
-      <direct name="TQ" input="OSERDESE2.TQ" output="OLOGICE3.TQ" />
       <direct name="SHIFTIN1" input="OLOGICE3.SHIFTIN1" output="OSERDESE2.SHIFTIN1" />
       <direct name="SHIFTIN2" input="OLOGICE3.SHIFTIN2" output="OSERDESE2.SHIFTIN2" />
+
+      <direct name="T1_to_T_INV.TI" input="OLOGICE3.T1" output="T_INV.TI"/>
+      <mux name="TQ" input="OSERDESE2.TQ T_INV.TO" output="OLOGICE3.TQ"/>
```
litghost, line 378: Don't you need a "fasm_mux" annotation here?

litghost:
> @litghost I've removed the `OSERDES` "without T" and used a routing mux to make the `OLIGICE3.TQ` output select from `OSERDES.TQ` and the synthetic T inverter `T_INV`.

I think the new structure looks better, and has less new primatives.  I think the last fixes are around yosys error handling and adding a timing path through the T_INV block.

xc/common/primitives/ologice3/ologice3.pb_type.xml
mkurc-ant: No. When the route though the `T_INV` is chosen this inverter is active and that corresponds to not emitting the `ZINV_T1` feature. The feature disables the inversion. When the `OSERDES` is used then the inversion is controlled with its parameters.

xc/xc7/fasm2bels/ioi_models.py
mkurc-ant: There: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1589

xc/xc7/techmap/cells_map.v
mkurc-ant: Fixed.

xc/common/primitives/ologice3/ologice3.pb_type.xml
mkurc-ant: Fixed. Shouldn't matter anyways as the `T_INV` is only used when routing to const1

xc/xc7/techmap/cells_map.v
mkurc-ant: I disagree. What I'm doing is exactly what the documentation says in the second paragraph. See the example below. I've deliberately set parameters of a `OSERDES` to trigger the techmap failure. When there is `wire _TECHMAP_FAIL_ = 1` then the techmap fails silently and we end up with unmapped `OSERDESE2` in EBLIF:
```
   Number of cells:                  8
     BUFGCTRL_VPR                    2
     IBUF_VPR                        3
     OBUFT_VPR                       1
     OSERDESE2                       1
     T_INV                           1
```

But when I don't assign the wire with anything then Yosys fails and prints out the error cause which is exactly what I intended it to do:
```
ERROR: Techmap yielded config wire _TECHMAP_FAIL_ with non-const value \_TECHMAP_FAIL_.
```

Besides that I realzed that some of these error checks for OSERDES are incorrect. I'll file an issue about that.

xc/common/utils/prjxray_routing_import.py
```diff
@@ -441,7 +441,11 @@ def check_feature(feature):
         features.append(
             '{}.{}_ACTIVE'.format(feature_path[0], feature_path[-1])
         )
-        features.append('{}.{}_USED'.format(feature_path[0], feature_path[-1]))
+
+        if "FREQ_REF" not in feature_path[-2]:
+            features.append(
```
litghost, line 446: Can you add a comment of the pattern you are matching here?  I think it will be helpful

xc/common/primitives/ologice3/ologice3.pb_type.xml
litghost: Can you add what you wrote as an XML comment please?  I don't think this case is obvious to the reader.

xc/common/primitives/ologice3/ologice3.pb_type.xml
litghost: Sure, but we may end up using T_INV for signals if we teach yosys to emit it

xc/common/primitives/ologice3/ologice3.pb_type.xml
mkurc-ant: Mhm, added the comment.

xc/common/utils/prjxray_routing_import.py
mkurc-ant: Right, added a comment explaining it.

litghost:
CI failure is the result of a segfault.  I think there is a lingering semi-random memory corruption issue.  I've also seen `std::bad_alloc` throws that could be the same root cause, but I lack evidence.  Probably worth running this test with valgrind or ASan or MSan.

I'll file an issue.  Re-running the CI now.

litghost:
@mkurc-ant I re-ran just the xc7 job to see if it is intermittent.  I also filed https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1607 to track this issue.  I'm afraid that if it is a race condition it might be hard to catch via sanitizers.

xc/common/primitives/ologice3/ologice3.pb_type.xml
litghost: Thanks

litghost:
@mkurc-ant The IOB/IOI changes need to go to separate fasm2bels repo now that @acomodi has split it into a seperate repo, and symbiflow-arch-defs now uses that seperate repo.  Sorry for hurdle :/

xc/xc7/techmap/cells_map.v
litghost: I agree, resolving

mkurc-ant:
@litghost No problem, I new that at some point it would happen. I'll integrate that as soon as I have some time to do it.

mkurc-ant:
@litghost I've separated fasm2bels changes and made a PR to its repo https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/13. I expect the CI to fail now. Once fasm2bels changes are integrated I'll bump the submodule.

mkurc-ant:
I can see a CI failure for Murax on A200T. Seems not to be related to this PR, more to fasm2bels:
```
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/env/conda/envs/symbiflow_arch_def_base/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/env/conda/envs/symbiflow_arch_def_base/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/__main__.py", line 4, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/fasm2bels.py", line 416, in main
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/fasm2bels.py", line 103, in process_tile
    PROCESS_TILE[tile_type](top.conn, top, tile, tile_features)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/models/clb_models.py", line 1218, in process_clb
    process_slice(top, slices[s])
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/models/clb_models.py", line 528, in process_slice
    assert not site.has_feature('{}LUT.RAM'.format(row))
AssertionError
```

mkurc-ant:
@litghost I see a fasm2bels assertion related to LUT RAMs. I filed an issue about it with some details because I'm a little bit puzzled whether what can be observed is correct or not: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1625. Could you take a look?

PR #1504, opened 2020-05-26
install: remove dependency on prjxray env variables

acomodi:
This PR removes the need of exporting XRAY_FASM2FRAMES and XRAY_TOOLS_DIR variables to have the binary toolchain working correctly.

This PR relies on https://github.com/SymbiFlow/prjxray/pull/1344 to pass the Install CI


CMakeLists.txt
```diff
@@ -179,6 +183,20 @@ add_thirdparty_package(
   DEPENDS ${PYTHON3_TARGET} ${LIBICONV_TARGET}
 )
 
+add_thirdparty_package(
```
litghost, line 186: If we are packaging prjxray into conda, why is this required?

acomodi: This is for the python dependencies for the toolchain_wrappers

litghost: But if prjxray is installed from conda, why isn't the python library already instead?

acomodi: Currently the conda-package contains only the binary tools, and not the python library.
We could add python libraries as well in the conda package to avoid installing the third_party package here

litghost: I think the current naming scheme is confusing.  prjxray is made up of multiple parts.  The bitstream tools, the python library, the fuzzers and the database output.  If this package only has the tools, it should be probably be called prjxray-tools or xc7bittools, etc.

acomodi: Ok, than, it would be ok to leave this thirdparty package and modify the conda-package naming, correct?

acomodi:
> Isn't this missing a submodule update?

Indeed, fixed now, thanks. With this, the Install CI should turn green

xc/common/cmake/arch_define.cmake
```diff
@@ -135,22 +135,18 @@ function(ADD_XC_ARCH_DEFINE)
         --input /dev/stdin \
         --output /dev/stdout \
         \${PLACE_CONSTR_TOOL_EXTRA_ARGS}"
-    BITSTREAM_EXTENSION frames
-    BIN_EXTENSION bit
-    FASM_TO_BIT ${PRJRAY_DIR}/utils/fasm2frames.py
+    BITSTREAM_EXTENSION bit
+    FASM_TO_BIT ${symbiflow-arch-defs_SOURCE_DIR}/env/conda/envs/symbiflow_arch_def_base/bin/xcfasm
```
litghost, line 139: As a rule, if you are writing a path with "conda" in a CMake file, you are doing it wrong.

In this case, you want to add xcfasm to https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/CMakeLists.txt#L32 and use `${XCFASM}` here.

xc/common/cmake/arch_define.cmake
```diff
@@ -161,15 +157,15 @@ function(ADD_XC_ARCH_DEFINE)
...
+        --bit_file \${OUT_BITSTREAM} \
+        --fasm_file \${OUT_BITSTREAM}.fasm \
         --iostandard ${DEFAULT_IOSTANDARD} \
         --drive ${DEFAULT_DRIVE} \
         \${PCF_INPUT_IO_FILE} \
         --eblif \${OUT_EBLIF} \
         --top \${TOP} \
         \${OUT_BIT_VERILOG} \${OUT_BIT_VERILOG}.xdc"
-
+    NO_BIT_TO_BIN
```
litghost, line 168: This doesn't do what you want.  I believe all the targets that install bitstream's on targets will break.

xc/xc7/toolchain_wrappers/symbiflow_write_bitstream
```diff
@@ -7,19 +7,21 @@ MYPATH=`dirname ${MYPATH}`
 source ${MYPATH}/env
 echo "Writing bitstream ..."
 
-if [ -z ${XRAY_FASM2FRAMES} ]; then
-	echo "ERROR: XRAY_FASM2FRAMES variable is unset."
+DATABASE_DIR=""
+if [ ! -z ${XRAY_DATABASE_DIR} ]; then
+	DATABASE_DIR=${XRAY_DATABASE_DIR}
+elif [ ! -z ${CONDA_PREFIX} ]; then
```
litghost, line 13: This is wrong

xc/xc7/toolchain_wrappers/symbiflow_write_bitstream
```diff
@@ -81,7 +83,5 @@ fi
 FRAMES="${BIT%.*}.frames"
```
litghost, line 83: You have some old variables lying around

xc/common/cmake/arch_define.cmake
acomodi: Actually, what I did here is to have only one step for the bitstream generation. Previously, as there were two separate steps (fasm2frames and xc7frames2bit, the FASM_TO_BIT and BIT_TO_BIN were required.

Given that with xcfasm we can have only one step, the BIT_TO_BIN is no longer required. All the xc7 `<test>_bin` targets have been deleted and substituted with the `<test>_bit`.

litghost: I understand what you are saying, but https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/cmake/devices.cmake#L1658 will break the way it is now.

acomodi: Ok, now I get it. I guess this is uncaught by CI. One way to fix it would be to modify devices.cmake to set the correct dependency. E.g.: if there is no BIT to BIN the PROG target takes the OUT_BITSTREAM as dependency

litghost: I wonder if we just need a BIT_IS_BIN, and passthrough the right variables.  It'll remove some of the more random changes here.

acomodi: I am uncertain of what the random changes are. IMO it would be better to have the OUT_BIN set to default to point at the OUT_BITSTREAM, and, if the BIT_TO_BIN step is enabled, than the OUT_BIN will get overridden.

xc/xc7/toolchain_wrappers/symbiflow_write_bitstream
acomodi: You mean relying on CONDA_PREFIX? This is one of the ways we can avoid the user to set the DATABASE path when using the toolchain, unless we get to integrate the db within the tarball.

Another solution would be to get a conda package that installs the toolchain wrappers as well as the db. This way we can set a default path to the DB that can get overriden by the XRAY_DATABASE_DIR var

litghost: We should probably provide something akin to "yosys-config" on the path, and then you could do something like:
```
DATABASE_DIR?=$(prjxray-config --database_dir)
```

Rather than depending on conda.  Then it will work under any packaging scheme, as long as `prjxray-config` is on the path.

acomodi: Right, this seems a very clean solution, so we can make the wrappers agnostic to the packaging system.

I have opened a PR in conda-packages that provides a script that has the only purpose to print out the database location: https://github.com/SymbiFlow/conda-packages/pull/121

HackerFoo: Configuration files are easier to work with (e.g. in `/usr/share/prjxray`) than a `yosys-config` style binary, but I can work with it if it's a script. Environment variables work well too.

I've gone as far as patching `yosys` with `bbe` to change its hardcoded data directory to avoid recompilation just to locate plugins, but I don't like that.

HackerFoo: The idea is that each set of artifacts generated from an independent source should be relocatable. The fact that Yosys and its plugins can be in separate repos causes trouble.

xc/xc7/toolchain_wrappers/symbiflow_write_bitstream
acomodi: Done

xc/common/cmake/arch_define.cmake
acomodi: Fixed

environment.yml
```diff
@@ -15,6 +15,7 @@ dependencies:
   - iverilog
   - libxml2=2.9.10=20200629_180127
   - icestorm
+  - prjxray-db
```
litghost, line 18: This should be pinned, not floating

xc/common/cmake/arch_define.cmake
```diff
@@ -90,6 +90,8 @@ function(ADD_XC_ARCH_DEFINE)
       add_file_target(FILE ${VPR_CELLS_SIM} ABSOLUTE)
   endif ()
 
+  get_target_property_required(XCFASM env XCFASM)
```
litghost, line 93: Why use the xc-fasm submodule versus the package?  I'd argue that we should use the xc-fasm package, but the prjxray-db submodule, but *shurg*?

environment.yml
litghost: I'm not sure we should be using the package here?  At least not in this PR?  Maybe continue to use the submodule, and make a future PR?

xc/xc7/tests/install_test/CMakeLists.txt
```diff
@@ -5,8 +5,6 @@ add_test(NAME binary_toolchain_test
 	PATH=${INSTALLATION_DIR_BIN}:$ENV{PATH}
 	${CMAKE_COMMAND} -E env
 	PYTHONPATH=${PRJXRAY_DIR}:${PRJXRAY_DIR}/third_party/fasm
-	XRAY_FASM2FRAMES=${PRJXRAY_DIR}/utils/fasm2frames.py
-	XRAY_TOOLS_DIR=${symbiflow-arch-defs_SOURCE_DIR}/build/third_party/prjxray/tools
-	XRAY_DATABASE_DIR=${PRJXRAY_DB_DIR}
+	FRAMES2BIT=$<TARGET_FILE:xc7frames2bit>
```
litghost, line 8: Should probably still be setting `DATABASE_DIR` given that `PRJXRAY_DB_DIR` exists?

xc/xc7/toolchain_wrappers/symbiflow_synth
```diff
@@ -92,13 +87,15 @@ if [ ${#VERILOG_FILES[@]} -eq 0 ]; then
 	exit 1
 fi
 
+DATABASE_DIR=`prjxray-config`
```
litghost, line 90: This should be:
```
: $(DATABASE_DIR:=$(prjxray-config)
```
So that overriding the DATABASE_DIR is possible?

xc/xc7/toolchain_wrappers/symbiflow_write_bitstream
```diff
@@ -57,9 +47,12 @@ while true; do
 	esac
 done
 
+DATABASE_DIR=`prjxray-config`
```
litghost, line 50: Same as above

xc/common/cmake/arch_define.cmake
acomodi: What you mean by xc-fasm package? Currently it is installed via pip in the requirements.txt, I have followed the same ways as the other python packages we are using (e.g. fasm, python-sdf-timing).

xc/xc7/tests/install_test/CMakeLists.txt
acomodi: Given that this tests the toolchain it would probably be better to leave the DATABASE_DIR unset, so that we are sure that the prjxray-db package is working properly

litghost: I understand what you are saying, but we have an inconsistency then.  We definitely should not have two copies of the prjxray-db during the CI, and I think it is more important to respect the `PRJXRAY_DB_DIR` CMake cache variable.  Still thinking.

xc/common/cmake/arch_define.cmake
litghost: Sorry, I agree.  I was thinking xcfasm would be coming from pip, and it is as a "+ third_party/..." package.

xc/xc7/toolchain_wrappers/symbiflow_synth
acomodi: Yes. makes sense, done

xc/xc7/toolchain_wrappers/symbiflow_write_bitstream
acomodi: Done

environment.yml
acomodi: As commented earlier, IMO it would be better to test this as well at least for the install_test that checks whether everything is all right with the toolchain and the toolchain wrappers.

xc/xc7/tests/install_test/CMakeLists.txt
litghost: I believe we should respect `PRJXRAY_DB_DIR ` in this case.  The cache variable can default to the result of `prjxray-config`, but if the use specifies another path, it should be used here.

You also need to remove the prjxray-db submodule.

acomodi: All right, I can than remove the dependency and override the DATABASE_DIR. Than we'll update that in a new PR in the future

Issue #1338, opened 2020-02-25
Fix inference of IBUF/OBUF in SymbiFlow


mkurc-ant:
We have two types of tests: ones that require a ROI and others that don't. For the tests with ROI there must be no I/O buffer inference in Yosys as it lead to unplacable situation.

What needs to be done is to pass a parameter to the TCL script ran in Yosys that would indicate whether a ROI is used. Based on that parameter the switch `-iopad` should be invoked or not.

litghost:
Fixed in #1349

PR #1506, opened 2020-05-27
Zynq 7020 support

andrewb1999:
Adds support for xc7z020 and some 7020 boards.  Full bitstream uploading for the pynq z1 works, other boards should only need an openocd config.  Relies on an updated database and prjxray submodule from [#1338](https://github.com/SymbiFlow/prjxray/pull/1338)


xc/xc7/archs/zynq7_z020/tiles/clbll_l/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+project_ray_tile(
+  ARCH zynq7_z020
+  TILE CLBLL_L
+  SITE_TYPES SLICEL/SLICEL0 SLICEL/SLICEL1
+  )
```
acomodi, line 5: I believe there was an error in the original PR I made for the zynq boards addition. There was a change in the tiles used here: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1484. This basically avoids using SLICEs as tiles, and reintroduces CLBs.

All the CLB tiles would require to be in the same format as the other archs:
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/5588c2891f4d773cbbeb85843626fd4972bebeee/xc/xc7/archs/zynq7/tiles/clbll_l/CMakeLists.txt#L1-L5

xc/xc7/archs/zynq7_z020/tiles/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+add_subdirectory(bram_l)
+add_subdirectory(bram_r)
+add_subdirectory(slicel)
+add_subdirectory(slicem)
```
acomodi, line 4: Also the CLB tiles should be added

xc/xc7/archs/zynq7_z020/devices/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Zynq 7020 contains tiles that are not present in 7010. Since we cannot have
+# different tile sets per device, this one is disabled for now.
```
acomodi, line 2: I think this comment can be removed

xc/xc7/archs/zynq7_z020/tiles/clbll_l/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 project_ray_tile(
```
acomodi, line 1: The cmake function should be changed as follows (for all CLBs)

```suggestion
project_ray_tile_capacity(
```

xc/xc7/archs/zynq7/tiles/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ set(IOPAD_M_SITES IOB33M IDELAYE2 ILOGICE3 OLOGICE3)
 
 project_ray_equiv_tile(
   ARCH zynq7
-  TILES RIOPAD_M RIOPAD_S RIOPAD_SING
+  TILES RIOPAD_M RIOPAD_S RIOPAD_SING LIOPAD_M LIOPAD_S LIOPAD_SING
```
acomodi, line 21: I think that there was another mistake during my rebase of the `zynq-board` branch.

LIOPADs are not present in the zynq010 device

xc/xc7/archs/zynq7/tiles/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ set(PLLE2_ADV_SITES PLLE2_ADV)
 
 project_ray_equiv_tile(
   ARCH zynq7
-  TILES CMT_TOP_L_UPPER_T
+  TILES CMT_TOP_L_UPPER_T CMT_TOP_R_UPPER_T
```
acomodi, line 31: Same goes for this change, which should be reverted

litghost:
The install failure occurred because you need the new prjxray-db, but after you update the submodule  it might go green!

andrewb1999:
Rebased on master with updated prjxray-db.  CI should go green now.

mithro:
@litghost - I think we can merge this, right?

xc/xc7/boards.cmake
```diff
@@ -95,3 +88,32 @@ add_xc_board(
...
+  PART xc7z020clg400-1
+  PROG_TOOL ${OPENOCD_TARGET}
+  PROG_CMD "${OPENOCD} -f ${PRJXRAY_DIR}/utils/openocd/board-digilent-pynqz1.cfg -c \\\"init $<SEMICOLON> pld load 0 \${OUT_BIN} $<SEMICOLON> exit\\\""
+)
+
+add_xc_board(
+  BOARD marszx3-full
+  DEVICE xc7z020
+  PACKAGE test
+  PART xc7z020clg484-1
```
litghost, line 118: 3 of these 4 boards have no PROG_CMD.  Is there not a supported OpenOCD config for these boards?  @mithro Should we have a place to track missing OpenOCD configuration support?

litghost:
How many of the target boards were you able to test on?

andrewb1999:
> How many of the target boards were you able to test on?

I was only able to test on the pynq z1 board, but I think the only different between boards should be the package, the openOCD config, and the pcf. @acomodi have you been able to test on any other boards?

xc/xc7/boards.cmake
andrewb1999: Not sure if there are openocd configs for the other boards already, but if there are I can't test if they work as I only have a pynq z1.

litghost: In general it is probably a good idea to leave off the untested configurations, or comment them out.  Making a small follow on PR with board support once it gets tested on other targets it preferable to checking in untested configurations.

We do have some testing via the vendor tool CI, but ideally we have both the "works on a board" and a green result from the CI.

acomodi:
@andrewb1999 IIRC I have tested the zedboard, I think also @tmichalak tested some of the other boards, but this happened several months ago.

I do not have the hardware at the moment, probably it is better to comment out as @litghost said the non-tested boards, and, once we get proof that tests pass on HW for the others, we can safely de-comment them.

litghost:
> @andrewb1999 IIRC I have tested the zedboard, I think also @tmichalak tested some of the other boards, but this happened several months ago.
> 
> I do not have the hardware at the moment, probably it is better to comment out as @litghost said the non-tested boards, and, once we get proof that tests pass on HW for the others, we can safely de-comment them.

I suggest we comment out the configurations without PROG_CMD's, and and submit.  Then @acomodi and/or @tmichalak can make a PR uncommenting the relevant boards and add their PROG_CMD's once they have been tested.

andrewb1999:
@litghost Commented out non tested boards and made changes to top level CMakeLists.txt. Assume we need to see what CI does now.

litghost:
> Assume we need to see what CI does now.

Yes

andrewb1999:
CI looks good to me.  Any other changes?

litghost:
```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/counter/counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test && /usr/bin/cmake -E remove -f counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test/design_counter_pynqz1_vivado.dcp && /usr/bin/cmake -E remove -f counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test/design_counter_pynqz1_vivado.xpr && /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/prjxray/utils/vivado.sh -mode batch -source /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/counter/counter_pynqz1_vivado_runme.tcl > /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/counter/counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test/vivado.stdout.log
ERROR: [DRC RTSTAT-2] Partially routed nets: 1 net(s) are partially routed. The problem bus(es) and/or net(s) are RIOB33_X73Y123_IOB_X1Y124_I_BUFG.
ERROR: [DRC RTSTAT-5] Partial antennas: 1 net(s) have a partial antenna. The problem bus(es) and/or net(s) are RIOB33_X73Y123_IOB_X1Y124_I.
ERROR: [DRC RTSTAT-6] Partial route conflicts: 2 net(s) have a partial conflict. The problem bus(es) and/or net(s) are RIOB33_X73Y123_IOB_X1Y124_I, and RIOB33_X73Y123_IOB_X1Y124_I_BUFG.
```

Looks like something is going wrong in the routing near the IOB in the zynq counter test when going back to Vivado.  Can you take a look?

andrewb1999:
> ```
> FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/counter/counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test && /usr/bin/cmake -E remove -f counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test/design_counter_pynqz1_vivado.dcp && /usr/bin/cmake -E remove -f counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test/design_counter_pynqz1_vivado.xpr && /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/prjxray/utils/vivado.sh -mode batch -source /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/counter/counter_pynqz1_vivado_runme.tcl > /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/counter/counter_pynqz1/zynq7_z020-xc7z020-virt-xc7z020-test/vivado.stdout.log
> ERROR: [DRC RTSTAT-2] Partially routed nets: 1 net(s) are partially routed. The problem bus(es) and/or net(s) are RIOB33_X73Y123_IOB_X1Y124_I_BUFG.
> ERROR: [DRC RTSTAT-5] Partial antennas: 1 net(s) have a partial antenna. The problem bus(es) and/or net(s) are RIOB33_X73Y123_IOB_X1Y124_I.
> ERROR: [DRC RTSTAT-6] Partial route conflicts: 2 net(s) have a partial conflict. The problem bus(es) and/or net(s) are RIOB33_X73Y123_IOB_X1Y124_I, and RIOB33_X73Y123_IOB_X1Y124_I_BUFG.
> ERROR: [Vivado 12-1345] Error(s) found during DRC. Bitgen not run.
> ```
> 
> Looks like something is going wrong in the routing near the IOB in the zynq counter test when going back to Vivado. Can you take a look?

Not sure why this is happening.  The counters test works on the physical pynq board.  How would you suggest trying to debug this?  Or should I just remove the counters test for now and create a pull request later to fix it?

Also, is this in anyway related to the zybo tests being commented out in CMakeLists.txt?

litghost:
> How would you suggest trying to debug this?

Open the dcp generated by building the relevant target and determine where the antenna nets are coming from.

> Or should I just remove the counters test for now and create a pull request later to fix it?

We should determine the problem now, as I expect that all future Z7020 tests will hit a similiar issue.

> Also, is this in anyway related to the zybo tests being commented out in CMakeLists.txt?

Doubt it.  This failure is on the pynqz1 board.

andrewb1999:
@litghost Looks like the issue is conflicting site pins between the clock net before the BUFG and the clock net after the BUFG.  Unsure why both of these would be routed to the slice clk pins.  I assume only the BUFG one should route to the clk pins?

litghost:
> @litghost Looks like the issue is conflicting site pins between the clock net before the BUFG and the clock net after the BUFG. Unsure why both of these would be routed to the slice clk pins. I assume only the BUFG one should route to the clk pins?

Sorry, your description of the problem doesn't make sense.  Can you attach an image or attach the DCP?

andrewb1999:
> Sorry, your description of the problem doesn't make sense to me. Can you attach an image or attach the DCP?

[dcp](https://github.com/andrewb1999/summer-of-code-resources/blob/master/design_counter_pynqz1_vivado_pre_route.dcp)

RIOB33_X73Y123_IOB_X1Y124_I highlighted:
image

RIOB33_X73Y123_IOB_X1Y124_I_BUFG highlighted:
image

xc/xc7/tests/counter/counter_pynqz1.v
```diff
@@ -0,0 +1,24 @@
...
+    output wire led5_g,
+    output wire led5_r
+);
+
+	localparam BITS = 4;
+	localparam LOG2DELAY = 18;
+
+	reg [BITS+LOG2DELAY-1:0] counter = 0;
+
+	always @(posedge clk) begin
```
litghost, line 19: You need an explicit BUFG between the clk pin and using it with logic.

litghost:
The problem has to do with a missing BUFG in the design.  Currently BUFG inference is intentionally disabled, which does mean that all top level clock ports need to explicitly go through a BUFG primative prior to faning out.

This is a known issue with the current tool chain configuration, but is intentional, as we don't want to debug yosys BUFG inference issues at the same time as dealing with place and route tool bringup.

andrewb1999:
Ok those changes seem to have fixed CI.  Good to go now?

mithro:
@andrewb1999 Congratulations!

PR #1511, opened 2020-05-28
Bump VPR + delayless switch fix

mkurc-ant:
This PR bumps VPR to the recent master+wip with post synthesis netlist write fix (analysis targets work). Also suppresses emission of `__vpr_delayless_switch__` to the arch.xml. It is now implicitly defined.


mkurc-ant:
@tcal-x Unfortunately yes. This is related to this issue https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1400

PR #1521, opened 2020-06-08
Fix USE_CONDA=FALSE

HackerFoo:
Allows an outside package manager to take over dependency management.

There may be better solutions, but the plan is to take package management (Conda) out of CMake later, so there's no point in perfecting this.

Related: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/564


common/cmake/util.cmake
```diff
@@ -8,11 +8,5 @@ function(get_target_property_required var target property)
         "${property} not set for target ${target}, check target definition."
     )
   endif()
-  if("${PROP}" STREQUAL "")
```
litghost, line 11: This needs to stay put

xc/xc7/tests/install_test/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+if(${USE_CONDA})
```
litghost, line 1: This needs to be fixed, rather than disabled.  I expect that the use of CONDA_DIR should be removed, and `conda activate` invoked instead.

ice40/icestorm.cmake
```diff
@@ -5,7 +5,7 @@ function(icestorm_setup)
   set(ICESTORM_SRC ${symbiflow-arch-defs_SOURCE_DIR}/third_party/icestorm CACHE PATH "Path to icestorm repository")
   set(PYUTILS_PATH ${symbiflow-arch-defs_SOURCE_DIR}/utils:${symbiflow-arch-defs_SOURCE_DIR}/ice40/utils/fasm_icebox)
 
-  get_target_property_required(PYTHON_PREFIX env CONDA_DIR)
+  get_target_property(PYTHON_PREFIX env CONDA_DIR)
```
litghost, line 8: This needs to be fixed, just changing it to get_target_property will break things.

common/cmake/util.cmake
HackerFoo: Fixed.

xc/xc7/tests/install_test/CMakeLists.txt
HackerFoo: Fixed.

ice40/icestorm.cmake
HackerFoo: Fixed.

ice40/icestorm.cmake
```diff
@@ -5,8 +5,12 @@ function(icestorm_setup)
   set(ICESTORM_SRC ${symbiflow-arch-defs_SOURCE_DIR}/third_party/icestorm CACHE PATH "Path to icestorm repository")
   set(PYUTILS_PATH ${symbiflow-arch-defs_SOURCE_DIR}/utils:${symbiflow-arch-defs_SOURCE_DIR}/ice40/utils/fasm_icebox)
 
-  get_target_property_required(PYTHON_PREFIX env CONDA_DIR)
-  set(ICESTORM_PREFIX "PREFIX=${PYTHON_PREFIX}")
+  if(${USE_CONDA})
+    get_target_property_required(PYTHON_PREFIX env CONDA_DIR)
+    set(ICESTORM_PREFIX "PREFIX=${PYTHON_PREFIX}")
+  else()
+    set(ICESTORM_PREFIX "")
```
litghost, line 12: I wonder if this should be a CACHE var when running outside of conda?

HackerFoo: Done.

Issue #1517, opened 2020-06-02
Place constraints script ignores ROI


rw1nkler:
The `prjxray_create_place_constraints.py` script does not use the ROI parameters. It tries to place the PLL instances from the BUFGMUX test, outside the region of interest:

```Traceback (most recent call last):
  File "/home/build/rwinkler/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 685, in <module>
    main()
  File "/home/build/rwinkler/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 673, in main
    vpr_grid, loc_in_use, block_locs, blocks, grid_capacities):
  File "/home/build/rwinkler/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 508, in place_clocks
    vpr_loc = vpr_locs[potential_loc]
KeyError: 'PLLE2_ADV_X1Y0'
```

The mentioned PLL is outside the region of interest.  The solution is to restrict the script to not use tiles outside the ROI.

rw1nkler:
I was trying to lock all BUFG and PLL for this example, however, the script fails on this attempt:
```
Traceback (most recent call last):
  File "/home/build/rwinkler/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 687, in <module>
    main()
  File "/home/build/rwinkler/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 675, in main
    vpr_grid, loc_in_use, block_locs, blocks, grid_capacities):
  File "/home/build/rwinkler/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 445, in place_clocks
    self.assign_cmts(vpr_grid, block_locs)
  File "/home/build/rwinkler/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 437, in assign_cmts
    solutions = problem.getSolutions()
  File "/home/build/rwinkler/symbiflow-arch-defs/build/env/conda/lib/python3.7/site-packages/constraint/__init__.py", line 271, in getSolutions
    return self._solver.getSolutions(domains, constraints, vconstraints)
  File "/home/build/rwinkler/symbiflow-arch-defs/build/env/conda/lib/python3.7/site-packages/constraint/__init__.py", line 567, in getSolutions
[...]

PR #1523, opened 2020-06-09
Fix clock placement in devices with limited grid

rw1nkler:
This commit adds the ability to verify if the clock location is available in the FPGA grid before placement. It uses architecture XML for the verification.

Resolves https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1517

### DIFF
--- a/xc/common/utils/prjxray_create_place_constraints.py
+++ b/xc/common/utils/prjxray_create_place_constraints.py
@@ -119,11 +119,15 @@ class VprGrid(object):
     """This class contains a set of dictionaries helpful
     to have a fast lookup at the various coordinates mapping."""
 
-    def __init__(self, vpr_grid_map):
+    def __init__(self, vpr_grid_map, graph_limit):
         self.site_dict = dict()
         self.site_type_dict = dict()
         self.cmt_dict = dict()
         self.tile_dict = dict()
+
+        if graph_limit is not None:
+            limits = graph_limit.split(",")
+            xmin, ymin, xmax, ymax = [int(x) for x in limits]
 
         with open(vpr_grid_map, 'r') as csv_vpr_grid:
             csv_reader = csv.DictReader(csv_vpr_grid)
@@ -136,6 +140,12 @@ class VprGrid(object):
                 can_x = row['canon_x']
                 can_y = row['canon_y']
                 clk_region = row['clock_region']
+
+                if graph_limit is not None:
+                    if int(can_x) < xmin or int(can_x) > xmax:
+                        continue
+                    if int(can_y) < ymin or int(can_y) > ymax:
+                        continue
 
                 clk_region = None if not clk_region else int(clk_region)
 
@@ -199,6 +209,7 @@ class ClockPlacer(object):
             io_locs,
             blif_data,
             roi,
+            graph_limit,
             allow_bufg_logic_sources=False
     ):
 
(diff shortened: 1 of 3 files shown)


mithro:
PTAL - @tcal-x

tcal-x:
> PTAL - @tcal-x

I will try the script on the test where I had to add the workaround LOC constraints.   I need to rebuild rr_graph in my sandbox though so it will be a few hours before I have an update.

tcal-x:
This fix may be unrelated to the issue I'm seeing.  In any case, I see exactly the same error using the new version of the script, just the line numbers are different.   In my case, the script has difficulty placing the PLLE2_ADV if it is not constrained.  I'm targeting the full 100T chip.

This is the error (line numbers are using the new version from this PR):
```
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 734, in <module>
    main()
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 722, in main
    vpr_grid, loc_in_use, block_locs, blocks, grid_capacities):
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 519, in place_clocks
    for potential_loc in sorted(available_placements[key]):
KeyError: ('PLLE2_ADV', 5)
```

rw1nkler:
I have changed the script to use limited device grid. Additionally, now it is possible to use only top/bottom half of the device. 

The script does not check if the used resource is available in the given clock domain. (This bug was present before) Because of that, after limiting the device grid, sometimes it tries to place i.e. `PLLE2_ADV` instance in the clock domain that, does not have it. I fixed the problem by excluding almost empty clock domains from the `basys-bottom` device definition when the problem occurs.

A universal solution requires rearranging the constraint problem and adding proper checks of variable domains to clocking resources like i.e. PLLs.

xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -200,15 +210,21 @@ def __init__(self, vpr_grid, io_locs, blif_data, roi):
...
 
-        _, thresh_top_y = cmt_dict[top_cmt_tile]['canon_loc']
-        _, thresh_bot_y = cmt_dict[bot_cmt_tile]['canon_loc']
+        try:
+            bot_cmt_tile = next(
+                k for k, v in cmt_dict.items() if k.startswith('CLK_BUFG_BOT')
+            )
+            _, thresh_bot_y = cmt_dict[bot_cmt_tile]['canon_loc']
+        except StopIteration:
+            thresh_bot_y = None
```
litghost, line 228: Should likely assert that either thresh_bot_y or thresh_top_y is present.  Otherwise things will likely break horribly.

litghost: Also if graph limit is not set, both thresh_bot_y and thresh_top_y should be `not None`.

rw1nkler: I have added the requested changes

xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -189,26 +199,39 @@ def get_tile_dict(self):
...
+
+        if args.graph_limit is None:
+            assert thresh_top_y is not None, "CMTs in the top half of the device not found"
+            assert thresh_bot_y is not None, "CMTs in the bottom half of the device not found"
+        else:
+            assert thresh_top_y is not None or thresh_bot_y is not None, (
+                "The device grid does not contain any CMTs"
```
litghost, line 233: A better message would be "does not contain any BUFG sites".  If now BUFG's are available, then any source to BUFG connections become effectively impossible, without something special.

xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -189,26 +199,39 @@ def get_tile_dict(self):
 
 
 class ClockPlacer(object):
-    def __init__(self, vpr_grid, io_locs, blif_data, roi):
+    def __init__(self, vpr_grid, io_locs, blif_data, args):
```
litghost, line 202: Rather than pass args directly, please pass roi and clock limit explicitly.  We want to avoid tying ClockPlacer directly to argparse.

xc/common/utils/prjxray_create_place_constraints.py
rw1nkler: Should I also change the previous assert messages to "BUFG sites in the top/bottom half of the device not found"?

litghost: Ya, that is more precise.

xc/common/utils/prjxray_create_place_constraints.py
rw1nkler: Fixed

PR #1526, opened 2020-06-15
fasm2bels: add --verbose to catch unknown bits

acomodi:

### DIFF
--- a/xc/common/cmake/vivado.cmake
+++ b/xc/common/cmake/vivado.cmake
@@ -112,6 +112,7 @@ function(COMMON_VIVADO_TARGETS)
           --part ${PART}
           --db-root ${PRJRAY_DB_DIR}/${PRJRAY_ARCH}
           --bitread $<TARGET_FILE:bitread>
+          --verbose
           ${CMAKE_CURRENT_BINARY_DIR}/${WORK_DIR}/design_${NAME}.bit
           > ${CMAKE_CURRENT_BINARY_DIR}/${WORK_DIR}/design_${NAME}.bit.fasm
       WORKING_DIRECTORY ${WORK_DIR}


acomodi:
@litghost Vendor tools CI shows several tests failing for unknown bits. We should track them down and analyze locally what is happening.

litghost:
> @litghost Vendor tools CI shows several tests failing for unknown bits. We should track them down and analyze locally what is happening.

Yes.  They all appear to be SERDES related bits.   I vaguely feel that @mkurc-ant might know about this?

mkurc-ant:
I doubt that I'll have time to look into that this week.

For sure there are unknown IOB bits that emerge whenever internal VREF is not set and there is something that uses an IOSTANDARD requiring VREF. Meaning that those bits enable external VREF.

As for SERDES I'd need to revisit the fuzzer that I wrote for the ISERDES some time ago. I dropped its integration as it wasn't a priority.

litghost:
FYI, fixing this should be part of the "getting ready for users" burndown.

mkurc-ant:
I've created an issue with the list of the unknown bits and added it to the "getting ready for users" burndown list.

mkurc-ant:
This is strange, When reproducing the problem these bits emerged even when converting a Symbiflow bitstream to fasm. But once I've re-tried that on the most recent Symbiflow master the problem was gone.

I suspect that the cause was some kind of bug in bittools that has already been fixed. I rebased the PR. Let's see how the CI goes.

mkurc-ant:
I must have messed up something. Now I can reproduce the issue locally.

mkurc-ant:
I've finished integration of the ISERDES fuzzer there: https://github.com/SymbiFlow/prjxray/pull/978 According to my local tests that allowed to get rid of at least some unknown bits.

Issue #1538, opened 2020-06-22
Multiple mapped tiles when creating synth io tiles


andrewb1999:
Added a synth tile definition to the design.json as follows:
`        {
            "name": "dout[8]",
            "node": "INT_R_X23Y100/LH3",
            "pin": "H7",
            "wire": "INT_INTERFACE_R_X23Y100/INT_INTERFACE_LH3",
            "wires_outside_roi": [
            ]
        }
`

And building the synth tiles gives the following error:
`  File "/scratch/safe/butta/symbiflow-arch-defs.syn/xc/common/utils/prjxray_create_synth_tiles.py", line 26, in map_tile_to_vpr_coord
    assert len(mapped_tiles) == 1, tile
AssertionError: INT_INTERFACE_R_X23Y100
`
I know this is not a normal location used as a synth tile, but is this a location I should be able to place a synth tile?  Is mapping to multiple vpr coordinates the correct result in this situation or a bug?

Link to branch containing these changes: [synth-tile-errors](https://github.com/andrewb1999/symbiflow-arch-defs/tree/synth-tile-errors)

andrewb1999:
This issue seems to only occurs when trying to place synth tiles at a location that borders a clock region boundary.  Do these tile names get mapped both to the actual tile coord and the break tile coord?

PR #1540, opened 2020-06-23
Fix synth tiles map to vpr coord

andrewb1999:

If synth tile is being placed at a tile that maps to two vpr coordinates, automatically choose the non-empty tile.  Solves: #1538

### DIFF
--- a/xc/common/utils/prjxray_create_synth_tiles.py
+++ b/xc/common/utils/prjxray_create_synth_tiles.py
@@ -23,8 +23,24 @@ def map_tile_to_vpr_coord(conn, tile):
         (phy_tile_pkey, )
     )
     mapped_tiles = c.fetchall()
-    assert len(mapped_tiles) == 1, tile
-    tile_pkey = mapped_tiles[0][0]
+    assert len(mapped_tiles) > 0, tile
+    if len(mapped_tiles) > 1:
+        c.execute(
+            """
+        SELECT COUNT(pkey) FROM wire WHERE tile_pkey = ?""", mapped_tiles[0]
+        )
+        num_wires_first_tile, = c.fetchall()
+        c.execute(
+            """
+        SELECT COUNT(pkey) FROM wire WHERE tile_pkey = ?""", mapped_tiles[1]
+        )
+        num_wires_second_tile, = c.fetchall()
+        if num_wires_first_tile[0] > 0:
+            tile_pkey = mapped_tiles[0][0]
+        else:
+            tile_pkey = mapped_tiles[1][0]
+    else:
+        tile_pkey = mapped_tiles[0][0]
 
     c.execute("SELECT grid_x, grid_y FROM tile WHERE pkey = ?", (tile_pkey, ))
     grid_x, grid_y = c.fetchone()


xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -23,8 +23,24 @@ def map_tile_to_vpr_coord(conn, tile):
         (phy_tile_pkey, )
     )
     mapped_tiles = c.fetchall()
-    assert len(mapped_tiles) == 1, tile
-    tile_pkey = mapped_tiles[0][0]
```
litghost, line 27: Docstring is out of date

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -23,8 +23,24 @@ def map_tile_to_vpr_coord(conn, tile):
...
+    assert len(mapped_tiles) > 0, tile
+    if len(mapped_tiles) > 1:
+        c.execute(
```
litghost, line 28: This code appears to be assuming that len(mapped_tiles) in [1,2]?

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -23,8 +23,24 @@ def map_tile_to_vpr_coord(conn, tile):
...
+            """
+        SELECT COUNT(pkey) FROM wire WHERE tile_pkey = ?""", mapped_tiles[0]
+        )
+        num_wires_first_tile, = c.fetchall()
+        c.execute(
+            """
+        SELECT COUNT(pkey) FROM wire WHERE tile_pkey = ?""", mapped_tiles[1]
+        )
+        num_wires_second_tile, = c.fetchall()
```
litghost, line 37: In words, what is the idea behind what this code is doing?

xc/common/utils/prjxray_create_synth_tiles.py
andrewb1999: Yes, I should fix that assertion

xc/common/utils/prjxray_create_synth_tiles.py
andrewb1999: From my understanding and testing, phy tile names that are right above a horizontal clock boundary map both to the actual tile with that name and the BRKH tile below.  For this reason, doing a database search for one of these tiles that borders a clock region boundary finds two vpr coordinates, causing the assertion that len(mapped_tiles) == 1 to fail.  I want to be able to place synth tiles at these locations that border a clock region boundary, so I want to pick the correct of the two vpr coordinate locations (not the BRKH tile).  BRKH tiles have no wires in the database, so we can determine the correct vpr location to place a synth tile at by finding the number of wires at each vpr coordinate and choosing the one with a non-zero number of wires.

litghost: Can you give some example SQL queries that demostrate what you are talking about?  Your description of the symptom sounds weird to be, so I'd like some illustrative examples of what you mean.

andrewb1999: `c.execute( "SELECT tile_pkey FROM tile_map WHERE phy_tile_pkey = (SELECT pkey FROM phy_tile WHERE name = ?)", ('INT_R_X23Y100', ) )
`

Returns `[(9332,), (9333,)]`

and

`c.execute( "SELECT tile_pkey FROM tile_map WHERE phy_tile_pkey = (SELECT pkey FROM phy_tile WHERE name = ?)", ('INT_L_X18Y100', ) )`

Returns `[(7892,), (7893,)]`
etc.  I do not see any reason why these names should map to multiple tile_pkeys.

litghost: Which arch is this?

andrewb1999: xc7a50t-arty-swbut-roi-virt

litghost: > I do not see any reason why these names should map to multiple tile_pkeys.

I agree with you.  I believe that might be the better issue to fix :/

andrewb1999: Ok.  So this issue is in connections.db generation right?  Any suggestions on where to look to fix this issue?

litghost: > Ok. So this issue is in connections.db generation right? Any suggestions on where to look to fix this issue?

I would first determine if something weird is happening around https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc/common/utils/prjxray_form_channels.py#L1691

litghost: The map is defined here: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc/common/utils/prjxray_form_channels.py#L1981

So tracing back why the INT tile got splitted the way it was should original there.

andrewb1999: In the vpr_grid_map.csv generated by prjxray_form_channels.py, INT_R_X23Y100 maps to only a single vpr grid location.  Is vpr_grid_map.csv a byproduct or is it used to generated a database later on?

litghost: By product

andrewb1999: What do you think is the correct behavior in this situation?  59,53 maps to INT_R_X23Y100 and 59,52 maps to a BRKH phy_tile_pkey?  Are BRKH tiles supposed to be in the vpr grid?

Based on an initial look, the two grid locations are mapped to the same phy_tile_pkey much earlier on (traced it back to here https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc/common/utils/prjxray_form_channels.py#L1847 so far).

litghost: Only tiles that get split should map to multiple phy_tiles. Right now we only split IO tiles, so it is suprising that INT tiles are getting split.

andrewb1999: It's not one tile mapping to multiple phy_tiles, it's two tiles mapping to the same phy_tile.  For both of the tiles split_tiles is False as well.

litghost: Still weird.  I believe unsplit / unmerged tiles should have a 1:1 tile <-> phy_tile relationship.

andrewb1999: Ok.  I'm not quite sure where to go from here when debugging.

andrewb1999: Ok looks like I was slightly wrong in my diagnosis of the issue.  In the vpr grid there is a NULL tile at coordinate (59,52) which maps to the phy_tile INT_R_X23Y100, an INT_R tile at (59,53) which maps to the phy_tile INT_R_X23Y100, and a BRKH_INT tile as (59,54) which maps to phy_tile BRK_INT_X22Y99.  Is there any reason that an INT_R tile be split into a NULL tile and an INT_R tile that point to the same phy_tile?

Also, looking through the database it appears all of these issues occur along clock region boundaries.

I think I could fix the issue in a few different places by just ignoring tiles with type 94 (NULL).  Would it be reasonable to change the database generation to just not have NULL tiles point to any phy_tile?

litghost: > I think I could fix the issue in a few different places by just ignoring tiles with type 94 (NULL).

This sounds better than the current solution.

> Would it be reasonable to change the database generation to just not have NULL tiles point to any phy_tile?

That is a good solution if it helps here and doesn't break anything else.

andrewb1999:
Making fix elsewhere, as suggested.

PR #1545, opened 2020-06-23
diff_fasm: speedup generation of vendor tools bitstreams

acomodi:

### DIFF
--- a/xc/common/utils/vivado_create_runme.py
+++ b/xc/common/utils/vivado_create_runme.py
@@ -47,15 +47,43 @@ create_clock -period {period} -name {pin} -waveform {{0.000 {half_period}}} [get
 
     print(
         """
+
+write_checkpoint -force design_{name}_pre_route.dcp
+save_project_as -force design_{name}.xpr
+close_project
+
+open_checkpoint design_{name}_pre_route.dcp
+
+set conflict_nets [get_nets -filter {{ ROUTE_STATUS == "CONFLICTS" }}]
+if {{$conflict_nets != ""}} {{
+    puts "There are conflicting nets!"
+
+    foreach net $conflict_nets {{
+        puts $net
+    }}
+    error "Exiting!"
+}}
+
+set unrouted_nets [get_nets -filter {{ ROUTE_STATUS == "UNROUTED" }}]
+
+set antenna_nets [get_nets -filter {{ ROUTE_STATUS == "ANTENNAS" }}]
+
+set partial_nets [get_nets -filter {{ ROUTE_STATUS == "PARTIAL" }}]
+
+set nets_to_route "$unrouted_nets $antenna_nets $partial_nets"
+
+set_property IS_ENABLED 0 [get_drc_checks *]
+
+report_route_status
+foreach net $nets_to_route {{
+    route_design -nets [get_nets $net]
+}}
+
+set_property IS_ENABLED 1 [get_drc_checks *]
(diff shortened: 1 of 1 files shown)


mithro:
@acomodi -- What type of impact does this have on the speed?

xc/common/utils/vivado_create_runme.py
```diff
@@ -47,16 +47,47 @@ def create_runme(f_out, args):
...
+if {$conflict_nets != ""} {
+    puts "There are conflicting nets!"
+
+    foreach net $conflict_nets {
+        puts $net
+    }
+    error "Exiting!"
+}
+
+set unrouted_nets [get_nets -filter { ROUTE_STATUS == "UNDROUTED" }]
```
litghost, line 67: Misspelled UNROUTED?

litghost:
@acomodi Have you profiled diff fasm targets to verify the problem lies in the routing step?

acomodi:
I have gathered runtimes for the baselitex test.

Current run time of diff_fasm:
- source tcl: 10m15s
- place: 20m43s
- route: 2m41s
- write bitstream: 2m33s
- total time: 36m12s

With this PR:
- source tcl: 10m13s
- place: 0m0s (no placement performed)
- route: 0m16s
- write bitstream: 2m25s
- total: 12m54s

litghost:
> I have gathered runtimes for the baselitex test.
> 
> Current run time of diff_fasm:
> 
> * source tcl: 10m15s
> * place: 20m43s
> * route: 2m41s
> * write bitstream: 2m33s
> * total time: 36m12s
> 
> With this PR:
> 
> * source tcl: 10m13s
> * place: 0m0s (no placement performed)
> * route: 0m16s
> * write bitstream: 2m25s
> * total: 12m54s
> 
> We get a huge improvement by just avoiding the place step which is the most time consuming and is not actually needed.
> As discussed, the `source tcl` step is the next to improve.

What this data shows is that avoiding placement (or accelerating it) is the first key modification.  The changes to the route step are helpful, but I'd argue for leaving it alone until the place and source tcl time is down.

I'm wondering whether the route time drop is primilarily the focusing on certain nets, or disabling some of the DRC checks?

litghost:
@acomodi Should this be rebased?  I'm seeing a lot more changes than just the diff_fasm changes?

acomodi:
> @acomodi Should this be rebased? I'm seeing a lot more changes than just the diff_fasm changes?

Yeah I'll rebase now

xc/common/utils/vivado_create_runme.py
```diff
@@ -65,9 +93,8 @@ def create_runme(f_out, args):
 
 write_checkpoint -force design_{name}.dcp
 write_bitstream -force design_{name}.bit
-save_project_as -force design_{name}.xpr
```
litghost, line 68: Don't remove the XPR, that is needed for simulation in ModelSim.

acomodi: Sure, done

acomodi:
This lets XC7 vendor tools to take ~38 minutes less w.r.t. to master
For XC7 200 the time gained is ~20 minutes though.

Not the best improvement, but it is still worth more than nothing.

Next step would be to use the XDC constraints instead of sourcing the tcl

xc/common/utils/vivado_create_runme.py
acomodi: Done

PR #1561, opened 2020-06-30
Add support for synth tiles creation from vivado node names only

andrewb1999:

Allows easier custom partition region/roi synth tile creation by automatically picking the correct synth tile wire from only a vivado node name.

### DIFF
--- a/xc/common/cmake/device_define.cmake
+++ b/xc/common/cmake/device_define.cmake
@@ -189,7 +189,6 @@ function(ADD_XC_DEVICE_DEFINE_TYPE)
   set(DEVICE ${ADD_XC_DEVICE_DEFINE_TYPE_DEVICE})
   set(ROI_DIR ${ADD_XC_DEVICE_DEFINE_TYPE_ROI_DIR})
   set(TILE_TYPES ${ADD_XC_DEVICE_DEFINE_TYPE_TILE_TYPES})
-
   get_target_property_required(FAMILY ${ARCH} FAMILY)
   get_target_property_required(DOC_PRJ ${ARCH} DOC_PRJ)
   get_target_property_required(DOC_PRJ_DB ${ARCH} DOC_PRJ_DB)
@@ -287,6 +286,7 @@ function(ADD_XC_DEVICE_DEFINE_TYPE)
       LIMIT_GRAPH_TO_DEVICE FALSE
       )
   endif()
+
 endfunction()
 
 function(ADD_XC_DEVICE_DEFINE)
--- a/xc/common/utils/prjxray_create_synth_tiles.py
+++ b/xc/common/utils/prjxray_create_synth_tiles.py
@@ -36,11 +36,10 @@ WHERE tile_map.phy_tile_pkey = ? AND tile_type.name != 'NULL'
     return grid_x, grid_y
 
 
-def in_roi(roi, x, y):
-    return x >= roi.x1 and x <= roi.x2 and y >= roi.y1 and y <= roi.y2
-
-
 def tile_in_roi(conn, g, roi, tile_pkey):
+    """
+    Checks if the given tile_pkey is at a location within the specified roi
+    """
     c = conn.cursor()
     c.execute(
         """
@@ -53,10 +52,10 @@ SELECT name FROM phy_tile WHERE pkey =
     return roi.tile_in_roi(loc)
 
 
-#    return in_roi(roi, x, y)
-
-
 def wire_in_roi(conn, g, roi, wire_pkey):
+    """
+    Checks if the given wire_pkey is at a location within the specified roi
+    """
     c = conn.cursor()
     c.execute("""
 SELECT tile_pkey FROM wire WHERE pkey = ?
@@ -66,6 +65,10 @@ SELECT tile_pkey FROM wire WHERE pkey = ?
 
 
 def wire_manhattan_distance(conn, wire_pkey1, wire_pkey2):
+    """
+    Determines the manhattan distance between two tiles containing
+    the given wire_pkeys
+    """
     c = conn.cursor()
     c.execute(
         """
@@ -83,6 +86,14 @@ SELECT grid_x, grid_y FROM tile WHERE pkey = (SELECT tile_pkey FROM wire WHERE p
 
 
 def find_wire_from_node(conn, g, roi, node_name):
+    """
+    Finds a pair on wires in the given node such that:
+    1. One wire is inside the roi and the other is outside
+    2. The manhattan distance between these the wires is the minimum of any such pair
+
+    Returns the wire from this pair that is outside the roi and the tile the returned
+    wire is contained in.
+    """
     tile, node = node_name.split('/')
 
     cur = conn.cursor()
--- a/xc/common/utils/prjxray_create_synth_tiles.py
+++ b/xc/common/utils/prjxray_create_synth_tiles.py
@@ -16,11 +16,8 @@ def map_tile_to_vpr_coord(conn, tile):
     c.execute("SELECT pkey FROM phy_tile WHERE name = ?;", (tile, ))
     phy_tile_pkey = c.fetchone()[0]
 
-    c.execute("SELECT pkey FROM tile_type WHERE name = 'NULL'")
-    null_tile_type_pkey, = c.fetchone()
-
-    # It is expected that this tile has only one logical location,
-    # because why split a tile with no sites?
+    # Filters NULL tiles to prevent selecting two tiles that point
+    # to the same phy_tile
     c.execute(
         """
 SELECT tile_map.tile_pkey FROM tile_map INNER JOIN tile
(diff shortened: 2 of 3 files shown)


xc/common/cmake/device_define.cmake
```diff
@@ -173,7 +173,7 @@ endfunction()
 
 function(ADD_XC_DEVICE_DEFINE_TYPE)
   set(options)
-  set(oneValueArgs ARCH PART DEVICE ROI_DIR GRAPH_LIMIT)
+  set(oneValueArgs ARCH PART DEVICE ROI_DIR GRAPH_LIMIT PARTITION_DIR)
```
litghost, line 176: Need to add documentation on PARTITION_DIR

xc/common/utils/prjxray_arch_import.py
```diff
@@ -742,9 +743,13 @@ def main():
     synth_loc_map = {}
     synth_tile_map = {}
     roi = None
-    if args.use_roi:
-        with open(args.use_roi) as f:
-            j = json.load(f)
+    if args.use_roi or args.partition_region:
```
litghost, line 746: Given that partition_region and roi in this case do the example same thing, why do we need another argument?

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -70,33 +179,50 @@ def main():
...
+                tile, wire = find_wire_from_node(conn, g, roi, port['node'])
+            else:
+                tile, wire = port['wire'].split('/')
 
-            vbrk_in_use.add(tile)
+            tile_in_use.add(tile)
 
             # Make sure connecting wire is not in ROI!
             loc = g.loc_of_tilename(tile)
+            print(loc)
```
litghost, line 218: Probably want to remove debug prints, or atleast put them behind a verbose flag?

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -57,8 +160,14 @@ def main():
     synth_tiles = {}
     synth_tiles['tiles'] = {}
 
-    with open(args.roi) as f:
-        j = json.load(f)
+    if args.roi:
```
litghost, line 163: Again, if the logic between --roi and --partition_region are the same, why do we need a new flag?

xc/xc7/archs/artix7/devices/xc7a50t-arty-swbut-pr1-virt/design.json
```diff
@@ -0,0 +1,58 @@
...
+    "info": {
+        "GRID_X_MAX": 59,
+        "GRID_X_MIN": 12,
+        "GRID_Y_MAX": 51,
+        "GRID_Y_MIN": 0
+    },
+    "ports": [
+        {
+            "name": "din[0]",
+	    "type": "in",
```
litghost, line 11: Inconsistent space formatting?

xc/common/utils/prjxray_arch_import.py
andrewb1999: Yes I agree it's unnecessary right now.  I'm using it for other related work (like heterogeneous synth tiles).  My thought was to leave it in for now to prevent having to make major changes to cmake when I post a pr for that in a day or two.  But maybe that's not the best idea...

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -70,33 +179,50 @@ def main():
...
-            if port['name'].startswith('dout['):
-                port_type = 'input'
-                is_clock = False
-            elif port['name'].startswith('din['):
-                is_clock = False
-                port_type = 'output'
-            elif port['name'].startswith('clk'):
-                port_type = 'output'
-                is_clock = True
+            if args.roi:
```
litghost, line 184: I'm wondering if the logic here can be unified to use your method, rather than having a giant `if` here?

For example, have the wire specified, and have the code go from wire -> node, and pick a suitable location.

I like having an explicitly type, why not propigate that to the ROI definition files?

xc/common/utils/prjxray_arch_import.py
litghost: As far as I can tell, your work is a pure superset of the original ROI concept.  I believe we should be treating it like an extension, rather than a side by side piece?

xc/common/utils/prjxray_create_synth_tiles.py
andrewb1999: I think this flag is still necessary for the distinction at line 209

xc/common/utils/prjxray_create_synth_tiles.py
andrewb1999: We can do that too.  Just was trying to limit how much I touched regular roi stuff, but if you're ok with it I can propagate to normal ROI definitions.

litghost: > For example, have the wire specified, and have the code go from wire -> node, and pick a suitable location.

I just looked, and the current ROI format specifies both the wire and the node.  Maybe simply have the logic go "if wire is not specified, pick a wire, otherwise check if the specified wire is legal based on the selection logic".  That enables both automatic and fine grained control, if needed.

litghost: > Just was trying to limit how much I touched regular roi stuff

I'd rather update the ROI stuff than maintain two extremely similar code paths.

andrewb1999: Ok that sounds good.

andrewb1999:
Will now require a new prjxray-db generated with https://github.com/SymbiFlow/prjxray/pull/1383 but in the mean time let me know how these changes look.

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -70,33 +175,37 @@ def main():
...
             if roi.tile_in_roi(loc):
                 # Or if in the ROI, make sure it has no sites.
                 gridinfo = g.gridinfo_at_tilename(tile)
-                assert len(
-                    db.get_tile_type(gridinfo.tile_type).get_sites()
-                ) == 0, tile
+                assert len(db.get_tile_type(gridinfo.tile_type).get_sites()
+                           ) == 0, "{}/{}".format(tile, wire)
 
+            print(tile)
```
litghost, line 208: I believe we can remove this debug print?

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -39,11 +36,116 @@ def map_tile_to_vpr_coord(conn, tile):
     return grid_x, grid_y
 
 
+def in_roi(roi, x, y):
```
litghost, line 39: Docstring and some doctests would be good here

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -39,11 +36,116 @@ def map_tile_to_vpr_coord(conn, tile):
...
+    return x >= roi.x1 and x <= roi.x2 and y >= roi.y1 and y <= roi.y2
+
+
+def tile_in_roi(conn, g, roi, tile_pkey):
+    c = conn.cursor()
```
litghost, line 44: Docstring

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -39,11 +36,116 @@ def map_tile_to_vpr_coord(conn, tile):
...
+    tile, = c.fetchone()
+    loc = g.loc_of_tilename(tile)
+    return roi.tile_in_roi(loc)
+
+
+#    return in_roi(roi, x, y)
+
+
+def wire_in_roi(conn, g, roi, wire_pkey):
+    c = conn.cursor()
```
litghost, line 60: Docstring

xc/common/utils/prjxray_create_synth_tiles.py
```diff
@@ -39,11 +36,116 @@ def map_tile_to_vpr_coord(conn, tile):
...
+        """
+SELECT grid_x, grid_y FROM tile WHERE pkey = (SELECT tile_pkey FROM wire WHERE pkey = ?)
+""", (wire_pkey2, )
+    )
+    x2, y2 = c.fetchone()
+    return abs(x1 - x2) + abs(y1 - y2)
+
+
+def find_wire_from_node(conn, g, roi, node_name):
+    tile, node = node_name.split('/')
```
litghost, line 86: Docstring

xc/common/utils/prjxray_create_synth_tiles.py
andrewb1999: I was actually able to just remove this function.  I think the other functions I added would be very challenging to write a doctest for given their reliance on the connection database.

litghost: > I think the other functions I added would be very challenging to write a doctest for given their reliance on the connection database.

Yep, which is why I didn't mention them

Issue #1546, opened 2020-06-24
fasm2bels: max recursion depth exceeded, make_routes.py, xc7a100t, soc/litex/base


tcal-x:
Error seen:
```
[100%] Generating baselitex_arty100t/artix7_100t-xc7a100t-virt-xc7a100t-test/top_bit.v
Traceback (most recent call last):
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/build/env/conda/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/build/env/conda/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/__main__.py", line 4, in <module>
    main()
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/fasm2bels.py", line 403, in main
    top.make_routes(allow_orphan_sinks=args.allow_orphan_sinks)
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/verilog_modeling.py", line 1502, in make_routes
    net_map=self.net_map,
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/make_routes.py", line 709, in make_routes
    allow_orphan_sinks=allow_orphan_sinks
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/make_routes.py", line 583, in expand_sink
    allow_orphan_sinks=allow_orphan_sinks
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/make_routes.py", line 453, in expand_sink
    allow_orphan_sinks=allow_orphan_sinks
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/make_routes.py", line 583, in expand_sink
    allow_orphan_sinks=allow_orphan_sinks
...
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/xc7/fasm2bels/make_routes.py", line 453, in expand_sink
    allow_orphan_sinks=allow_orphan_sinks
[...]

Issue #1542, opened 2020-06-23
prjxray_create_place_constraints.py fails for unconstrained PLLE2_ADV on xc7a100T device


tcal-x:
Error seen:
```
Traceback (most recent call last):
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 693, in <module>
    main()
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 681, in main
    vpr_grid, loc_in_use, block_locs, blocks, grid_capacities):
  File "/home/tcal/2nd-tcal-x/symbiflow-arch-defs/xc/common/utils/prjxray_create_place_constraints.py", line 514, in place_clocks
    for potential_loc in sorted(available_placements[key]):
KeyError: ('PLLE2_ADV', 5)
```

To reproduce:
* Sync to current head of master (a8c60d95 or later).
* Rerun "make env" to rerun cmake to get the new 100T test targets.
* Edit the file `xc/xc7/tests/soc/litex/mini_ddr/minilitex_ddr_arty100t.v` as follows:
  * Delete or comment out the line `(* LOC="PLLE2_ADV_X1Y1" *)`
* `cd` to `build/xc/xc7/tests/soc/litex/mini_ddr`.
* Run `make minilitex_ddr_arty100t_bin` to produce the error.
  * This will take several hours the first time, so run overnight.
  * It takes less than a minute to get to the error in subsequent runs.

Notes:
* The original 35T design constrained the IDELAYCTRL to X1Y0.  With the 100T it is constrained to X1Y1.
* In the original design, the PLLE2_ADV did not need to be constrained.   With 100T, if it is not constrained, we encounter the script error above.   If it _is_ constrained to X1Y1, then everything works.

This is not blocking, since there is the workaround of adding the LOC constraint to the PLLE2_ADV.

tcal-x:
Hi @litghost , who is the best person to own this?   It involves the constraint solver in the clock placement script.   This should be  a pretty easy problem to solve, but I can't find anything obvious, such as a hardcoding of clock region X1Y0 (which wouldn't work on 100T)

litghost:
[...]

PR #1563, opened 2020-07-01
New 100t LiteX test (soc/litex/base)

tcal-x:
Add Linux-capable LiteX test (soc/litex/base)  for xc7a100t, including `vivado_diff_fasm`.

The `vivado_diff_fasm` test takes about 5hrs -- 3hrs for SymbiFlow and 2hrs for Vivado.

The implementation has setup violations (-6ns slack) related to `netsoc_csrbankarray_interface5_bank_bus` but shows basic functionality on the Arty A7 100T board.

Note: this is the test where I saw aberrant behavior (issue #1546) from vpr creating a path > 1000, which then crashed make_routes.py in fasm2bels due to hitting the maximum recursion limit in Python.   This issue went away when I rebased the code, so I assume it must just pop up randomly.


xc/xc7/tests/soc/litex/base/README.md
```diff
@@ -23,3 +23,7 @@ export CPU_VARIANT=linux
 export PLATFORM=arty
 export FIRMWARE=linux
 ```
+
+To create `baselitex_arty100t.v`, copy from `baselitex_arty.v` and then make the following edits:
+* *Add* constraint `(* LOC="PLLE2_ADV_X1Y1" *)` to instance `PLLE2_ADV`.
```
litghost, line 28: This shouldn't be required, we should identify and fix the underlying bug that resulted in this workaround.

tcal-x: I agree, I filed issue #1542 last week on this issue.   When this test is merged, I'll mention it in the issue so that we remember to come back and take out the workaround when the issue is fixed.

tcal-x:
@acomodi thanks for the force run.   The run took 7:30, which is longer than previous xc7 vendor runs, which are in the range 5:15--6:30.   I may rerun to see if it might have been due to a heavy load on Kokoro at the time.

@litghost , do you consider a one-hour increase in runtime acceptable?   @acomodi's improvements in the pipeline should bring it back down...I guess we can wait until they land and then re-evaluate the delta for adding this test.

litghost:
> @litghost , do you consider a one-hour increase in runtime acceptable? @acomodi's improvements in the pipeline should bring it back down...I guess we can wait until they land and then re-evaluate the delta for adding this test.

Which CI?  If 7-series jumps by an hour that's not great (2.5 hrs -> 3.5 hrs).  If 7-series Vendor tests jumps, that's likely okay (6.5 hrs -> 7.5 hrs).

tcal-x:
> > @litghost , do you consider a one-hour increase in runtime acceptable? @acomodi's improvements in the pipeline should bring it back down...I guess we can wait until they land and then re-evaluate the delta for adding this test.
> 
> Which CI? If 7-series jumps by an hour that's not great (2.5 hrs -> 3.5 hrs). If 7-series Vendor tests jumps, that's likely okay (6.5 hrs -> 7.5 hrs).

Right, it's the latter case, the xc7 vendor tests, so now up to 7.5 hrs before considering Alessandro's improvements.

litghost:
> > > @litghost , do you consider a one-hour increase in runtime acceptable? @acomodi's improvements in the pipeline should bring it back down...I guess we can wait until they land and then re-evaluate the delta for adding this test.
> > 
> > 
> > Which CI? If 7-series jumps by an hour that's not great (2.5 hrs -> 3.5 hrs). If 7-series Vendor tests jumps, that's likely okay (6.5 hrs -> 7.5 hrs).
> 
> Right, it's the latter case, the xc7 vendor tests, so now up to 7.5 hrs before considering Alessandro's improvements.

Ya, that's fine.  In this case, 6.5 hrs is all day, and 7.5 hrs is still all day.  IMO, not a big qualitative difference.  Hopefully someone has time to see if we can shrink the TCL runtime, and bring the time down even more.  There is also likely easy gains in FASM parsing (textx replaced with something0.

tcal-x:
The last two xc7-vendor tests have taken 5:38 and 5:37.

litghost:
Awesome, thanks!

PR #1564, opened 2020-07-01
lcu: add test to check the correctness of LCU.

acomodi:

This PR adds a very simple LCU test.

LCU are not currently enabled in the current master+wip yosys, therefore the comparison operations will not be synthesized using the LCU arithmetic of arith_map.v

The newer version of yosys master+wip though enables LCU inference, and this needs to be correctly tested.

The test has two 16-bit wide counters (a counter and a threshold) on which the comparison operations are performed.


acomodi:
> This needs a testbench to verify the pre- and post-results.

@litghost sure, done

litghost:
@acomodi We should also add a count-down version of the test.  This ensures that both CYINIT=1 and CYINIT=-1 are tested.

Issue #1550, opened 2020-06-25
Support multiple synth IO tiles at the same vpr coordinate


andrewb1999:
For custom harnesses and partition regions, there should be the ability to place multiple synth IO tiles at the same VPR coordinate, connecting to different wires at the border of the ROI.

I think this can be implemented relatively easily using by increasing the tile capacity, but I am unsure how to handle an INPAD and and OUTPAD at the same vpr coordinate.  My initial thought is to merge these together into an SYN-IOPAD and maybe have the INPAD and OUTPAD as equivalent sites.

andrewb1999:
@litghost Can you let me know your thoughts?

litghost:
So @acomodi add the feature to support this, which is heterogeneous tiles.  There are documentation and examples upstream.  I'll find a link.

litghost:
Link to doc: https://docs.verilogtorouting.org/en/latest/tutorials/arch/heterogeneous_tiles/#heterogeneous-tiles-tutorial

PR #1565, opened 2020-07-01
Heterogeneous synth tiles

andrewb1999:

Made to solve #1550 by allowing multiple synth ios to be placed at the same grid location using increased capacity and heterogeneous tiles.


andrewb1999:
@litghost can you take a look at this and see if it makes sense?  I'm able to fully generate the architecture and rr_graphs, but the io placement constraints do not seem to be working correctly.

litghost:
> io placement constraints do not seem to be working correctly.

Ya, they were not design to handle stacked IO tiles.  You will need to enhance that code.

andrewb1999:
> > io placement constraints do not seem to be working correctly.
> 
> Ya, they were not design to handle stacked IO tiles. You will need to enhance that code.

I'm not surprised, but I do not understand where the issue is arising.  To me, the top_constraints.place seems reasonable for the case where there's a single input or output at each vpr grid location because the z location should still be 0 afaik.  Is the issue the VPR code that determines if it is a valid placement constraint?

litghost:
> I'm not surprised, but I do not understand where the issue is arising. To me, the top_constraints.place seems reasonable for the case where there's a single input or output at each vpr grid location because the z location should still be 0 afaik. Is the issue the VPR code that determines if it is a valid placement constraint?

Sorry, I might be missing something.  If there is only 1 IO tile at the location, z=0 is correct.  I had assumed you were referring to a stacked instance.

So what is the problem that you are having now?

andrewb1999:
> > I'm not surprised, but I do not understand where the issue is arising. To me, the top_constraints.place seems reasonable for the case where there's a single input or output at each vpr grid location because the z location should still be 0 afaik. Is the issue the VPR code that determines if it is a valid placement constraint?
> 
> Sorry, I might be missing something. If there is only 1 IO tile at the location, z=0 is correct. I had assumed you were referring to a stacked instance.
> 
> So what is the problem that you are having now?

Even in the case with only 1 IO at the location, I get the following error:

```Reading locations of IO pads from '/scratch/safe/butta/symbiflow-arch-defs.hetero/build/xc/xc7/tests/buttons/buttons_arty/artix7-xc7a50t-arty-swbut-roi-virt-xc7a50t-arty-swbut-test/top_constraints.place'.
# Placement took 0.22 seconds (max_rss 664.4 MiB, delta_rss +0.0 MiB)
Error 1:
Type: Placement file
File: vpr/src/base/read_place.cpp
Line: 255
Message: build/xc/xc7/tests/buttons/buttons_arty/artix7-xc7a50t-arty-swbut-roi-virt-xc7a50t-arty-swbut-test/top_constraints.place
```

And this is what the top_constraints.place file looks like (which doesn't appear to be trying to place multiple IOs at the same grid location):
```
#Block Name x   y   z    pcf_line
#---------- --  --  -    ----
sw[0]       10  48  0  # set_io sw[0] A8
sw[1]       10  44  0  # set_io sw[1] C11
sw[2]       10  40  0  # set_io sw[2] C10
sw[3]       10  36  0  # set_io sw[3] A10
sw[4]       10  19  0  # set_io sw[4] D9
sw[5]       10  32  0  # set_io sw[5] C9
sw[6]       10  28  0  # set_io sw[6] B9
sw[7]       10  23  0  # set_io sw[7] B8
out:led[0]  62  26  0  # set_io led[0] F6
out:led[1]  62  22  0  # set_io led[1] J4
out:led[2]  62  18  0  # set_io led[2] J2
out:led[3]  62  14  0  # set_io led[3] H6
out:led[4]  62  35  0  # set_io led[4] H5
out:led[5]  62  31  0  # set_io led[5] J5
out:led[6]  10  35  0  # set_io led[6] T9
out:led[7]  10  31  0  # set_io led[7] T10

```

litghost:
> I get the following error:

That error message is not great, it needs to be fixed.  Anyways, if you look here: https://github.com/SymbiFlow/vtr-verilog-to-routing/blob/master%2Bwip/vpr/src/base/read_place.cpp#L255
you will see that is a illegal placement based on type.  This makes me suspect that the I or O cluster you are trying to place doesn't match that location.  Did the in/out get reversed?

utils/lib/rr_graph/graph2.py
```diff
@@ -484,6 +484,13 @@ def create_pin_name_from_tile_type_and_pin(
     ):
         return '{}.{}[{}]'.format(tile_type, port_name, pin_idx)
 
+    def create_pin_name_from_tile_type_sub_tile_num_and_pin(
```
litghost, line 487: Needs a docstring

xc/common/utils/prjxray_arch_import.py
```diff
@@ -116,6 +116,187 @@ def create_synth_io_tiles(complexblocklist_xml, tiles_xml, pb_name, is_input):
...
+    )
+
+    tile_xml = ET.SubElement(tiles_xml, 'tile', {
+        'name': 'SYN-IOPAD',
+    })
+
+    sub_tile_xml_in = ET.SubElement(
+        tile_xml, 'sub_tile', {
+            'name': 'SYN_IN_SUB_TILE',
+            'capacity': '50'
```
litghost, line 141: This makes doesn't make sense to me?  Shouldn't you be taking in a list of in/out tiles, or tile counts?

xc/common/utils/prjxray_arch_import.py
```diff
@@ -116,6 +116,187 @@ def create_synth_io_tiles(complexblocklist_xml, tiles_xml, pb_name, is_input):
...
+        }
+    )
+
+    port_pin_in = '{}.{}'.format('SYN-INPAD', pad_name_in)
+    pad_pin_in = '{}.{}'.format(pad_name_in, pad_name_in)
+
+    port_pin_out = '{}.{}'.format('SYN-OUTPAD', pad_name_out)
+    pad_pin_out = '{}.{}'.format(pad_name_out, pad_name_out)
+
+    ET.SubElement(site_in, 'direct', {'from': port_pin_in, 'to': port_pin_in})
```
litghost, line 223: This is probably wrong, because neither port_pin_in nor port_pin_in has indicies.  @acomodi please review too!

xc/common/utils/prjxray_arch_import.py
```diff
@@ -550,17 +731,26 @@ def get_none_tile_prefix(single_xml):
         yield vpr_tile_type, grid_x, grid_y, meta_fun
 
 
-def add_synthetic_tiles(model_xml, complexblocklist_xml, tiles_xml, need_io):
+def add_synthetic_tiles(
```
litghost, line 734: There shouldn't need to be two add_synthetic_tiles tile functions?  add_synthetic_tiles(num_input=1) and add_synthetic_tiles(num_output=1) should exactly the same?

xc/common/utils/prjxray_routing_import.py
```diff
@@ -865,10 +866,12 @@ def add_synthetic_edges(conn, graph, node_mapping, grid, synth_tiles):
             assert len(option) > 0, (pin, len(option))
 
             if pin['port_type'] == 'input':
-                tile_type = 'SYN-OUTPAD'
+                tile_type = 'SYN-IOPAD'
```
litghost, line 869: This is very wrong.  The tiletype should be SYN-INPAD or SYN-OUTPAD?

xc/common/utils/prjxray_arch_import.py
andrewb1999: My thought is that it doesn't matter as long as the capacity is higher than the number of synth tiles at the location.  As long as the constraints have the IOs at the correct Z location it should work fine right? Although it is probably more correct to have the capacity match the number of synth IOs that can be placed there.

litghost: Kind of, but during routing import you will have tiles with no connections, which is weird.  Best to avoid that.  The whole point of heterogeneous tiles is to precisely describe situations like this.

xc/common/utils/prjxray_routing_import.py
andrewb1999: You're probably right.  Trying now.

litghost: This code will definitely need to understand the z-stacking too.

xc/common/utils/prjxray_arch_import.py
litghost: More specifically, somewhere in the flow coordinates should be choosen, and propigated to arch_import, routing_import and the IO placement description.

xc/common/utils/prjxray_routing_import.py
andrewb1999: In the graph pin_name_map, the key it expects is of the form SYN-IOPAD[x].inpad[x] which is why I made this change here from SYN-OUTPUT to SYN-IOPAD.  Does this mean the pin_name_map is being produced incorrectly somewhere else?

xc/common/utils/prjxray_arch_import.py
andrewb1999: Ok that makes sense.  Would like to get the heterogeneous tiles working in the single IO per tile case first and then I will work on this.

xc/common/utils/prjxray_routing_import.py
litghost: That actually makes sense, as the routing interface belongs to the tile.  You need to take care to select the right IPIN or OPIN to bind the synth edge too, otherwise things won't line up right.

andrewb1999:
Might help with debugging to see the architecture currently generated: [arch.timing.xml](https://github.com/andrewb1999/summer-of-code-resources/blob/master/arch.timing.xml)

litghost:
> Might help with debugging to see the architecture currently generated: [arch.timing.xml](https://github.com/andrewb1999/summer-of-code-resources/blob/master/arch.timing.xml)

I would examine warnings from the VPR output around "SYN-".  I believe:
```
          <site pb_type="SYN-INPAD">
            <direct from="SYN-INPAD.inpad" to="SYN-INPAD.inpad"/>
          </site>
```

is an invalid direct, as non of the outer containers have SYN-INPAD.

acomodi:
@andrewb1999 Looking at the generated [arch.timing.xml](https://github.com/andrewb1999/summer-of-code-resources/blob/master/arch.timing.xml) it looks like that there is an issue with the direct name.

Incorrect:
```
          <site pb_type="SYN-INPAD">
            <direct from="SYN-INPAD.inpad" to="SYN-INPAD.inpad"/>
          </site>
```

Correct:
```
          <site pb_type="SYN-INPAD">
            <direct from="SYN_IN_SUB_TILE.inpad" to="SYN-INPAD.inpad"/>
          </site>
```

The `from` value in the direct map should have the name of the sub tile the pb_type belongs to.

I'll took a more detailed look at the PR in a bit to give some more feedback.

xc/common/utils/prjxray_arch_import.py
acomodi: I believe indices are actually not strictly needed as VPR will handle that if capacity is > 1 for the sub_tile corresponding to this site.

I think also here the `from` should be in the form of `<sub_tile_name>.<pin_name>` to have a valid direct

andrewb1999:
@litghost @acomodi I made these changes and am still running into the same issue during placement.  Here is the current [arch.timing.xml](https://github.com/andrewb1999/summer-of-code-resources/blob/master/arch.timing.xml)

acomodi:
@andrewb1999 I might know why you get the placement error actually, I didn't think about that in the first place.

The problem is that you currently have a SYN-IOPAD tile that has two sub-tiles, each one with `N` capacity.

The fact is that, given that the SYN-OUTPAD sub tile is defined after the SYN-INPAD one, its first capacity location will have an offset equal to `N`. So all the outputs should have the z location set to the `N` offset forward.

andrewb1999:
> @andrewb1999 I might know why you get the placement error actually, I didn't think about that in the first place.
> 
> The problem is that you currently have a SYN-IOPAD tile that has two sub-tiles, each one with `N` capacity.
> 
> The fact is that, given that the SYN-OUTPAD sub tile is defined after the SYN-INPAD one, its first capacity location will have an offset equal to `N`. So all the outputs should have the z location set to the `N` offset forward.

That makes sense.  Will work on a fix for that.

andrewb1999:
@litghost If I were to have a different capacity for each synth tile location based on the number of IOs to be placed there, would I have to create a new tile definition for each location, or is there some way to dynamically alter the capacity at each location?

litghost:
> @litghost If I were to have a different capacity for each synth tile location based on the number of IOs to be placed there, would I have to create a new tile definition for each location, or is there some way to dynamically alter the capacity at each location?

Different tile for each instance.  A simple name template should be fine here.

xc/common/utils/prjxray_create_ioplace.py
```diff
@@ -9,6 +9,13 @@
 from lib.parse_pcf import parse_simple_pcf
 
 
+def get_synth_tile_from_pad(synth_tiles, pad):
+    for _, tile in synth_tiles['tiles'].items():
+        for pin in tile['pins']:
+            if pin['pad'] == pad:
+                return tile
```
litghost, line 16: Should have an error if this falls off the end? Also docstring?

andrewb1999:
@litghost This seems to work with all current uses of synth tiles.  I have tested it in cases with stacked tiles, but am cautious about including one now as there is no overlay generation required to test the bistreams.  I think I would rather have this work merged and then post overlay generation work soon with a full test case.

xc/common/utils/prjxray_create_ioplace.py
```diff
@@ -128,6 +145,21 @@ def main():
             sys.exit(1)
 
         loc, is_output, iob = pad_map[pcf_constraint.pad]
+
+        if args.synth_tiles:
+            tile = get_synth_tile_from_pad(synth_tiles, pcf_constraint.pad)
+            if tile:
```
litghost, line 151: When would `get_synth_tile_from_pad` return None, and it not be an error?

andrewb1999: In the case of overlays, where the architecture has both synth IOs and real IOs (so not all pcf_constraint pads would be in the synth_tiles).

litghost: Ah, sure.  Please add a comment about that, that is not obvious, but does make sense in retrospect.  Should there be logic to make sure that it isn't possible to have an ambiguous match between the synth tile and the "real" IO tiles?

andrewb1999:
> Comments are in. I think this is looking pretty good. What is and isn't working right now?

From my testing and running CI before a rebase, everything seems to be working now.  We cannot yet have a full test (through working bitstream) for multiple IOs at the same vpr location because overlay support isn't finished yet (but I need this work to finish overlay generation).

litghost:
> @litghost This seems to work with all current uses of synth tiles. I have tested it in cases with stacked tiles, but am cautious about including one now as there is no overlay generation required to test the bistreams. I think I would rather have this work merged and then post overlay generation work soon with a full test case.

That's great news!  I'm fine with that approach.  Do you believe this PR will go green, or do you need https://github.com/SymbiFlow/prjxray/pull/1389 merged and propigated?

litghost:
@acomodi Please re-review, especially if this goes green in CI.

andrewb1999:
> That's great news! I'm fine with that approach. Do you believe this PR will go green, or do you need [SymbiFlow/prjxray#1389](https://github.com/SymbiFlow/prjxray/pull/1389) merged and propigated?

I expect this PR to go green without [SymbiFlow/prjxray#1389](https://github.com/SymbiFlow/prjxray/pull/1389), but the propagation of that fix it required for #1561 which is also necessary to finish the overlay generation work.

xc/common/utils/prjxray_create_ioplace.py
andrewb1999: I don't think this should have an error based on the discussion below.  Is that correct?

litghost: Agree in principle.  I believe it is worth having an explicit comment that it is okay to return `None`, because ...

andrewb1999:
@litghost @acomodi Looks like there is a single vendor tools failure for dram_test_32x1d:

>[636/2644] cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test && diff -u /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_dram_test_32x1d_vivado.bit.fasm

>FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test && diff -u /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_dram_test_32x1d_vivado.bit.fasm
--- /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit.fasm	2020-07-08 11:51:36.277679549 -0700
+++ /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_dram_test_32x1d_vivado.bit.fasm	2020-07-08 11:57:36.553386778 -0700

>@@ -2156,9 +2156,11 @@
 >INT_L_X10Y115.ER1BEG_S0.WW4END_S0_0
>
 >INT_L_X10Y116.EL1BEG2.NL1BEG_N3
>-INT_L_X10Y116.ER1BEG2.WW2END1
>+INT_L_X10Y116.ER1BEG2.SR1END1
 >INT_L_X10Y116.NL1BEG_N3.WW4END0
>
>+INT_L_X10Y117.SR1BEG1.WR1END1
>+
 >INT_L_X12Y113.NL1BEG0.WL1END0
 >INT_L_X12Y113.NW2BEG1.WL1END0
>
>@@ -2280,7 +2282,6 @@
> INT_L_X12Y116.SW2BEG1.LOGIC_OUTS_L23
> INT_L_X12Y116.WL1BEG0.WL1END1
> INT_L_X12Y116.WL1BEG_N3.WW2END0
>-INT_L_X12Y116.WW2BEG1.SR1END1
>
 >INT_L_X12Y117.BYP_ALT0.GFAN0
 >INT_L_X12Y117.BYP_ALT1.WR1END1
>@@ -2314,7 +2315,6 @@
 >INT_L_X12Y117.SE2BEG1.EL1END1
 >INT_L_X12Y117.SL1BEG2.SR1END2
 >INT_L_X12Y117.SL1BEG3.SW2END3
>-INT_L_X12Y117.SR1BEG1.SW2END0
 >INT_L_X12Y117.SR1BEG2.WR1END2
 >INT_L_X12Y117.SR1BEG3.NN2END3
 >INT_L_X12Y117.SW2BEG0.WL1END0
>@@ -3981,6 +3981,7 @@
 >INT_R_X11Y117.FAN_ALT0.NL1END0
 >INT_R_X11Y117.SL1BEG2.SR1END2
 >INT_R_X11Y117.SL1BEG3.SW2END3
>+INT_R_X11Y117.WR1BEG1.WL1END_N1_3
>
> INT_R_X11Y118.EE2BEG3.NN2END3
> INT_R_X11Y118.EE4BEG2.NN2END2

Not sure why this is showing up now, and only for one test case, but can you take a look at this?  Going to rebase to fix merging issues and let it run again overnight incase it's a fluke.

andrewb1999:
@litghost @acomodi Same vendor tools failure.  Any idea why this would happen for only a single test?  The fasm diff seems to be in the middle of the roi so I'm not sure how my changes would cause that.  Possibly a rebase error?

litghost:
> @litghost @acomodi Same vendor tools failure. Any idea why this would happen for only a single test? The fasm diff seems to be in the middle of the roi so I'm not sure how my changes would cause that. Possibly a rebase error?

I don't see how it could be a rebase error.  My guess is that this is either a regression, or a latent bug.

As for why it only happened in one test, I don't have a good theory.

In terms of debugging, it is probably worth starting at if the VPR nets were valid to begin with.  If VPR output an invalid route, that may indicate that the bug is at or near the routing import synth tile logic.  If the routing import logic connects to the wrong RR inode, I could cause some weird behaviors.

acomodi:
@andrewb1999 Hard to say. I see though that the failing test is the DRAM32X1D, which is a DRAM mode that is not entirely supported at the moment by symbiflow. From my experience with this kind of errors, it could be due to a bug in fasm2bels which leads to constant net routing difference in Vivado.

[Here is the bitstream](https://storage.cloud.google.com/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/presubmit/xc7_vendor/1194/20200709-065214/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit) produced by VPR.

By using that to generate the `dcp` we could get more insights on the issue. I'll try to give a look first thing tomorrow morning.

andrewb1999:
> @andrewb1999 Hard to say. I see though that the failing test is the DRAM32X1D, which is a DRAM mode that is not entirely supported at the moment by symbiflow. From my experience with this kind of errors, it could be due to a bug in fasm2bels which leads to constant net routing difference in Vivado.
> 
> [Here is the bitstream](https://storage.cloud.google.com/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/presubmit/xc7_vendor/1194/20200709-065214/build/xc/xc7/tests/dram_test/dram_test_32x1d/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/top.bit) produced by VPR.
> 
> By using that to generate the `dcp` we could get more insights on the issue. I'll try to give a look first thing tomorrow morning.

Ran fasm2bels and generated the dcp: [design.dcp](https://github.com/andrewb1999/summer-of-code-resources/blob/master/design.dcp)

Everything seems reasonable to me so far.  Investigating more.

acomodi:
By looking at the DCP there is indeed one net that did not get constrained correctly:
Screenshot from 2020-07-10 11-43-03

This may suggest that there might be an issue in fasm2bels. Unsure why this showed up in this PR, as it seems to be unrelated, but we would need to fix it. I'll try to replicate locally.

andrewb1999:
> By looking at the DCP there is indeed one net that did not get constrained correctly:
> Screenshot from 2020-07-10 11-43-03
> 
> This may suggest that there might be an issue in fasm2bels. Unsure why this showed up in this PR, as it seems to be unrelated, but we would need to fix it. I'll try to replicate locally.

Oh, I see now.  It becomes much easier to see before routing the design in vivado:
image

I do not know how to debug this yet, but willing to help with any debugging you would like me to try.

andrewb1999:
@litghost @acomodi Here is the preroute dcp if you have time to look at debugging this: [design_preroute.dcp](https://github.com/andrewb1999/summer-of-code-resources/blob/master/design_preroute.dcp)

andrewb1999:
@acomodi  @litghost  Just traced back the pips through the preroute dcp while looking at the fasm.  The vpr fasm contains all the correct pips for this net to be routed, so it must be a fasm2bels issue.  My guess now is that my changes to synth IOs are causing VPR to place and route slightly differently than before, which has exposed this latent bug.

litghost:
@andrewb1999: I believe @acomodi found the error and mentioned it earlier.  fasm2bels missed constraining a net, which resulted in the differences we see.  This was not caused by your change, but it did uncover it.  Hopefully next week @acomodi or I could propose a fix.

In the meantime, are you blocked because of this PR not being merged until this gets root caused and fixed?

andrewb1999:
> @andrewb1999: I believe @acomodi found the error and mentioned it earlier. fasm2bels missed constraining a net, which resulted in the differences we see. This was not caused by your change, but it did uncover it. Hopefully next week @acomodi or I could propose a fix.
> 
> In the meantime, are you blocked because of this PR not being merged until this gets root caused and fixed?

This blocks all the rest of my work being upstreamed, but I still have plenty else to work on locally so I'm fine waiting a bit while a fix is created.  Just trying to finalize some work to get it off my plate.

andrewb1999:
@litghost after a rebase, this is now failing the litex design instead of the dram test.  Can you look at why this is failing now?

acomodi:
This most probably relates to https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1607. It shouldn't be related to the changes in this PR

andrewb1999:
> This most probably relates to #1607. It shouldn't be related to the changes in this PR

Ok.  The good news is dram_test_32x1d is no longer failing the diff fasm, so this should be able to be merged soon.  Do the vendor tool tests need to be rerun?

litghost:
> > This most probably relates to #1607. It shouldn't be related to the changes in this PR
> 
> Ok. The good news is dram_test_32x1d is no longer failing the diff fasm, so this should be able to be merged soon. Do the vendor tool tests need to be rerun?

Ya a re-run is a good idea.  Status should update in a couple minutes.  I'll do code review now.

utils/lib/rr_graph/graph2.py
litghost: Still needs a docstring?  Maybe you forgot to push?

acomodi:
> Do the vendor tool tests need to be rerun?

@andrewb1999 Yes, that would be best, as some of the tests were not completed due to the `bad_alloc` failure

xc/common/utils/prjxray_create_ioplace.py
```diff
@@ -144,6 +164,21 @@ def main():
             sys.exit(1)
 
         loc, is_output, iob = pad_map[pad]
+
+        if args.synth_tiles:
```
litghost, line 168: Something that is confusing me now is that pad_map contains an entry for the pad, so why is this if statement needed?  Is the data from pad_map wrong?  Why not just fix that data instead of adding the code here?  Feels inconsistent to me.  Thoughts?

xc/common/utils/prjxray_create_ioplace.py
```diff
@@ -44,6 +54,12 @@ def main():
         default=sys.stdout,
         help='The output io.place file'
     )
+    parser.add_argument(
```
litghost, line 57: I don't see any CMake change to actually pass `--synth_tiles` to create_ioplace.  Is this code even needed?

xc/common/utils/prjxray_create_ioplace.py
andrewb1999: So my thought is that for something like an overlay or an roi that aligns with a complex overlay, the format of the pad map doesn't make a whole lot of sense because the pad map csv maps the synth_tile to some real IO.  I have been using names like SYN0 to represent a synth pad that doesn't directly correspond to a real pad.

In an roi region it doesn't matter whether you use information directly from the pad map or from the synth_tiles, which is why it's not currently enabled. But, for overlay generation we would either have to reconsider the pad map csv format to handle synth IOs that don't directly correspond to a real IO and change the pin map csv generation to handle both synth ios and real ios at the same time, or just take the synth tiles as an input to create_ioplace.

This method was the easiest to get overlay generation working, but maybe the better solution is to figure out the correct format for the pinmap csv and merge together the synth tile to pinmap csv with the normal pinmap csv script.

Another method that could make this more consistent, is completely remove the pinmap csv generation for synth_tiles and only use the args.synth_tiles option in create_ioplace.

Thoughts?

andrewb1999:
> Overall looks really good, especially with the green CI. Missing one docstring, and I have some questions about the new code in `create_ioplace`, now wondering if that code can just be deleted?

Commented on the create_ioplace changes.  Which docstring is missing?

litghost:
> Which docstring is missing?

https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1565#discussion_r458921486

xc/common/utils/prjxray_create_ioplace.py
litghost: > But, for overlay generation we would either have to reconsider the pad map csv format to handle synth IOs that don't directly correspond to a real IO and change the pin map csv generation to handle both synth ios and real ios at the same time, or just take the synth tiles as an input to create_ioplace.

I'd rather have the relevant code to generate the correct pin map csv rather than the current code?  Pinmap CSV is simply a list of pads to VPR tile locations.  In the case where they are purely synthetic, the name is invented, but still has meaning.  Like eventually you'll have `CLK_MHZ` or `BUS_READY` or something, yes?

litghost: Something that is bothering me about this change is that it isn't being tested at all right now, because that code path is unused.  It feels weird to have dead code being added.

utils/lib/rr_graph/graph2.py
andrewb1999: Does this need a docstring?  Nothing else in this file has a docstring (including the create_pin_name_from_tile_type_num_and_pin this function is based on).

xc/common/utils/prjxray_create_ioplace.py
andrewb1999: > > But, for overlay generation we would either have to reconsider the pad map csv format to handle synth IOs that don't directly correspond to a real IO and change the pin map csv generation to handle both synth ios and real ios at the same time, or just take the synth tiles as an input to create_ioplace.
> 
> I'd rather have the relevant code to generate the correct pin map csv rather than the current code? Pinmap CSV is simply a list of pads to VPR tile locations. In the case where they are purely synthetic, the name is invented, but still has meaning. Like eventually you'll have `CLK_MHZ` or `BUS_READY` or something, yes?

Yes.  I guess what I mean by "the format of the pad map doesn't make sense" is what should go in the iob column for synth tiles that do not correspond to a real IO?

Current pinmap format:
```
name,x,y,z,is_clock,is_input,is_output,iob
A1,114,73,0,1,1,1,IOB_X1Y81
```

> Something that is bothering me about this change is that it isn't being tested at all right now, because that code path is unused. It feels weird to have dead code being added.

I agree completely now that I'm thinking about it.  This shouldn't be merged as is.  Maybe the correct thing to do it delete the unused code in this pr and include the merged pinmap generation fixes in the overlay generation pr?

utils/lib/rr_graph/graph2.py
andrewb1999: I can write a docstring easily, it just seems weird to only have one function in the file with a docstring.

xc/common/utils/prjxray_create_ioplace.py
litghost: > I agree completely now that I'm thinking about it. This shouldn't be merged as is. Maybe the correct thing to do it delete the unused code in this pr and include the merged pinmap generation fixes in the overlay generation pr?

That sounds good to me.  I think a focused PR on dealing with overlay generation issues might be more productive, as the other changes in this PR all otherwise look good and generalize well.

andrewb1999: > That sounds good to me. I think a focused PR on dealing with overlay generation issues might be more productive, as the other changes in this PR all otherwise look good and generalize well.

Now that I'm thinking about it more, this fix is necessary for heterogeneous synth tiles to work.  Currently the pinmap from synth tiles always assigns the z value to 0, so if there are multiple synth tiles at the same location the pinmap will be wrong.  It doesn't get tested currently because there is no test in here that actually uses multiple synth tiles at the same location yet (that plan was to add that with the overlay generation test).

litghost: Ok, given that, shouldn't we update the CMake to pass the synth tiles JSON?  Or fix the tool to create the pinmap CSV so that it works?

andrewb1999: > Ok, given that, shouldn't we update the CMake to pass the synth tiles JSON? Or fix the tool to create the pinmap CSV so that it works?

I think I should fix the tool to create the pinmap csv.  If I merge together the synth_tile_to_pinmap_csv and create_pinmap_csv, can I just put some arbitrary name in the iob column for synth_tiles whose pad name isn't in the real io pad list?  Should I add another column to the csv for whether it's an non-real-io-associated synth_tile?  I think the iob is only used for setting io standards, so I just need to not set those for the unassociated synth_tiles.

litghost: > can I just put some arbitrary name in the iob column for synth_tiles whose pad name isn't in the real io pad list

I'd just leave it blank

> Should I add another column to the csv for whether it's an non-real-io-associated synth_tile

I think this is prudent

xc/common/utils/prjxray_create_ioplace.py
```diff
@@ -9,6 +9,16 @@
 from lib.parse_pcf import parse_simple_pcf
 
 
+def get_synth_tile_from_pad(synth_tiles, pad):
```
litghost, line 12: This code is now dead?

andrewb1999: Yes, removing now.

andrewb1999:
@litghost passing everything except for an std::bad_alloc in VtR + Arch Defs

litghost:
> @litghost passing everything except for an std::bad_alloc in VtR + Arch Defs

Ya, not sure what is up with that.  I'm hitting the same failure in another PR.  Maybe an OOM situation on the CI?  Anyways, I'll review soon.

PR #1569, opened 2020-07-02
WIP: Changes for new yosys

acomodi:
This PR requires the new [yosys master+wip ](https://github.com/SymbiFlow/yosys/pull/81) to go green.

It adds a new techmap for the carrychain and adjusts some tests, as well as the synth.tcl script which needs some parameters to be added/changed.

### DIFF
--- a/xc/xc7/tests/simple_ff/CMakeLists.txt
+++ b/xc/xc7/tests/simple_ff/CMakeLists.txt
@@ -51,4 +51,6 @@ add_custom_target(
   DEPENDS ${OUT_FASM}
   )
 
-add_dependencies(all_xc7_tests simple_ff_full_loc_check)
+# XXX: need to fix attribute trimming when running with abc9
+# https://github.com/YosysHQ/yosys/issues/2176
+#add_dependencies(all_xc7_tests simple_ff_full_loc_check)
--- a/xc/xc7/tests/srl/srl16_amc31/basys3_dffmux_top.v
+++ b/xc/xc7/tests/srl/srl16_amc31/basys3_dffmux_top.v
@@ -12,7 +12,7 @@ input  wire [15:0]  sw,
 output wire [15:0]  led
 );
 
-parameter SRL_COUNT = 4;
+parameter SRL_COUNT = 8;
 parameter PRESCALER = 4;
 
 // UART loopback
--- a/xc/xc7/tests/srl/srl32_chain/basys3_top.v
+++ b/xc/xc7/tests/srl/srl32_chain/basys3_top.v
@@ -90,9 +90,6 @@ always @(posedge clk)
 
 // ============================================================================
 
-wire net_0;
-LUT2 #(.INIT(4'd0)) lut_0 (.I0(1'b0), .I1(|sw), .O(net_0));
-
 // LEDs
 genvar j;
 generate for(j=0; j<8; j=j+1) begin
@@ -100,8 +97,8 @@ generate for(j=0; j<8; j=j+1) begin
     assign led[j  ] = (sw[1]) ? error_lat[j] : error[j];
     assign led[j+8] = srl_q[j];
   end else begin
-    assign led[j  ] = net_0;
-    assign led[j+8] = net_0;
+    assign led[j  ] = |sw;
+    assign led[j+8] = |sw;
   end
 end endgenerate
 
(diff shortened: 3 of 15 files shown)


xc/xc7/tests/simple_ff/CMakeLists.txt
```diff
@@ -51,4 +51,6 @@ add_custom_target(
   DEPENDS ${OUT_FASM}
   )
 
-add_dependencies(all_xc7_tests simple_ff_full_loc_check)
+# XXX: need to fix attribute trimming when running with abc9
```
litghost, line 54: This isn't required anymore is it?  There was a fix upstream?

xc/xc7/tests/srl/srl16_amc31/basys3_dffmux_top.v
```diff
@@ -12,7 +12,7 @@ input  wire [15:0]  sw,
 output wire [15:0]  led
 );
 
-parameter SRL_COUNT = 4;
+parameter SRL_COUNT = 8;
```
litghost, line 15: Why is this change required?

xc/xc7/tests/srl/srl32_chain/basys3_top.v
```diff
@@ -90,18 +90,15 @@ always @(posedge clk)
...
 // LEDs
 genvar j;
 generate for(j=0; j<8; j=j+1) begin
   if (j < 4) begin
     assign led[j  ] = (sw[1]) ? error_lat[j] : error[j];
     assign led[j+8] = srl_q[j];
   end else begin
-    assign led[j  ] = net_0;
-    assign led[j+8] = net_0;
+    assign led[j  ] = |sw;
```
litghost, line 100: Why is this change required?  Are these SRL changes from debugging?

xc/xc7/yosys/synth.tcl
```diff
@@ -44,6 +44,9 @@ opt_clean
 setundef -zero -params
 stat
 
+# TODO: remove this as soon as new VTR master+wip is pushed: https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/525
```
litghost, line 47: Should have an issue to track this one.

xc/xc7/tests/simple_ff/CMakeLists.txt
acomodi: I need to double check this, but this should be a different issue w.r.t. the one that did not allow the BUFGCTRL or PLL to be LOCed. This issue seems to be related on the new code that handles FF for ABC in yosys.

xc/xc7/tests/srl/srl16_amc31/basys3_dffmux_top.v
acomodi: Apparently, this and the other changes to SRLs were required as VPR is not able to route GND to the top level outputs.

litghost: And masking that bug is a good solution because?

xc/xc7/tests/simple_ff/CMakeLists.txt
litghost: So we still need to fix yosys before we integrate it?  It still has outstanding issues that prevent it from being usable?

acomodi: It is currently usable for the regression tests within symbiflow-arch-defs, apart from this last one. I am in the process of verifying once again that the issue is still present.

xc/xc7/tests/srl/srl16_amc31/basys3_dffmux_top.v
acomodi: I'll file an issue about this, and debug it. One reason I have fixed the design was to get the all_xc7 tests to completion to check whether there were other blocking issues while using the new yosys.

xc/xc7/tests/simple_ff/CMakeLists.txt
acomodi: This issue is actually solved, the LOC constraint is present after synthesis. I will re-enable the test.

litghost: Awesome!

xc/xc7/tests/srl/srl16_amc31/basys3_dffmux_top.v
litghost: Both ninja and make have a "-k" to avoid stopping at failures for just this purpose.  We use this feature for the vendor tests CI for exactly this reason.

litghost:
Superseded #1577

PR #1570, opened 2020-07-02
xc7/toolchain_wrappers: find device/arch using part name

mglb:
Platforms and designs for xc7 devices in frameworks like Migen or Litex specify only part name, as it is enough for Vivado.

With this change, it becomes possible to add Symbiflow backend as a replacement for Vivado to such frameworks without requiring additional information (i.e. device and arch) from specific platforms and designs.

It is still possible to manually set the device and arch.


mithro:
@kgugala - Please merge if the CI goes green.

kgugala:
CI is red, but the fail is unrelated to this PR (https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1400). We can merge this

Issue #1429, opened 2020-04-18
Vendor Tools IO placement error


acomodi:
The following error shows up from time to time on the Vendor tools CI, regarding the baselitex test:

```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/soc/litex/base/baselitex_arty/artix7-xc7a50t-virt-xc7a50t-test && /usr/bin/cmake -E remove -f baselitex_arty/artix7-xc7a50t-virt-xc7a50t-test/design_baselitex_arty_vivado.dcp && /usr/bin/cmake -E remove -f baselitex_arty/artix7-xc7a50t-virt-xc7a50t-test/design_baselitex_arty_vivado.xpr && /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/prjxray/utils/vivado.sh -mode batch -source /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/soc/litex/base/baselitex_arty_vivado_runme.tcl > /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/soc/litex/base/baselitex_arty/artix7-xc7a50t-virt-xc7a50t-test/vivado.stdout.log
ERROR: [Place 30-294] The following IO terminals are loads of BUFIO/BUFR/BUFH
 and they should be placed to the same clock region as BUFIO/BUFR/BUFH instance.
 They have incompatible IO Standards that makes their placement illegal
 IO Standard1: SioStd: DIFF_SSTL135 VCCO = 1.35 Termination: 0  TermDir:  Out  BufioId: 2 Bank: 34 Placed LVDS 
     List of IO terminals :
	Term: ddram_dqs_p[0]
	Term: ddram_dqs_n[0]
	Term: ddram_dqs_p[1]
	Term: ddram_dqs_n[1]
	Term: ddram_clk_p
	Term: ddram_clk_n

 IO Standard2: SioStd: LVCMOS33   VCCO = 3.3 Termination: 0  TermDir:  Out  BufioId: 2 Bank: 15 Drv: 12 Placed 
     List of IO terminals :
	Term: eth_ref_clk

ERROR: [Place 30-374] IO placer failed to find a solution
Below is the partial placement that can be analyzed to see if any constraint modifications will make the IO placement problem easier to solve.

[...]

PR #1573, opened 2020-07-03
brams: fix VCC fan bounce issue

acomodi:

This PR should fix https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1400

### DIFF
--- a/xc/xc7/fasm2bels/bram_models.py
+++ b/xc/xc7/fasm2bels/bram_models.py
@@ -165,6 +165,9 @@ def clean_up_to_bram18(top, site):
         site.mask_sink(bel, 'ADDRATIEHIGH[{}]'.format(idx))
         site.mask_sink(bel, 'ADDRBTIEHIGH[{}]'.format(idx))
 
+    site.mask_sink(bel, 'REGCLKB')
+    site.mask_sink(bel, 'REGCLKARDRCLK')
+
     site.mask_sink(bel, 'WEA[1]')
     site.mask_sink(bel, 'WEA[3]')
     site.mask_sink(bel, 'WEBWE[1]')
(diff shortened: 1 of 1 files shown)


kgugala:
now it fails in `fasm2bels` with:

```  
File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/xc/xc7/fasm2bels/verilog_modeling.py", line 1685, in find_sources_from_sink
    wire_pkey = site.site_wire_to_wire_pkey[site_wire]
KeyError: 'REGCLKBL'
```

acomodi:
> now it fails in `fasm2bels` with:
> 
> ```
> File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/xc/xc7/fasm2bels/verilog_modeling.py", line 1685, in find_sources_from_sink
>     wire_pkey = site.site_wire_to_wire_pkey[site_wire]
> KeyError: 'REGCLKBL'
> ```

Fixed, CI should turn green now.

acomodi:
The FAN_BOUNCE issue seems to be solved, but now the https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1429 has reappeared.

acomodi:
The IO placement issue does not appear when using https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1545. This because the placement step is not performed.

litghost:
> The IO placement issue does not appear when using #1545. This because the placement step is not performed.

The problem still exists, in that it appears the clock placement script isn't taking into account BUFR/BUFIO/BUFMR <-> input as it should.

acomodi:
@litghost Agreed, the problem though is that we do not make use of any of those BUFIO/BUFR and BUFMRs. Unsure how the solution would look like.

On a side note, also the full Vivado flow does not implement those clock tiles.

PR #1581, opened 2020-07-08
Transition to use XDC constraints instead of TCL in fasm2bels

acomodi:
This PR is an initial work to have the XDC constraints, which will allow for faster fasm2bels step.

On an initial test (counter) everything seems to be working all right except for the CLK net constraints.

In fact, the XDC clock for a net constraints look like the following (this is generated by doing `write_xdc` on an implemented design from the old fasm2bels:

```
set_property FIXED_ROUTE { { CLK_HROW_CK_HCLK_OUT_L9 CLK_HROW_CK_BUFHCLK_L9 <8>HCLK_LEAF_CLK_B_TOPL3  { <7>GCLK_L_B9_WEST CLK_L1 CLBLL_LL_CLK }   { <4>GCLK_L_B9_WEST CLK_L1 CLBLL_LL_CLK }   { <8>GCLK_L_B9_WEST  CLK_L1 CLBLL_LL_CLK }   { <5>GCLK_L_B9_WEST CLK_L1 CLBLL_LL_CLK }   { <9>GCLK_L_B9_WEST CLK_L1 CLBLL_LL_CLK }   { <6>GCLK_L_B9_WEST CLK_L1 CLBLL_LL_CLK }  <3>GCLK_L_B9_WEST  { CLK_L1 CLBLL_LL_CLK }  CLK_L0      CLBLL_L_CLK }  } [get_nets CLK_HROW_BOT_R_X60Y26_BUFHCE_X0Y9_O]
```

The following node is the source of the issue: `<8>HCLK_LEAF_CLK_B_TOPL3` (as well as others nodes with the `<N>` prefix).
By inspecting the device view in Vivado, I came to the conclusion that the number reported in the prefix is related to the number of the HCLK tile from which the CLK net gets into the vertical clock lines.

### DIFF
--- a/xc/xc7/fasm2bels/make_routes.py
+++ b/xc/xc7/fasm2bels/make_routes.py
@@ -9,6 +9,25 @@ ONE_NET = -2
 
 DEBUG = False
 
+# The following constants are used to wrap lists so that 1-length
+# lists do not get trimmed away producing errors during parsing
+#
+# Error example:
+#   - Input: [list a b c [list d e] [list f] [list g h]]
+#   - Output: a b c {d e} f {g h}
+#
+#   In this example, the `f` list gets evaluated into a string, which is not
+#   the expected result
+#
+# Correct example:
+#   - Input: [list a b c " [list d e] " " [list f] " " [list g h] " ]]
+#   - Output: a b c {d e} {f} {g h}
+#
+#   The correct example has the `f` list correctly evaluated into an
+#   actual list
+TCL_LIST_OPEN = '"[list '
+TCL_LIST_CLOSE = '] "'
+
 
 def create_check_downstream_default(conn, db):
     """ Returns check_for_default function. """
@@ -90,13 +109,8 @@ def find_downstream_node(conn, check_downstream_default, source_node_pkey):
     return None
 
 
-def output_builder(fixed_route):
-    # TCL cannot express 1-length list, so the curly brackets are
-    # used to prevent TCL from collapsing the 1-length list.
-    if len(fixed_route) == 1:
-        yield '{ '
-    else:
-        yield '[list '
+def output_builder(fixed_route, first_run=False):
+    yield TCL_LIST_OPEN
 
     for i in fixed_route:
         if type(i) is list:
@@ -105,10 +119,7 @@ def output_builder(fixed_route):
         else:
             yield i
 
-    if len(fixed_route) == 1:
-        yield ' }'
-    else:
-        yield ' ]'
+    yield TCL_LIST_CLOSE
 
 
 class Net(object):
@@ -309,12 +320,7 @@ class Net(object):
                 if parent_node == self.source_wire_pkey:
                     source_nodes.append(node)
 
-            # TCL cannot express 1-length list, so the curly brackets are
-            # used to prevent TCL from collapsing the 1-length list.
-            if len(source_nodes) == 1:
-                yield '{ '
-            else:
-                yield '[list '
+            yield TCL_LIST_OPEN
 
             for source_node in source_nodes:
                 yield '('
@@ -325,10 +331,7 @@ class Net(object):
 
                 yield ')'
 
-            if len(source_nodes) == 1:
-                yield ' }'
-            else:
-                yield ' ]'
+            yield TCL_LIST_CLOSE
 
 
 def create_check_for_default(db, conn):
--- a/xc/xc7/fasm2bels/make_routes.py
+++ b/xc/xc7/fasm2bels/make_routes.py
@@ -100,9 +100,7 @@ def output_builder(fixed_route):
         else:
             yield i
 
-    # TCL cannot express 1-length list, so add an additional element to
-    # prevent TCL from collapsing the 1-length list.
-    yield ' {} ]'
+    yield ']'
 
 
 class Net(object):
--- a/xc/xc7/fasm2bels/make_routes.py
+++ b/xc/xc7/fasm2bels/make_routes.py
@@ -91,7 +91,12 @@ def find_downstream_node(conn, check_downstream_default, source_node_pkey):
 
 
 def output_builder(fixed_route):
-    yield '[list'
+    # TCL cannot express 1-length list, so the curly brackets are
+    # used to prevent TCL from collapsing the 1-length list.
+    if len(fixed_route) == 1:
+        yield '{ '
+    else:
+        yield '[list '
 
     for i in fixed_route:
         if type(i) is list:
@@ -100,7 +105,10 @@ def output_builder(fixed_route):
         else:
             yield i
 
-    yield ']'
+    if len(fixed_route) == 1:
+        yield ' }'
+    else:
+        yield ' ]'
 
 
 class Net(object):
@@ -301,7 +309,12 @@ class Net(object):
                 if parent_node == self.source_wire_pkey:
                     source_nodes.append(node)
 
-            yield '[list '
+            # TCL cannot express 1-length list, so the curly brackets are
+            # used to prevent TCL from collapsing the 1-length list.
+            if len(source_nodes) == 1:
+                yield '{ '
+            else:
+                yield '[list '
 
             for source_node in source_nodes:
                 yield '('
@@ -312,9 +325,10 @@ class Net(object):
 
                 yield ')'
 
-            # TCL cannot express 1-length list, so add an additional element
-            # to prevent TCL from collapsing the 1-length list.
-            yield '{} ]'
+            if len(source_nodes) == 1:
+                yield ' }'
+            else:
+                yield ' ]'
 
 
 def create_check_for_default(db, conn):
--- a/xc/xc7/fasm2bels/verilog_modeling.py
+++ b/xc/xc7/fasm2bels/verilog_modeling.py
@@ -1578,18 +1578,21 @@ class Module(object):
     def output_bel_locations(self):
         """ Yields lines of tcl that will assign set the location of BELs. """
         for bel in sorted(self.get_bels(), key=lambda bel: bel.priority):
-            yield """\
-set cell [get_cells *{cell}]""".format(cell=bel.get_prefixed_name())
+            get_cell = "[get_cells *{cell}]".format(
+                cell=bel.get_prefixed_name()
+            )
 
             if bel.bel is not None:
                 yield """\
-set_property BEL "[get_property SITE_TYPE [get_sites {site}]].{bel}" $cell""".format(
-                    site=bel.site,
+set_property BEL {bel} {get_cell}""".format(
                     bel=bel.bel,
+                    get_cell=get_cell,
                 )
 
             yield """\
-set_property LOC [get_sites {site}] $cell""".format(site=bel.site)
+set_property LOC {site} {get_cell}""".format(
+                site=bel.site, get_cell=get_cell
+            )
 
     def output_nets(self):
         """ Yields lines of tcl that will assign the exact routing path for nets.
@@ -1629,7 +1632,9 @@ set net [get_nets -of_object $pin]""".format(
                 assert net_wire_pkey in [ZERO_NET, ONE_NET]
                 continue
 
-            yield """set route {fixed_route}""".format(fixed_route=' '.join(fixed_route))
+            yield """set route {fixed_route}""".format(
+                fixed_route=' '.join(fixed_route)
+            )
 
             # Remove extra {} elements required to construct 1-length lists.
             yield """set_property FIXED_ROUTE $route $net"""
--- a/xc/xc7/fasm2bels/verilog_modeling.py
+++ b/xc/xc7/fasm2bels/verilog_modeling.py
@@ -1579,10 +1579,7 @@ class Module(object):
         """ Yields lines of tcl that will assign set the location of BELs. """
         for bel in sorted(self.get_bels(), key=lambda bel: bel.priority):
             yield """\
-set cell [get_cells *{cell}]
-if {{ $cell == {{}} }} {{
-    error "Failed to find cell!"
-}}""".format(cell=bel.get_prefixed_name())
+set cell [get_cells *{cell}]""".format(cell=bel.get_prefixed_name())
 
             if bel.bel is not None:
                 yield """\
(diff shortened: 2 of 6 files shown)


litghost:
> `<8>HCLK_LEAF_CLK_B_TOPL3` (

The `<8>` is used to say "the n'th instance, in distance from the pip", however it is sufficient to explicitly list the node, which is how fasm2bels does it.  There is some TCL logic for dealing with lists of 1 element, but I believe I know how to deal with that now.

I'll make another comment in a bit.

xc/xc7/fasm2bels/make_routes.py
```diff
@@ -102,7 +102,7 @@ def output_builder(fixed_route):
 
     # TCL cannot express 1-length list, so add an additional element to
     # prevent TCL from collapsing the 1-length list.
```
litghost, line 104: The above comment is important and needs to be handled.  I'll make a comment about it in a minute.

xc/xc7/fasm2bels/verilog_modeling.py
```diff
@@ -1604,9 +1600,9 @@ def output_nets(self):
 
         for net_wire_pkey, net in self.nets.items():
             if net_wire_pkey == ZERO_NET:
-                yield 'set net [get_nets {<const0>}]'
```
litghost, line 1607: You broke the constant net logic, this will not work.

litghost:
> `<8>HCLK_LEAF_CLK_B_TOPL3` (

The `<8>` is used to say "the n'th instance, in distance from the pip", however it is sufficient to explicitly list the node, which is how fasm2bels does it.  There is some TCL logic for dealing with lists of 1 element, but I believe I know how to deal with that now.

I'll make another comment in a bit.

xc/xc7/fasm2bels/verilog_modeling.py
acomodi: Fixed

acomodi:
@litghost I was able to correctly run fasm2bels for some of the most complex designs in the regression tests and they correctly passed using the XDC constraints and the latest changes.

I fixed the 1-length list issue to transition to use the `{ element }` format instead of `[list element]` format. This avoids the usage of the extra `{}` empty element which was causing issues when parsing the XDC.

The baselitex design now takes ~3m20s to parse the XDC, versus the 10m15s of sourcing the TCL from here https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1545#issuecomment-648329620

litghost:
> I fixed the 1-length list issue to transition to use the `{ element }` format instead of `[list element]` format. This avoids the usage of the extra `{}` empty element which was causing issues when parsing the XDC

Great

xc/xc7/fasm2bels/make_routes.py
```diff
@@ -303,7 +309,12 @@ def descend_fixed_route(source_node_pkey, fixed_route):
                 if parent_node == self.source_wire_pkey:
                     source_nodes.append(node)
 
-            yield '[list '
+            # TCL cannot express 1-length list, so the curly brackets are
+            # used to prevent TCL from collapsing the 1-length list.
+            if len(source_nodes) == 1:
+                yield '{ '
```
litghost, line 315: Let's just use `'{ '`, I expect it work in all cases.

xc/xc7/fasm2bels/verilog_modeling.py
```diff
@@ -1639,14 +1632,12 @@ def output_nets(self):
...
-""".format(fixed_route=' '.join(fixed_route))
+            yield """set route {fixed_route}""".format(
+                fixed_route=' '.join(fixed_route)
+            )
 
             # Remove extra {} elements required to construct 1-length lists.
-            yield """\
-regsub -all {{}} $route_with_dummy "" route
-set_property FIXED_ROUTE $route $net"""
+            yield """set_property FIXED_ROUTE $route $net"""
```
litghost, line 1640: This works in xdc parser?  That's really great!

litghost:
> The baselitex design now takes ~3m20s to parse the XDC, versus the 10m15s of sourcing the TCL from here

Did the `diff_fasm` test pass in this case?

litghost:
Something I'm worried about is error behavior.  Part of the changes you made remove the error checking.  What is Vivado's behavior if a `get_nets` or `get_cells` call returns no matches?  Is an error generated?  Is it debuggable?

xc/xc7/fasm2bels/make_routes.py
acomodi: I actually tried this, but I got errors such as the following:

```
CRITICAL WARNING: [Common 17-178] Regular expression error 'Unmatched [ or [^ in character class declaration.  The error occurred while parsing the regular expression fragment: 'GET_NODES$>>>HERE>>>'.' for pattern '[GET_NODES' [/data/symbiflow/symbiflow-arch-defs/build/xc/xc7/tests/soc/litex/mini/minilitex_arty/artix7-xc7a50t-virt-xc7a50t-test/top_bit.v.xdc:9490] 
```

Apparently the `[get_nodes ...]` expression is not compatible with the curly brackets style.

acomodi:
> Did the diff_fasm test pass in this case?

Yes, diff_fasm produced no error

xc/xc7/fasm2bels/make_routes.py
```diff
@@ -314,9 +325,10 @@ def descend_fixed_route(source_node_pkey, fixed_route):
 
                 yield ')'
 
-            # TCL cannot express 1-length list, so add an additional element
-            # to prevent TCL from collapsing the 1-length list.
-            yield '{} ]'
+            if len(source_nodes) == 1:
+                yield ' }'
+            else:
+                yield '{} ]'
```
litghost, line 331: I think we can change `'{} ]'` to `]` because the 1-length case is handled differently

acomodi:
> Something I'm worried about is error behavior. Part of the changes you made remove the error checking. What is Vivado's behavior if a get_nets or get_cells call returns no matches? Is an error generated? Is it debuggable?

I have noticed we cannot have if/else statements in the XDC constraints, therefore the error checking needs to be performed differently.

I believe that there could be some way to enable a `strict` mode that would raise warnings to errors. I need to check this.

xc/xc7/fasm2bels/make_routes.py
litghost: Weird :/  Well if this works, great.  Might be worth adding a comment summarizing what you just told me.  It's quiet surprising to me what is and is not working the XDC file, so capturing what doesn't work with a note is useful to future readers of the code.

litghost:
> > Something I'm worried about is error behavior. Part of the changes you made remove the error checking. What is Vivado's behavior if a get_nets or get_cells call returns no matches? Is an error generated? Is it debuggable?
> 
> I have noticed we cannot have if/else statements in the XDC constraints, therefore the error checking needs to be performed differently.
> 
> I believe that there could be some way to enable a `strict` mode that would raise warnings to errors. I need to check this.

I think the first thing we should do is intentionally test what happens if get_nets or get_cells fails in the XDC method.  It might already generate an error, in which case we don't need the if statements anyways.  I just knew that get_nets and get_cells only warned when used in the TCL environment.  Possible that the XDC environment has different rules that remove the need to for the error checking?

xc/xc7/fasm2bels/make_routes.py
acomodi: Sure, fixed

xc/xc7/fasm2bels/make_routes.py
litghost: We should revisit this.  I believe we should be able to get everything to work with just using `{` and `}`.  I'm fine merging this as is, but coming back and trying to clean this up is probably worthwhile.

xc/xc7/fasm2bels/verilog_modeling.py
acomodi: Indeed, what happens though is the following, from what I understood:
- the $route and $net variables are being evaluated (they include some tcl getters such as get_node and such)
- the evaluation result of those variables is substituted into the set_property command
- the XDC is being overwritten with a "standard" XDC which is the resulting one after the parsing pass.

In other words:
```
set pin [get_pins *CLBLM_R_X25Y92_SLICE_X38Y92_RAM64X1D_CD/DPO]
set net [get_nets -of_object $pin]
set route [list  [get_nodes -of_object [get_wires CLBLM_R_X25Y92/CLBLM_M_C]] CLBLM_R_X25Y92/CLBLM_M_CMUX CLBLM_R_X25Y92/CLBLM_LOGIC_OUTS22 INT_R_X25Y92/NL1BEG_N3 INT_R_X25Y92/FAN_ALT1 INT_R_X25Y92/FAN_BOUNCE1 INT_R_X25Y92/BYP_ALT4 INT_R_X25Y92/BYP4 CLBLM_R_X25Y92/CLBLM_M_BX  ]
set_property FIXED_ROUTE $route $net
```

The above original XDC (that comes out of fasm2bels) gets overwritten wth this:
```
set_property FIXED_ROUTE { { CLBLM_M_C CLBLM_M_CMUX CLBLM_LOGIC_OUTS22 NL1BEG_N3 FAN_ALT1 FAN_BOUNCE1 BYP_ALT4 BYP4 CLBLM_M_BX }  } [get_nets CLBLM_R_X25Y95_SLICE_X38Y95_RAM64X1D_CD/DPO]
```

litghost:
> The baselitex design now takes ~3m20s to parse the XDC, versus the 10m15s

This is a great runtime improvement, thanks for doing this!

xc/xc7/fasm2bels/verilog_modeling.py
litghost: Okay, your example shows why using `{}` all the time won't work, and it shows a weakness in this new approach.  I don't have a solution, but I can talk through what is happening.

```
set route [list  [get_nodes -of_object [get_wires CLBLM_R_X25Y92/CLBLM_M_C]] CLBLM_R_X25Y92/CLBLM_M_CMUX CLBLM_R_X25Y92/CLBLM_LOGIC_OUTS22 INT_R_X25Y92/NL1BEG_N3 INT_R_X25Y92/FAN_ALT1 INT_R_X25Y92/FAN_BOUNCE1 INT_R_X25Y92/BYP_ALT4 INT_R_X25Y92/BYP4 CLBLM_R_X25Y92/CLBLM_M_BX  ]
```

In the example above, `[get_nodes -of_object [get_wires CLBLM_R_X25Y92/CLBLM_M_C]]` needs to be evaluated to become `CLBLM_R_X25Y92/CLBLM_M_C`, which Vivado reduces to `CLBLM_M_C` in the output XDC.   In TCL when you use `{}` it supresses `[]` expansion.

If you run the following commands in TCL you can see this in action.

Run:
```
set l1 [list [get_nodes -of_object [get_wires CLBLM_R_X25Y92/CLBLM_M_C]] CLBLM_R_X25Y92/CLBLM_M_CMUX]
set l2 { [get_nodes -of_object [get_wires CLBLM_R_X25Y92/CLBLM_M_C]] CLBLM_R_X25Y92/CLBLM_M_CMUX }
set l3 { CLBLM_R_X25Y92/CLBLM_M_C CLBLM_R_X25Y92/CLBLM_M_CMUX }
puts [expr $l1 == $l2]
puts [expr $l1 == $l3]
```

xc/xc7/fasm2bels/make_routes.py
litghost: From your example below I believe I know how to remove the if statement, and avoid breaking things.

As discussed before, using `{}` causes `[]` expansions to not be done.  This is why when you tried to use `{}` for all lists it didn't work.  But that problem applies to the `len(sources_nodes) == 1` case too!  You just haven't tested a design that hits the problem yet.  I have a solution.

The problem that I was original trying to solve is how to express the following list:

```
a b c { d e } { CLBLM_R_X25Y92/CLBLM_M_C } { g h }
```

Where `CLBLM_R_X25Y92/CLBLM_M_C` is the return value of a `[]` expansion (like get_nodes).

The original problem arises because the behavior of `[list ]` when you send a 1 element list into a list.  This problem can be clearly seen in the following example:
```
set l [list a b c [list d e] [list f] [list g h]]
```
In which case, l becomes `a b c {d e} f {g h}` instead of `a b c {d e} { f } {g h}`.  One solution is a follows:
```
set l [list a b c [format " %s " [list d e]] [format " %s " [list f]] [format " %s " [list g h]]]
```

Which results in: `a b c { d e } { f } { g h }`.

So the real solution is to replace `[list ` with `[format " %s " [list ` and `{} ]` with `]]`.

litghost: I recommend making to global constants TCL_LIST_OPEN and TCL_LIST_CLOSE, which are 
```
TCL_LIST_OPEN = '[format " %s " [list '
TCL_LIST_CLOSE = ']]'
```

and a block comment explaining why `[list ` and `]` doesn't work as expected.

acomodi: Awesome, tested this solution and it works all right. Hopefully all the tests will not show issues.

I have added also a comment on top of the global constants.

acomodi:
> Possible that the XDC environment has different rules that remove the need to for the error checking?

@litghost I have checked this. The XDC parser returns a Warning for the `not found` cell and a CRITICAL WARNING when applying the constraint being the cell empty.
By raising the messages from CRITICAL WARNING to ERROR does not make Vivado to kill the process and return the error, but it continues until the end of the tcl.
We get an error of course on the diff fasm as the routes and bels used are different from the VPR bitstream.

acomodi:
@litghost Got some numbers from the master CI runs and the ones in this PR:

From https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1563:
- Vendor Tools XC7: 5h37m

This PR:
- Vendor Tools XC7: 4h56m

Maybe not a great win from a CI point of view, but definitely worth when running single tests, as makes iterations faster.

litghost:
> @litghost Got some numbers from the master CI runs and the ones in this PR:
> 
> From #1563:
> 
> * Vendor Tools XC7: 5h37m
> 
> This PR:
> 
> * Vendor Tools XC7: 4h56m
> 
> Maybe not a great win from a CI point of view, but definitely worth when running single tests, as makes iterations faster.

The next step to drop the runtime is likely working on fasm2bels runtime.  For example minilitex_ddr_arty100t:
 - Time in fasm2bels: 351.033 seconds
 - Time in Vivado: 615.667 seconds

I'm not suggesting that we profile and optimize fasm2bels right now, but to get further runtime improvements will require taking a holistic view for each test case.  Also keep in mind as we approach the xc7 only runtime (2 hr, 35 min, 47 sec) we will see diminishing returns as we focus on the Vivado portion of the tests.

PR #1590, opened 2020-07-14
[WIP/DNM] Add earlgray test design

rw1nkler:
This PR adds the OpenTitan earlgray test to SymbiFlow.


rw1nkler:
The newer version of the example, which uses sv2v, has been created in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1653

Issue #1450, opened 2020-04-29
Add kokoro build badge to the README file


mithro:
Kokoro publishes a build bage. We should include it in the README.

PR #1600, opened 2020-07-15
Fix the build badges.

mithro:
All the badges should be accessible in the symbiflow-badges GCS bucket.

*Really* fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1450


mithro:
Screenshot from 2020-07-15 15-55-29

The "Xilinx XC7A200T - Vendor Tool Tests" hasn't finished running yet. Once that happens the broken image should be fixed.

The HTML page which the build badges currently link too will redirect to the external build results once the build updates.

Issue #1597, opened 2020-07-15
Carry Chain COUT drives multiple net sinks


acomodi:
When updating the baslitex test with the newest version of Litex, I have encountered the following assertion failure in VPR:

```
vpr/src/place/place_macro.cpp:135 find_all_the_macro: Assertion 'cluster_ctx.clb_nlist.net_sinks(curr_net_id).size() == 1' failed.
```

The assertion checks whether nets between macros (in this case carry chains) have more than one sink. This because a net connecting to blocks in a macro should not drive other sinks, as this would create inconsistencies with the macro placement.

I have tracked down the possible cause of the issue being in the fact that COUT pins from carry chain get to be connected to external nets rather than just the CIN pin of a next carry in the chain (if no other carry is present in the chain, the COUT pin should not be connected to any other sink).

eblif:
```
[...]

PR #1605, opened 2020-07-20
DNM/WIP: changes to trigger carry chain COUT issue

acomodi:
This PR is intended to have a test that triggers https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1597


acomodi:
The same error as in the issue #1597 appeared on CI

Issue #1608, opened 2020-07-21
Add an install target for the 200t


HackerFoo:
Because generating the architecture files takes a long time and generates a lot of data, it is useful to separate this from the other architectures.

Longer term, it would make sense to allow installing specific architectures.

litghost:
I believe this was fixed, so this issue can be closed?

PR #1609, opened 2020-07-21
Add a CMake variable to select an architecture to install.

HackerFoo:
This is to allow installing the `xc7a200t` architecture separately. #1608


HackerFoo:
> `INSTALL_DEVICE` should probably be a CMake cache variable?

Does this apply if I don't want a default value? When `INSTALL_DEVICE` is not specified, it falls back to the previous behavior, installing every architecture except `xc7a200t`.

Is this just a way of attaching documentation for the variable?

litghost:
> Is this just a way of attaching documentation for the variable?

Yes, and it is also a discoverability issue.  CMake cache variables are exposed in a way that is discoverable, either in CMakeCache or the various cmake interfaces.

As for mechanically what to set `INSTALL_DEVICE` as a default, there are several reasonable defaults to choose from.  Just using the string "default", and adding a comment about the default install behavior would be sufficent.

CMakeLists.txt
```diff
@@ -27,6 +27,8 @@ add_custom_target(all_ice40_tests)
 set(ENV_DIR ${symbiflow-arch-defs_SOURCE_DIR}/env/conda/envs/symbiflow_arch_def_base/ CACHE PATH "Path to environment")
 set(YOSYS_DATADIR ${ENV_DIR}/share/yosys CACHE PATH "Path to yosys data directory")
 set(VPR_CAPNP_SCHEMA_DIR ${ENV_DIR}/capnp CACHE PATH "Path to VPR schema directory")
+set(INSTALL_DEVICE "small_devices" CACHE STRING "The device or device set to install")
```
litghost, line 30: I assume you meant for one of those `device` to be `devices`?  It's probably worth adding a comment that `xc7a200t` is specifically not included in "small_devices"?

HackerFoo: I meant "device set", but maybe "set of devices" is clearer? I thought about changing `INSTALL_DEVICE` to `INSTALL_DEVICE_SET`, but `small_devices` is the only device set and will probably be removed at some point.

HackerFoo: I changed to the comment to "The device to install, or 'small_devices', which includes all devices except xc7a200t."

PR #1612, opened 2020-07-22
WIP: Change how CARRY4 modelling is handled

### LINKED ISSUE #1597: Carry Chain COUT drives multiple net sinks

litghost:
This PR attempts to fix #1597 by change the strategy used to decongest SLICEL CARRY4 outputs.  Previously, all ALU CO uses were modeled as a `O ^ S`, but yosys's abc9 pass is now good enough to undo this change.  So rather than continue to try to fight abc9 as it gets better at merging logic, this PR uses a new CARRY4 congestion avoidance strategy.

Changes:
 - Yosys arith mapping using upstream CARRY4 mapping
 - New pass (implemented as a Python3 script) de-congests O + CO carry rows by using a LUT equation from the inputs
 - CARRY4_VPR is modelled as a blackbox instead of a whitebox to prevent abc9 from modifying/swapping outputs during cleanup optimization


litghost:
There is a bug, working on it.

litghost:
I still need to add a test that the CO LUT explosion calculates correctly.

xc/xc7/techmap/fix_carry.py
```diff
@@ -0,0 +1,230 @@
+import json
```
acomodi, line 1: Docstrings missing on several functions

xc/xc7/yosys/synth.tcl
```diff
@@ -33,7 +33,38 @@ if { [info exists ::env(INPUT_XDC_FILE)] && $::env(INPUT_XDC_FILE) != "" } {
...
+techmap -map  $::env(TECHMAP_PATH)/carry_map.v
+write_json $::env(OUT_JSON).carry_fixup.json
+exec $::env(PYTHON3) $::env(TECHMAP_PATH)/fix_carry.py < $::env(OUT_JSON).carry_fixup.json > $::env(OUT_JSON).carry_fixup_out.json
+design -push
+read_json $::env(OUT_JSON).carry_fixup_out.json
+
+techmap -map  $::env(TECHMAP_PATH)/clean_carry_map.v
+
+# Re-read baseline libraries
+read_verilog -lib -specify -D_EXPLICIT_CARRY +/xilinx/cells_sim.v
```
acomodi, line 48: Why is EXPLICIT_CARRY needed here if we are transitioning to have the native CARRY map from yosys?

xc/xc7/yosys/synth.tcl
```diff
@@ -33,7 +33,38 @@ if { [info exists ::env(INPUT_XDC_FILE)] && $::env(INPUT_XDC_FILE) != "" } {
...
+}
+
+# Re-run optimization flow to absorb carry modifications
+hierarchy -check -auto-top
+
+write_ilang $::env(OUT_JSON).pre_abc9.ilang
+if { $::env(USE_ROI) == "TRUE" } {
+    synth_xilinx -flatten -abc9 -nosrl -noclkbuf -nodsp -noiopad -nowidelut -run map_ffs:check
+} else {
+    synth_xilinx -flatten -abc9 -nosrl -noclkbuf -nodsp -iopad -nowidelut -run map_ffs:check
```
acomodi, line 62: How does this re-synthesis step impact on run-time?

xc/xc7/yosys/synth.tcl
```diff
@@ -33,7 +33,38 @@ if { [info exists ::env(INPUT_XDC_FILE)] && $::env(INPUT_XDC_FILE) != "" } {
...
+}
+
+write_ilang $::env(OUT_JSON).post_abc9.ilang
+
+chtype -map CARRY4_VPR CARRY4_FIX
```
acomodi, line 67: Why the cell type needs to be changed to CARRY4_FIX here?

xc/xc7/techmap/cells_sim.v
```diff
@@ -1911,3 +1979,12 @@ module PS7_VPR (
 
 endmodule
 
+module BANK();
```
acomodi, line 1982: What is this module related to and why is it needed?

xc/xc7/techmap/iobs.v
```diff
@@ -1,4 +1,5 @@
 module IBUF (
+  (* iopad_external_pin *)
```
acomodi, line 2: Are these changes to IOBs intended? They shouldn't be related to the CARRY_CHAIN changes.

I also see a failure in CI related to IOBUFs:
```
ERROR: Module `IBUF' referenced in module `top' in cell `$iopadmap$top.cpu_reset' does not have a parameter named 'IO_LOC_PAIRS'.
```

Could the error be related to these changes?

xc/xc7/yosys/synth.tcl
litghost: It's a mistake

xc/xc7/yosys/synth.tcl
litghost: I haven't measured it yet.

xc/xc7/yosys/synth.tcl
litghost: I'll add a comment

xc/xc7/techmap/cells_sim.v
litghost: The `check` after re-reading the JSON notices that the "read_xdc" plugin added the "BANK" module, which does exists, and yosys complains that it has parameters that do not exist.

xc/xc7/techmap/iobs.v
litghost: These changes are required to avoid problems with the iopadmap double inserting.  @mkurc-ant made similiar changes in his IDDR/ODDR PR I believe.  These fixes are unrelated to the carry chain stuff, except in that they are needed bug fixes for other reasons.

litghost: I understand the issue with the `IO_LOC_PAIRS`, working on a fix.

xc/xc7/yosys/synth.tcl
litghost: Fixed

acomodi:
@litghost I have pushed an updated version of litex with the fixes to the clock which now is correctly set to 60MHz as well as the changes in the mem.init files. The commit is this one: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1605/commits/aa2fb936db257db07aecc57c81d6df5d478dfff1

litghost:
> @litghost I have pushed an updated version of litex with the fixes to the clock which now is correctly set to 60MHz as well as the changes in the mem.init files. The commit is this one: [aa2fb93](https://github.com/SymbiFlow/symbiflow-arch-defs/commit/aa2fb936db257db07aecc57c81d6df5d478dfff1)

Great, I'll ingest those changes.

xc/xc7/techmap/fix_carry.py
litghost: Yep, fixed now

litghost:
I believe all review comments have been addressed other than the re-abc9 cost.

litghost:
> One other thing we would also need to verify is the QoR (as well as runtime) reached with these changes. I believe we can just inspect the qor.csv file generated as an artifact from CI.

Agreed.  Once it's green I'll post the results.

litghost:
The new install failure is now fixed.  The diff fasm failure is understood, but no solution is immediately obvious.  More concerning is that the place and route time for designs goes up with this change.  Investigating potentially solutions.

litghost:
A fasm2bels bug showed up in the CI, fix is here: https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/15

litghost:
Superceeded by #1674

PR #1613, opened 2020-07-22
Switch to cached database for routing import

andrewb1999:

On our hard drive based storage system, this change decreases the runtime of the import graph nodes step of routing import from over an hour to under 2 minutes.

### DIFF
--- a/xc/common/utils/prjxray_routing_import.py
+++ b/xc/common/utils/prjxray_routing_import.py
@@ -40,6 +40,7 @@ import functools
 import pickle
 
 import sqlite3
+from prjxray_db_cache import DatabaseCache
 
 now = datetime.datetime.now
 
@@ -1323,8 +1324,7 @@ def main():
     if synth_tiles is None:
         synth_tiles = find_constant_network(graph)
 
-    with sqlite3.connect("file:{}?mode=ro".format(args.connection_database),
-                         uri=True) as conn:
+    with DatabaseCache(args.connection_database, read_only=True) as conn:
 
         populate_bufg_rebuf_map(conn)
 


litghost:
The flip side is this increases the RAM usage dramatically.  How's your runtime with the "create_edges" step right now?  "create_edges" doesn't use a database cache either.

Rather than bringing all of the database into memory, we should identify which query's are problematic and fix them.  [`py-spy`](https://github.com/benfred/py-spy) uses a sampling methodology that should quickly be able to identify where wall-clock time is being spent.

litghost:
In general, the access patterns for the routing import step should be:
 - index lookups that SQLite should be caching already
 - Row scans that should be disk lookahead friendly

andrewb1999:
> In general, the access patterns for the routing import step should be:
> 
>     * index lookups that SQLite should be caching already
> 
>     * Row scans that should be disk read cache friendly
> 
> 
> If there are access patterns that don't match this, we should fix them, as I'd expect they will improve routing import time on SSD based machines as well.

https://github.com/andrewb1999/summer-of-code-resources/blob/master/profile.svg

Here is an example profile from py-spy for the import graph nodes step.  You can see that most of the time is spent on line 652, 673, and calls to update_connection_box.

Any idea why these queries would be so taxing in specific?  I assume this means they do not match the access pattern described, but I'm unsure how to fix it.

litghost:
Those all look good.  I believe the issue is that the SQLite cache is too small and results in evicting the b-trees on use, rather than holding them in memory.  The b-trees shouldn't be too big, but the default cache size is only 2 MB, which they will likely thrash.

litghost:
I just checked the 3 indicies that get hit by the queries you listed, and they total about ~120 MB of b-trees.  I believe if you try setting the sqlite cache to 256 MB, the runtime should improve dramatically, with a bounded increase in memory usage:

```
PRAGMA main.cache_size = -256000;
```

andrewb1999:
> I just checked the 3 indicies that get hit by the queries you listed, and they total about ~120 MB of b-trees. I believe if you try setting the sqlite cache to 256 MB, the runtime should improve dramatically, with a bounded increase in memory usage:
> 
> ```
> PRAGMA main.cache_size = -256000;
> ```

I added a line in the python `cur.execute('PRAGMA cache_size = -256000;')` right after opening the sqlite db and did not observe a noticeable runtime improvement.

litghost:
Which graph is this on?

andrewb1999:
> Which graph is this on?

arty-swbut

PR #1615, opened 2020-07-24
Update to version of fasm2bels with runtime improvements.

litghost:
Expected speedup of ~2-5 minutes per target on A100T and A200T, and
larger designs.


litghost:
Looks like the runtime for the CI jumped from 4:05 to 4:55, so I expect there is some small cases that are more expensive with the big map.  I'll think about what to do about that.  On larger designs, I expect the win is bigger, but the smaller designs the complexity hurts the runtime.

litghost:
For reference, baselitex_arty100:
Before: 466.162 seconds
After: 224.993 seconds

I'll look to see if I can figure out where the extra time went

litghost:
I've checked all of the bit_v times, and all of the times dropped.  I'm trying to determine where the runtime jump went :/

litghost:
> I've checked all of the bit_v times, and all of the times dropped. I'm trying to determine where the runtime jump went :/

Looks like the machine might've been slower than the previous baseline run.  Let's see how the next run goes.  However I have the data to support that the new fasm2bels is significantly faster, so this should be merged if no failures crop up.

litghost:
> I have noticed that the CI builds on master achieve a lower run-time w.r.t. to the PR CI builds, probably that is a coincidence.

Ya, I mentioned that in the comments above.  I checked the runtime data and the fasm2bels time is consistently lower, even with a higher overall time

Issue #1625, opened 2020-07-30
Feature DLUT.RAM not emitted


mkurc-ant:
This issue emerged in the PR https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1502 (SHA `c6a66c0`) on Murax built for A200T (test `tests/9-soc/murax`, target `murax_nexys_video_100_bit_v`).

While decoding the bitstream fasm2bels throws an assertion error meaning that there are LUT RAMs in a SLICEM enabled, but none of them is packed into DLUT hence the `DLUT.RAM` feature is not set and presumably it should be:
```
Traceback (most recent call last):
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/env/conda/envs/symbiflow_arch_def_base/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/env/conda/envs/symbiflow_arch_def_base/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/__main__.py", line 4, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/fasm2bels.py", line 416, in main
    process_tile(top, tile, tile_features)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/fasm2bels.py", line 103, in process_tile
    PROCESS_TILE[tile_type](top.conn, top, tile, tile_features)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/models/clb_models.py", line 1218, in process_clb
    process_slice(top, slices[s])
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7a200t-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/models/clb_models.py", line 528, in process_slice
[...]

PR #1629, opened 2020-08-04
Make sure DLUT.RAM is emitted if using LUT-RAMs.

litghost:
Trying out suggested fix for #1625


litghost:
A200T vendor is green this go around. I'm gonna re-run, but this might fix #1625 for now?

PR #1634, opened 2020-08-11
[DNM] WIP: Architecture definition for UltraScale+

tmichalak:
This is a WIP PR that aims at adding UltraScale+ architecture to arch-defs.
The main idea is to generate the architecture XML file without having to write the pb_types and model by hand, but by using v2x.


tmichalak:
Superseded by #1651

PR #1638, opened 2020-08-17
Configure Kokoro to build the 200t install package

HackerFoo:

PR #1640, opened 2020-08-18
xc7: yosys: Use params plugin to update PLL's PHASE params

tmichalak:
This PR integrates Yosys's params plugin created in the [yosys-symbiflow-plugins](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/26)
Due to the fact that Yosys doesn't support floating parameter values i.e. treats them as strings, the parameter values need to be multiplied by 1000 for the PLL registers to have correct values calculated during techmapping.

This PR depends on the [PR to the conda-packages repo](https://github.com/SymbiFlow/conda-packages/pull/128)


xc/xc7/yosys/utils.tcl
```diff
@@ -0,0 +1,22 @@
...
+# i.e. treats them as strings, the parameter values need to be multiplied by 1000
+# for the PLL registers to have correct values calculated during techmapping.
+proc update_pll_params {} {
+    set temp_file "[pid].tmp"
+    select t:PLLE2_ADV -list -write $temp_file
+    set fh [open $temp_file r]
+    foreach cell [gets $fh] {
+	for {set output 0} {$output < 6} {incr output} {
+	    set param_name "CLKOUT${output}_PHASE"
+	    set param_value [getparam $param_name t:PLLE2_ADV]
```
mkurc-ant, line 12: Shouldn't you select the cell by its name here?

xc/xc7/yosys/utils.tcl
```diff
@@ -0,0 +1,22 @@
...
+	    if {$param_value > 0} {
```
mkurc-ant, line 13: I think that this condition should be removed. I'm not sure about that but the value of "0.0" may not be interpreter by Yosys as integer 0. So its better to process it always.

xc/xc7/yosys/utils.tcl
```diff
@@ -0,0 +1,22 @@
...
+		set new_param_value [expr {$param_value * 1000}]
```
mkurc-ant, line 14: That may still yield in floating point value but I'm not a TCL expert. I'd do `[expr int(round($param_value * 1000))]`

xc/xc7/yosys/utils.tcl
tmichalak: `t:PLLE2_ADV` selects all cells of type PLLE2_ADV. Following Yosys's doc:
`t:<pattern>
        all cells with a type matching the given pattern
`

xc/xc7/yosys/utils.tcl
tmichalak: The concern here was more about the param value being empty.
I changed the condition to `ne ""` (non-empty string) so the `0.0` get converted to 0.

xc/xc7/yosys/utils.tcl
tmichalak: Who is a TCL expert? :) Nice catch though, I will use your TCL snippet so that we actually only set integer values on this parameter.

xc/xc7/yosys/utils.tcl
mkurc-ant: Yeah, that is correct. But what if there is more than one PLLE2_ADV cell in the design? Which one you will retrieve the parameter from?

tmichalak: Oh, now I see what you meant. I thought you were referring to `select` not `getparam`. 
Of course, that's a bug, should be `$cell` instead. Thanks.

litghost:
@tmichalak Doesn't look like the way you are using the TCL include worked on install?  Maybe the file didn't get copied?

xc/xc7/tests/soc/litex/base/baselitex_arty100t.v
```diff
@@ -13399,7 +13399,7 @@ PLLE2_ADV #(
 	.CLKOUT1_DIVIDE(3'd5),
 	.CLKOUT1_PHASE(1'd0),
 	.CLKOUT2_DIVIDE(3'd5),
-	.CLKOUT2_PHASE(90000),
+	.CLKOUT2_PHASE(7'd90),
```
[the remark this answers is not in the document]
litghost: Still looks like you have it as 7'd90 rather than 90.0?

xc/xc7/yosys/utils.tcl
```diff
@@ -0,0 +1,22 @@
+# Update the CLKOUT[0-5]_PHASE parameter values.
+# Due to the fact that Yosys doesn't support floating parameter values
+# i.e. treats them as strings, the parameter values need to be multiplied by 1000
+# for the PLL registers to have correct values calculated during techmapping.
+proc update_pll_params {} {
+    set temp_file "[pid].tmp"
+    select t:PLLE2_ADV -list -write $temp_file
```
litghost, line 7: Can we avoid writing this file?

litghost: Maybe "select" should have an option to set the TCL return as a list, rather than writing a file?

tmichalak: I didn't find any other solution to this, since `select` doesn't return anything in TCL

tmichalak:
> CI is red because of utils script not being found?

Yes, I fixed it for 50T as it was failing before, probably forgot it for 200T. Should be a simple fix.

tmichalak:
@litghost @HackerFoo is there a different install target for 200t? I fixed the `util.tcl` script not being copied to the install package, but for some reason 200t fails and the log is empty.

litghost:
> @litghost @HackerFoo is there a different install target for 200t? I fixed the `util.tcl` script not being copied to the install package, but for some reason 200t fails and the log is empty.

200T install is still being worked on in #1638 , which is why it failed in this PR.

xc/xc7/yosys/utils.tcl
litghost: I'm suggesting we modify yosys to take a flag like `-tcl_return` or something.

tmichalak:
> 200T install is still being worked on in #1638 , which is why it failed in this PR.

I guess this PR is then good to merge then?

xc/xc7/tests/soc/litex/base/baselitex_arty100t.v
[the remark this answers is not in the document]
litghost: @tmichalak you still have this as a decimal instead of float?

xc/xc7/yosys/utils.tcl
litghost: @tmichalak This issue is still outstanding?

xc/xc7/tests/soc/litex/base/baselitex_arty100t.v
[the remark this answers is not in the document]
tmichalak: `7'd90` is how Litex generates the attribute value and we wanted to have it as close to the original as possible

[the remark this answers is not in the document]
litghost: Okay, got it.

tmichalak:
> Feedback still outstanding

Which are the outstanding parts?

litghost:
> > Feedback still outstanding
> 
> Which are the outstanding parts?

I @'d you on them

xc/xc7/yosys/utils.tcl
tmichalak: I agree this would be better... But isn't changing Yosys something we want to avoid through using plugins?
Getting this change upstreamed can take some time. Do you want to use SymbiFlow's Yosys fork?

litghost: I'm hopeful this is something that the upstream Yosys folks would quickly accept or offer a solution too.  I'll reach out to Eddie to test the waters

litghost: So Eddie proposed another idea that we could do that could be done as a plugin, and with further iteration could be pushed upstream.  So rather than directly modify `select` to set the `Tcl_Return`, we could create a new yosys command like `selection_to_tcl` which converts the current selection into a Tcl list and sets the Tcl_Return as expected.

Sounds good?

tmichalak: The new command variant sound less intrusive thus more likely for the Yosys guys to accept.

litghost: > The new command variant sound less intrusive thus more likely for the Yosys guys to accept.

That was the idea.  It is also something we can do as a plugin right now, while also proposing and iterating it upstream.

tmichalak: The plugin is ready: see [PR](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/29)

tmichalak:
The PR will fail until the [selection plugin](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/29) gets merged and the plugin package version updated.

tmichalak:
@litghost CI is green (apart from 200t install) so I guess it can be merged.

litghost:
Phase stuff looks good.  Isn't this PR missing the DUTY_CYCLE parameter, per: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1648/files#diff-26e20e1f3e19fe08c5ecc785605bb29fR28 ?

tmichalak:
> Phase stuff looks good. Isn't this PR missing the DUTY_CYCLE parameter

Right, added the DUTY_CYCLE parameter. If at some point we will need to add more we can create separate PRs?

xc/xc7/yosys/utils.tcl
```diff
@@ -0,0 +1,21 @@
+# Update the CLKOUT[0-5]_PHASE and CLKOUT[0-5]_DUTY_CYCLE parameter values.
+# Due to the fact that Yosys doesn't support floating parameter values
+# i.e. treats them as strings, the parameter values need to be multiplied by 1000
+# for the PLL registers to have correct values calculated during techmapping.
+proc update_param { param_name cell } {
+    set param_value [getparam $param_name $cell]
+    if {$param_value ne ""} {
+	set new_param_value [expr int(round($param_value * 1000))]
```
mkurc-ant, line 8: There is one issue with that multiplier - its different for PHASE and DUTY_CYCLE parameters. For phase its 1000 but for DUTY_CYCLE its 100*1000=100000. I think that it is due to my mistake when adapting the Verilog code that calculates PLL register content. The code required duty cycle as percent (0 - 100), hence the additional x100. Vivado requires it to be given as a fraction (0.0 - 1.0).

The issue can be corrected here or in the PLL techmap.

mkurc-ant (changes requested):

xc/xc7/yosys/utils.tcl
tmichalak: Ok, corrected the multiplier value.

litghost (approved):
LGTM, merge if green

PR #1643, opened 2020-08-19
toolchain: remove necessity of PCF file

acomodi:

### DIFF
--- a/xc/xc7/toolchain_wrappers/symbiflow_generate_constraints
+++ b/xc/xc7/toolchain_wrappers/symbiflow_generate_constraints
@@ -4,12 +4,16 @@ set -e
 MYPATH=`realpath $0`
 MYPATH=`dirname ${MYPATH}`
 
-PCF=$1
-EBLIF=$2
-NET=$3
-PART=$4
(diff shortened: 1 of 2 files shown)


acomodi:
Indeed, the XDC file needs to be given as a parameter to the synthesis step.
I could add that at placement step as well, to enforce the fact that either one of PCF or XDC needs to be provided, otherwise IOs would get placed randomly.

I am thinking whether it is a good idea to support only the XDC files as input constraints, to avoid having confusion on which one to use (from a user perspective). This would also make it easier to error out if the XDC is not given as input.

litghost:
> Indeed, the XDC file needs to be given as a parameter to the synthesis step.
> I could add that at placement step as well, to enforce the fact that either one of PCF or XDC needs to be provided, otherwise IOs would get placed randomly.
> 
> I am thinking whether it is a good idea to support only the XDC files as input constraints, to avoid having confusion on which one to use (from a user perspective). This would also make it easier to error out if the XDC is not given as input.

I was just asking to check.  I agree we should phase out use of PCF files completely now that XDC support is in place.

PR #1647, opened 2020-08-25
SDC plugin integration

tmichalak:
This PR adds support for the use of SDC plugin and adjusts the mini_ddr litex design to demonstrate that it works.
It's WIP because it depends on the [SDC plugin](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/27) and has to be rebased once the [params plugin](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1640] gets merged.


xc/xc7/yosys/synth.tcl
```diff
@@ -21,10 +25,20 @@ if { $::env(USE_ROI) == "TRUE" } {
...
 
     # Start flow after library reading
-    synth_xilinx -vpr -flatten -abc9 -nosrl -noclkbuf -nodsp -iopad -nowidelut -run prepare:check
+    synth_xilinx -vpr -flatten -abc9 -nosrl -nodsp -iopad -nowidelut -run prepare:check
+}
+
+# Read the SDC file
+if { [info exists ::env(INPUT_SDC_FILE)] && $::env(INPUT_SDC_FILE) != "" } {
+  read_sdc $::env(INPUT_SDC_FILE)
+  propagate_clocks
```
litghost, line 37: Why do you call it 3 times?

xc/xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
```diff
@@ -17,6 +18,7 @@ add_fpga_target(
     minilitex_ddr_arty.v
     VexRiscv_Lite.v
   INPUT_XDC_FILE minilitex_ddr_arty.xdc
+  INPUT_SDC_FILE arty.input.sdc
```
litghost, line 21: What is the difference between `INPUT_SDC_FILE` and `SDC_FILE`?

xc/xc7/yosys/synth.tcl
tmichalak: That is a workaround to the fact that the buffer propagation doesn't reach a net that drives a PLL thus fails to find a PLL and perform the clock divider propagate. Basically, the first buffer propagation traverses through the IBUF, hence doesn't find the PLL. Only the second `propagate_clocks` triggers another buffer propagation from IBUF to BUFG and the following ClockDivider propagation does find the PLL and succeeds the clock divider propagation.
Long story short, that needs a fix, but was good enough to test.

xc/xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
tmichalak: INPUT_SDC_FILE contains a SDC command given by the user, e.g. with a single `create_clock` command, the SDC_FILE is the one generated by Yosys. I kept it for backward compatibility with other existing test cases that do not use the plugin yet. In the future there should be no need for the generated SDC, so SDC_FILE in this case.

xc/xc7/yosys/synth.tcl
litghost: Ya, this needs to be handled internally.  3 iterations might be enough, 1 might be enough.

xc/xc7/tests/soc/litex/mini_ddr/CMakeLists.txt
litghost: Why does the generated file (e.g. `SDC_FILE`) need to be exposed to the user at all?

tmichalak: It doesn't. This was just a way of introducing the change without affecting the rest of the tests which would still use the "old" flow. However, if the plugin works then if the input SDC already contains all necessary clocks then the output sdc shouldn't differ, hence there should be no impact on the test.
I made the changes to have VPR use the generated SDC.

tmichalak: This is fixed. But the PR still depends on the SDC plugin.

common/cmake/devices.cmake
```diff
@@ -1305,6 +1305,8 @@ function(ADD_FPGA_TARGET)
   set(OUT_SYNTH_V ${OUT_LOCAL}/${TOP}_synth.v)
   set(OUT_SYNTH_V_REL ${OUT_LOCAL_REL}/${TOP}_synth.v)
   set(OUT_FASM_EXTRA ${OUT_LOCAL}/${TOP}_fasm_extra.fasm)
+  set(OUT_SDC ${OUT_LOCAL}/${TOP}.sdc)
```
litghost, line 1308: Should be `${TOP}_synth.sdc` to be clear this is a post-synthesis SDC file.

Also worth making a comment in the CMake about what this file is

tests/9-scalable_proc/basys3.xdc
```diff
@@ -0,0 +1,2 @@
+create_clock -period 10.0 [get_ports clk]
```
litghost, line 1: Is the "get_ports" call required here?

tests/9-soc/murax/CMakeLists.txt
```diff
@@ -69,7 +69,6 @@ foreach(FREQ ${BASYS_FREQS})
         basys3-full_toplevel_${FREQ}.v
         Murax.v
       INPUT_IO_FILE basys3-full.pcf
-      SDC_FILE basys3-full_${FREQ}.sdc
```
litghost, line 72: Why not merge the PCF with the SDC in the XDC here?

tests/9-scalable_proc/basys3.xdc
tmichalak: It is not, `clk` alone will be enough

xc/xc7/tests/soc/litex/base/baselitex_arty.xdc
```diff
@@ -271,7 +271,7 @@ set_property INTERNAL_VREF 0.750 [get_iobanks 34]
 
 #create_clock -name sys_clk -period 16.666 [get_nets sys_clk]
 #
-#create_clock -name clk100 -period 10.0 [get_nets clk100]
+create_clock -name clk100 -period 10.0 [get_ports clk100]
```
tmichalak, line 274: `get_nets` will basically return the same thing as `get_ports` in this case, but if it is us creating these XDCs we can just skip the `get_` commands and make `clk100` alone.

common/cmake/devices.cmake
```diff
@@ -1472,8 +1466,9 @@ function(ADD_FPGA_TARGET)
     )
 
   if(NOT "${ADD_FPGA_TARGET_SDC_FILE}" STREQUAL "")
```
tmichalak: Shouldn't we make this part dependent on whether we have the INPUT_XDC_FILE and add `--sdc_file ${OUT_SDC}`
I mean where do we add the dependency of VPR on the generated SDC?

xc/xc7/yosys/synth.tcl
```diff
@@ -25,16 +26,20 @@ if { $::env(USE_ROI) == "TRUE" } {
...
+    synth_xilinx -vpr -flatten -abc9 -nosrl -nodsp -iopad -nowidelut -run prepare:check
 }
+
 if { [info exists ::env(INPUT_XDC_FILE)] && $::env(INPUT_XDC_FILE) != "" } {
   read_xdc -part_json $::env(PART_JSON) $::env(INPUT_XDC_FILE)
   write_fasm -part_json $::env(PART_JSON)  $::env(OUT_FASM_EXTRA)
 }
 
+# Perform clock propagation based on the information from the SDC commands
+propagate_clocks
```
tmichalak, line 40: I guess propagate_clocks  should go into the INPUT_XDC_FILE if. It doesn't make anything if we don't read the SDC commands and in this flow we do this in `read_xdc`.

litghost: Sorry, can you rephrase your statement.  I'm not sure I understand what you are saying.

common/cmake/devices.cmake
litghost: Sorry, can you rephrase your statement.  I'm not sure I understand what you are saying.

xc/xc7/yosys/synth.tcl
tmichalak: The `propagate_clocks` command should be called inside the `if` statement right after `read_xdc`.

common/cmake/devices.cmake
tmichalak: It looks to me that we are not passing the generated SDC to VPR, but the one specified as SDC_FILE in the CMakeLists.txt. We should have something akin:
```
    append_file_dependency(VPR_DEPS ${OUT_SDC_REL})
    set(SDC_ARG --sdc_file ${OUT_SDC})
```

tests/9-soc/murax/CMakeLists.txt
acomodi: I think this could be done as a followup cleanup PR, where we finally get all the constraints file unified.
This PR is intended to just make use of the XDC instead of SDC regarding clock constraints.

common/cmake/devices.cmake

tests/9-scalable_proc/basys3.xdc
acomodi: Changed to `clk` alone

xc/xc7/tests/soc/litex/base/baselitex_arty.xdc
acomodi: set to `clk100` alone

common/cmake/devices.cmake
acomodi: Fixed the dependencies. Now, there are three possibilities:

- NO SDC, and only the XDC: dependencies added on the generated SDC file
- NO XDC, and only the SDC: dependencies added on the input SDC file
- SDC and XDC: dependencies added on both the generated and input SDC, but the SDC that is going to be used is the user-provided one instead of the autogenerated one.

xc/xc7/yosys/synth.tcl
acomodi: Done

common/cmake/devices.cmake
litghost: > NO XDC, and only the SDC: dependencies added on the input SDC file

Sorry, do you mean that the input SDC is modified?  Or do you mean that a new SDC is generated with the propigated outputs?

acomodi: This means that the SDC is just read, without propagation and it needs to be read manually.

common/cmake/devices.cmake
```diff
@@ -1420,10 +1436,25 @@ function(ADD_FPGA_TARGET)
...
+
+  set(SDC_ARG "")
+  if(NOT "${ADD_FPGA_TARGET_INPUT_XDC_FILE}" STREQUAL "")
+    append_file_dependency(VPR_DEPS ${OUT_SDC_REL})
+    get_file_location(SDC_LOCATION ${OUT_SDC_REL})
+    set(SDC_ARG --sdc_file ${SDC_LOCATION})
+  endif()
+
+  # Overrides the SDC if the INPUT_SDC_FILE parameter is set
+  if(NOT "${ADD_FPGA_TARGET_INPUT_SDC_FILE}" STREQUAL "")
```
litghost, line 1449: Generate an error if both are supplied?

acomodi: Done

acomodi:
@litghost CI went green and latest comments addressed.

Issue #1639, opened 2020-08-18
Add sv2v support


rw1nkler:
Integrate sv2v with `symbiflow-arch-defs` build system. This feature will allow us to use Ibex and EarlGrey sources directly and convert them during the bitstream generation process.

PR #1648, opened 2020-08-25
Add sv2v conversion to SymbiFlow build system

rw1nkler:
This PR adds the sv2v conversion mechanism to the `symbiflow-arch-defs` build system. 
The provided changes have been tested on the Ibex example which now uses the original SystemVerilog sources.
The generated Ibex bitstreams work on both Arty and Nexys Video boards.

Resolves https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1639


common/cmake/devices.cmake
```diff
@@ -1153,11 +1154,13 @@ function(ADD_FPGA_TARGET)
   #   [DEFINES <definitions>]
   #   [SDC_FILE <sdc file>]
   #   [BIT_TO_V_EXTRA_ARGS]
+  #   [SV2V_CONVERT]
```
litghost, line 1157: SV2V should be done before ADD_FPGA_TARGET, rather than as part of it.

common/cmake/devices.cmake
```diff
@@ -1143,6 +1143,7 @@ function(ADD_FPGA_TARGET)
   #   NAME <name>
   #   [TOP <top>]
   #   BOARD <board>
+  #   [INCLUDES <include list>]
```
litghost, line 1146: This shouldn't be required, if you set `add_file( ... SCANNER_TYPE verilog)`

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,138 @@
...
+  prim_assert_yosys_macros.svh
+  prim_assert_standard_macros.svh
+)
+
+add_custom_command(
+  OUTPUT ${SOURCE_FILE_LIST} ${INCLUDE_FILE_LIST}
+  COMMAND
+    bash -c \"cd ${IBEX_DIR} && git apply ${CMAKE_CURRENT_SOURCE_DIR}/ibex.patch\"
+  COMMAND
+    ${PYTHON3} -m pip install -r ${IBEX_DIR}/python-requirements.txt &&
```
litghost, line 62: Python dependencies should be installed in the root `make env`, not within CMake.

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,138 @@
...
+    fusesoc --cores-root=${IBEX_DIR} run --target=synth --setup lowrisc:ibex:top_artya7 --part xc7a35ticsg324-1L
+  COMMAND
+    for file in ${SOURCE_FILE_LIST}\; do find build -name \"$$file\" -exec cp {} . \\\; \; done
```
litghost, line 65: Why the copy step here?

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,138 @@
...
+  COMMAND
+    for file in ${INCLUDE_FILE_LIST}\; do find build -name \"$$file\" -exec cp {} . \\\; \; done
+  COMMAND
+    bash -c \"cd ${IBEX_DIR} && git apply -R ${CMAKE_CURRENT_SOURCE_DIR}/ibex.patch\"
```
litghost, line 69: Should probably use "patch", rather than git apply?

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,138 @@
...
+  prim_util_memload.svh
+  prim_assert_dummy_macros.svh
+  prim_assert_yosys_macros.svh
+  prim_assert_standard_macros.svh
+)
+
+add_custom_command(
+  OUTPUT ${SOURCE_FILE_LIST} ${INCLUDE_FILE_LIST}
+  COMMAND
+    bash -c \"cd ${IBEX_DIR} && git apply ${CMAKE_CURRENT_SOURCE_DIR}/ibex.patch\"
```
litghost, line 60: Why is the patch applied twice?

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,138 @@
...
-# Arty
+foreach(INC ${INCLUDE_FILE_LIST})
+  add_file_target(FILE ${INC} GENERATED)
+endforeach()
+
+# sv2v conversion
+
+add_sv2v_target(
+  NAME ibex
+  INCLUDES ${INCLUDE_FILE_LIST}
```
litghost, line 88: Rather than add flags for "INCLUDES", the include list should be attached to the file target like so: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/common/cmake/file_targets.cmake#L246-L295

common/cmake/devices.cmake
litghost: SCANNER_TYPE won't work because the files are generated during the build, and scanning happens at CMake config time.  Is there any way for fusesoc to output a dependency graph for use in CMake?

common/cmake/sv2v.cmake
```diff
@@ -0,0 +1,87 @@
...
+  #
+  # Output files:
+  #
+  # * <name>_sv2v.v - Output Verilog file
+  # * <name>_sv2v.v.log - Log file with errors and warnings during the sv2v conversion
+  #
+  # Outputs for this target will be located in the ${CMAKE_CURRENT_BINARY_DIR}
+  # ~~~
+
+  set(options EXPLICIT_ADD_FILE_TARGET)
```
litghost, line 33: We should not add "EXPLICIT_ADD_FILE_TARGET" to new commands.

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,138 @@
-add_file_target(FILE led.vmem)
-add_file_target(FILE pins_artya7.sdc)
+get_filename_component(IBEX_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../../../../third_party/ibex ABSOLUTE)
+
+set(SV2V_ROM_INIT_FILE led.vmem)
+set(SV2V_IBEX_FLAGS --define=RegFile=ibex_pkg::RegFileFF --define=SRAMInitFile=${SV2V_ROM_INIT_FILE})
+
+set(SOURCE_FILE_LIST
```
litghost, line 6: Can `fusesoc` supply this list of files during CMake configure time?

common/cmake/devices.cmake
rw1nkler: I think that this is not possible directly. However, `fusesoc` produces auxiliary YAML file with the configuration data.
The information about the files can be extracted from that file.

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: To move the source files, from rather complicated tree-like `fusesoc` build directory to the cmake binary directory. 
Probably, I probably should have used symbolic links there.

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: In the second time, I'm removing the patch using the `-R` option. Without this removal, the test will fail if run multiple times

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: We have to generate all the sources to obtain this list.

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: Unfortunately, the `python-requirements.txt` file cannot be added directly to the `environment.yml`, because it uses nested recursive imports (https://github.com/lowRISC/ibex/blob/master/python-requirements.txt#L20). For some reason, the path cannot be correctly resolved, and the mentioned import causes the following error:
```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/third_party/ibex/vendor/google_riscv-dv/requirements.txt'
```

I will copy the necessary packages directly to the `environment.yml`

xc/xc7/tests/soc/ibex/CMakeLists.txt
litghost: This solution is not acceptable.  Consider what happens if a SIGINT (e.g. Ctrl-C) is sent after the patch is applied, but before it is reversed.

common/cmake/devices.cmake
litghost: I agree that the "SCANNER_TYPE verilog" won't work, but there is still not reason to add the "INCLUDES" flag.

xc/xc7/tests/soc/ibex/CMakeLists.txt
litghost: @mithro thoughts?

common/cmake/sv2v.cmake
```diff
@@ -0,0 +1,87 @@
...
+  # Find file locations for INCLUDES and SOURCES
+
+  set(SOURCES_LOCATION "")
+  foreach(SRC ${ADD_SV2V_TARGET_SOURCES})
+    append_file_location(SOURCES_LOCATION ${SRC})
+  endforeach()
+
+  set(INCLUDES_LOCATION "")
+  foreach(INC ${ADD_SV2V_TARGET_INCLUDES})
+    append_file_location(INCLUDES_LOCATION ${INC})
```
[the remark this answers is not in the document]
rw1nkler: Thank you! Done

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: Done

common/cmake/sv2v.cmake
rw1nkler: Removed

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: I've used `add_dependency` function for creating proper dependencies between the files. Since add_dependency cannot be used with the file path, I used the `get_file_target`:
https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1648/files#diff-d597ff989f92a27dd2043d153d9fb67dR40

common/cmake/devices.cmake
rw1nkler: See https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1648#discussion_r483595644

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: Changed in the latest version of the test

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: I proposed another solution based on the copying the Ibex repository to tmp directory:
https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1648/files#diff-8c90f694e22df81d3f6935e0c2c40dadR42

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: I created  a `generate.sh` script which can be used for obtaining the dependency list. During the CMake configuration, the script copies the files to tmp directory, applies the patches and generates the sources. Then the `fusesoc_get_sources.py` is used to obtain the list of source files.

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1648/files#diff-b7cb75ca59993a879e29bed8cd8469b8R22

rw1nkler:
@litghost I see that the VTR+SymbiFlow Architecture Definitions CI is failing. Probably this is caused by the fact that the building script still contains the old environment management mechanism:

https://github.com/SymbiFlow/vtr-verilog-to-routing-kokoro/blob/master/steps/arch-defs-test.sh#L21

xc/xc7/tests/soc/ibex/ibex.patch
```diff
@@ -0,0 +1,112 @@
...
+     .STARTUP_WAIT         ("FALSE"),
+     .DIVCLK_DIVIDE        (1),
+     .CLKFBOUT_MULT        (12),
+-    .CLKFBOUT_PHASE       (0.000),
++    .CLKFBOUT_PHASE       (0),
+     .CLKOUT0_DIVIDE       (24),
+-    .CLKOUT0_PHASE        (0.000),
+-    .CLKOUT0_DUTY_CYCLE   (0.500)
++    .CLKOUT0_PHASE        (0),
++    .CLKOUT0_DUTY_CYCLE   (500)
```
mkurc-ant, line 28: @rw1nkler I don't remember why but in SymbiFlow `*_PHASE` should be multiplied by 1000 but `*_DUTY_CYCLE` should be multiplied by **100000**. So to get 50% duty cycle it has to be either changed to 50000 or the parameter should not be given at all (then it defaults to 50% duty).

mkurc-ant (changes requested):

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,107 @@
...
+)
 
-# Arty
+# Get all the file targets for the includes and vmem files
+
+set(SOURCE_DEPS "")
+foreach(INC ${INCLUDE_FILE_LIST})
+  add_file_target(FILE ${INC} GENERATED)
+  get_file_target(TARGET ${INC})
+  list(APPEND SOURCE_DEPS ${TARGET})
```
litghost, line 46: So the files from the `INCLUDE_FILE_LIST` should be added to the source file `INCLUDE_FILES` properties, and that should be all that is required here.  It will simplify these couple lines.

E.g. https://github.com/SymbiFlow/symbiflow-arch-defs/blob/b3864ae63d668fc7d321b94ca946841e80dd0bb6/common/cmake/file_targets.cmake#L293

common/cmake/sv2v.cmake
```diff
@@ -0,0 +1,62 @@
+# This CMake include defines the following functions:
+#
+# * ADD_SV2V_TARGET - Converts SystemVerilog sources to Verilog using sv2v
+
+function(ADD_SV2V_TARGET)
```
litghost, line 5: This looks much better, thanks

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: This is no longer needed after the changes introduced in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1640

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: Done!

utils/fusesoc_get_sources.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
```
litghost, line 1: Should this move upstream into fusesoc in the future?

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,102 @@
...
+get_file_target(SV2V_ROM_INIT_FILE_TARGET ${SV2V_ROM_INIT_FILE})
+
+foreach(INC ${INCLUDE_FILE_LIST})
+  add_file_target(FILE ${INC} GENERATED)
+endforeach()
+
+foreach(SRC ${SOURCE_FILE_LIST})
+  add_file_target(FILE ${SRC} GENERATED)
+  get_file_target(TARGET ${SRC})
+  add_dependencies(${TARGET} ${SV2V_ROM_INIT_FILE_TARGET})
```
litghost, line 52: Why not just have the `SV2V_ROM_INIT_FILE` also be part the INCLUDE_FILES property?

xc/xc7/tests/soc/ibex/CMakeLists.txt
```diff
@@ -1,78 +1,102 @@
-add_file_target(FILE led.vmem)
-add_file_target(FILE pins_artya7.sdc)
+get_filename_component(IBEX_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../../../../third_party/ibex ABSOLUTE)
```
litghost, line 1: Overall this file is looking a lot better, thanks!

xc/xc7/tests/soc/ibex/ibex.patch
```diff
@@ -0,0 +1,110 @@
...
+diff --git a/shared/rtl/fpga/xilinx/clkgen_xil7series.sv b/shared/rtl/fpga/xilinx/clkgen_xil7series.sv
+index e41b4b0..e080de7 100644
+--- a/shared/rtl/fpga/xilinx/clkgen_xil7series.sv
++++ b/shared/rtl/fpga/xilinx/clkgen_xil7series.sv
+@@ -27,9 +27,9 @@ module clkgen_xil7series (
+     .STARTUP_WAIT         ("FALSE"),
+     .DIVCLK_DIVIDE        (1),
+     .CLKFBOUT_MULT        (12),
+-    .CLKFBOUT_PHASE       (0.000),
++    .CLKFBOUT_PHASE       (0),
```
litghost, line 23: What's going on here with `0.000` -> `0`?  Is it tracked on the burndown list?

xc/xc7/tests/soc/ibex/ibex.patch
```diff
@@ -0,0 +1,110 @@
+diff --git a/examples/fpga/artya7/rtl/top_artya7.sv b/examples/fpga/artya7/rtl/top_artya7.sv
+index c26ca45..be19828 100644
+--- a/examples/fpga/artya7/rtl/top_artya7.sv
++++ b/examples/fpga/artya7/rtl/top_artya7.sv
+@@ -11,7 +11,7 @@ module top_artya7 (
+   parameter int          MEM_SIZE     = 64 * 1024; // 64 kB
+   parameter logic [31:0] MEM_START    = 32'h00000000;
+   parameter logic [31:0] MEM_MASK     = MEM_SIZE-1;
+-  parameter              SRAMInitFile = "";
++  parameter              SRAMInitFile = "led.vmem";
```
litghost, line 10: Why is this change required?

xc/xc7/tests/soc/ibex/ibex.patch
```diff
@@ -0,0 +1,110 @@
...
+     .O (clk_50_buf)
+diff --git a/vendor/lowrisc_ip/prim/rtl/prim_lfsr.sv b/vendor/lowrisc_ip/prim/rtl/prim_lfsr.sv
+index cf35642..3b60d04 100644
+--- a/vendor/lowrisc_ip/prim/rtl/prim_lfsr.sv
++++ b/vendor/lowrisc_ip/prim/rtl/prim_lfsr.sv
+@@ -299,7 +299,7 @@ module prim_lfsr #(
+   ////////////////
+   // Galois XOR //
+   ////////////////
+-  if (64'(LfsrType) == 64'("GAL_XOR")) begin : gen_gal_xor
```
litghost, line 60: Is this something sv2v doesn't handle?

xc/xc7/tests/soc/ibex/ibex.patch
```diff
@@ -0,0 +1,110 @@
...
+-    .CLKFBIN             (clk_fb_buf),
++    .CLKFBIN             (clk_fb_unbuf),
+     .CLKIN1              (io_clk_buf),
+     .CLKIN2              (1'b0),
+      // Tied to always select the primary input clock
+@@ -59,12 +59,6 @@ module clkgen_xil7series (
+     // Do not reset PLL on external reset, otherwise ILA disconnects at a reset
+     .RST                 (1'b0));
+ 
+-  // output buffering
```
litghost, line 43: How confident are we that this is the right thing to do?  It looks wrong to my eyes?

litghost (commented):
This PR is starting to look really good!  Most of the remaining questions are really about the patches we make to Ibex, and some of that can be moved to the burn down list if they are of high complexity.

utils/fusesoc_get_sources.py
rw1nkler: Definitely! It would be a great feature! 

On the fusesoc side, it can be done using internal data structures. 
In this small example, I think It is better to parse the Edalize YAML produced by fusesoc. Otherwise, we will have to maintain another forked tool. We can independently prepare PR to fusesoc and change the example when it is merged.

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: This is required by the diff_fasm step. Currently, VPR routes the feedback clock through:
``CLKFBOUT->BUFG->BUFHCE->CLKFBIN``, this configuration seems to be not allowed by the Vivado:

```
ERROR: [DRC REQP-1684] Feedback check: Unsupported PLLE2_ADV connectivity. The signal CLK_HROW_TOP_R_X60Y78_BUFHCE_X1Y20_O on the CLKFBIN pin of clkgen.pllCMT_TOP_L_UPPER_T_X106Y96_PLLE2_ADV_X1Y1_PLLE2_ADV with COMPENSATION mode 'ZHOLD' must be driven by the CLKFBOUT pin on the same cell through a BUFG, BUFH, or BUFR.
```

However, the bitstream generated by both runs (with and without this patch) works on HW.

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: Currently, It is required since we don't have any option to change the parameter values.
This can be done using `chparam -set SRAMInitFile="led.vmem"` in Yosys after reading the design.

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: This seems to be a corner case. I will add it to the burndown list. 
Without this modification, Yosys fails during synthesis.

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: Both cases with or without the patch work on HW

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: Yes, those variable casting creates strange wrappers in the final Verilog file, which cannot be read by the Yosys:
```
/mnt/hdd1/arch_defs_ws/ibex_PR/build/xc/xc7/tests/soc/ibex/ibex_sv2v.v:582: ERROR: Condition for generate if is not constant!
```

xc/xc7/tests/soc/ibex/CMakeLists.txt
rw1nkler: Done!

xc/xc7/tests/soc/ibex/ibex.patch
litghost: This should be in the burndown list, and identify if this is yosys bug or a sv2v bug.

xc/xc7/tests/soc/ibex/ibex.patch
litghost: Just because it works in HW doesn't mean this is the right thing to do.  This is actually likely a fasm2bels deficency, as the explicit BUFHCE is a choice made by fasm2bels.

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: Added to burndown list

xc/xc7/tests/soc/ibex/ibex.patch
rw1nkler: This is a missing Yosys functionality. It seems to be added to the Yosys upstream, since the latest Yosys can parse the sv2v output without the patch.

PR #1651, opened 2020-08-28
Initial U-Ray database import

mkurc-ant:
This PR adds a skeleton for Xilinx US/US+ support in SymbiFlow. There were some minor changes to `xc/common/cmake` files to accomodate X-Ray and U-Ray specifics. I assumed that all db import scripts will be named like `prjxray_<something>.py` and `prjuray_<something>.py`. Most of the required files that are not critical to early stages of db import are just templates/placeholders for now.

Apart from all necessary directories and CMakes that allows `make env` to succeed it also adds the channel formation script. `prjuray_form_channels.py`. The script has been adjusted to US/US+ specifics (eg. different const network connections) so that it succeeds and the `channels.db` can be successfully created.

With this PR two submodules were added: `prjuray-db` and `prjuray-tools`. The latter one is meant to provide U-Ray database access code after https://github.com/SymbiFlow/prjuray-tools/pull/3 is successfully landed (without it the db import fails).


xc/common/cmake/arch_define.cmake
```diff
@@ -106,7 +106,7 @@ function(ADD_XC_ARCH_DEFINE)
     CELLS_SIM ${YOSYS_CELLS_SIM} ${VPR_CELLS_SIM}
     VPR_ARCH_ARGS ${VPR_ARCH_ARGS}
     RR_PATCH_TOOL
-      ${symbiflow-arch-defs_SOURCE_DIR}/xc/common/utils/prjxray_routing_import.py
+      ${symbiflow-arch-defs_SOURCE_DIR}/xc/common/utils/${PRJRAY_NAME}_routing_import.py
```
litghost, line 109: I'd prefer if don't hard fork these scripts.  They aren't small pieces of code to begin with, and I'm not excited about needing to maintain two sets.  Can we factor out the xc7 vs xcu specific stuff?

mkurc-ant: Actually I agree. I took that approach with `assign_tile_pin_direction.py`.

So the idea would be to have the common code in eg. `rounting_import.py` plus two wrapper scripts `prjxray_routing_import.py` and `prjuray_routing_import.py` that would contain the specific code. These specific scripts would then be used in CMake. Does that make sense?

litghost: Sounds good to me.

mkurc-ant:
@litghost I created a doc with brief description of my ideas for the const network: https://docs.google.com/document/d/1Lhkqsa9YTY03ZpmeB6Q2POEYa8Tgm1yIrSBuV5bNADQ

xc/common/utils/prjuray_timing.py
```diff
@@ -0,0 +1,530 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
litghost, line 2: Why does this need a full copy/paste?

xc/xcu/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+add_subdirectory(tests)
+
+get_target_property_required(PYTEST env PYTEST)
+
+add_custom_target(
+  test_python_xcu
+  DEPENDS
+    conda_pytest all_pip
+  COMMAND ${CMAKE_COMMAND} -E env
+    PYTHONPATH=${symbiflow-arch-defs_SOURCE_DIR}/third_party/prjuray-tools:${symbiflow-arch-defs_SOURCE_DIR}/utils#:${symbiflow-arch-defs_SOURCE_DIR}/third_party/prjuray/third_party/fasm
```
litghost, line 28: We should change requirements.txt to stop needing to modify the paths like this, yes?

xc/common/utils/prjuray_form_channels.py
```diff
@@ -0,0 +1,481 @@
...
+from form_channels import form_tracks
+
+from prjxray_db_cache import DatabaseCache
+
+from prjuray_define_segments import SegmentWireMap
+
+# =============================================================================
+
+
+def yield_downstream_nodes(conn, node_pkey):
```
litghost, line 60: Why is this specialized for prjuray?  Should be the same between prjxray and prjuray?

xc/common/utils/prjuray_form_channels.py
```diff
@@ -0,0 +1,481 @@
...
+            )
+        )
+
+    write_cur.execute("""COMMIT TRANSACTION""")
+
+
+def identify_const_nodes(conn):
+    """
+    Creates a new table for const nodes, identifies them and inserts their
+    pkeys along with const signal name to the table.
```
litghost, line 230: This docstring is insufficient.  In generally all of the docstrings in this file need to be expanded.

xc/xcu/CMakeLists.txt
mkurc-ant: Right. I've fixed that and bumped `prjuray-tools` to the version that has `setup.py`

xc/common/utils/prjuray_timing.py
mkurc-ant: I'm not entirely happy with that but I didn't wanted any U-Ray scripts depend on X-Ray stuff.

The content of this file is copied from `third_party/prjxray/prjxray/timing.py`. This code is not present in `prjuray-tools` so I couldn't import it from there. Since it is not directly related to X-Ray/U-Ray database but rather to how it is handled in arch-defs I decided to put it in arch-defs, at least for U-Ray import for now.

It may be worth to rename this to eg. `timing.py` and make X-Ray import use it as well since it is common to both archs. Any thoughts on that?

xc/common/utils/prjuray_form_channels.py
mkurc-ant: The function is only used in `classify_const_nodes` which is prjuray specific so it is not needed in prjxray. Even though that it would work for xray data in the same way.

xc/common/utils/prjuray_form_channels.py
mkurc-ant: I've updated docstrings. Please point me out where else you find them to be insufficient.

litghost:
This PR needs a rebase

xc/common/utils/prjuray_timing.py
litghost: We don't have a uray routing timing model just yet, so can we stub this out, rather than assuming the 7-series and US+ route timing model will be the same?

xc/common/utils/prjuray_form_channels.py
```diff
@@ -0,0 +1,516 @@
...
+from prjxray_db_cache import DatabaseCache
+
+from prjuray_define_segments import SegmentWireMap
+
+# =============================================================================
+
+
+def yield_downstream_nodes(conn, node_pkey):
+    """
+    For the given node pkey yields pkeys of all its immediate downstream nodes.
```
litghost, line 62: This comment does mention that pseudo pips are avoided, etc.  This function comment doesn't provide enough context of what and why this function is doing.

xc/common/utils/prjuray_form_channels.py
```diff
@@ -0,0 +1,516 @@
...
+                wire_in_tile_pkey,
+            )
+        )
+
+    write_cur.execute("""COMMIT TRANSACTION""")
+
+
+def identify_const_nodes(conn):
+    """
+    Creates a new table for const nodes named "const _nodes". Identifies them
```
litghost, line 233: "const _nodes" -> "const_nodes"?

litghost: Rather than add a table, maybe should add a node property on the nodes table?

litghost (commented):
Comments in

Issue #1656, opened 2020-09-04
Generated FASM files get appended to existing ones


acomodi:
In case build files for a specific test are already present, and another build is performed, the `top.merged.fasm`, which is used to generate the bitstream, is not overwritten, but instead, the new fasm data of the second build gets appended. This obviously causes FASM collisions when generating the frames.

The problematic lines are the following: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/b3864ae63d668fc7d321b94ca946841e80dd0bb6/common/cmake/devices.cmake#L1058-L1073

Could be solved by removing the `arch.merged.fasm` file if present, before running the FASM files merge procedure.

@andrewb1999 I believe this was an addition for the partial regions, therefore we have multiple FASM files that need to be collapsed in one only for the bitsream generation step. Do you think the above solution is valid and does not break the partial regions stuff?

andrewb1999:
Yes, this change was made for partial regions work.  Removing the merged.fasm if it exists should work fine.

PR #1660, opened 2020-09-11
devices.cmake: remove OUT_MERGED_FASM before appending fasm files

acomodi:

This PR fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1656

### DIFF
--- a/common/cmake/devices.cmake
+++ b/common/cmake/devices.cmake
@@ -1061,14 +1061,14 @@ function(ADD_BITSTREAM_TARGET)
         DEPENDS ${BITSTREAM_DEPS}
         COMMAND
           ${CMAKE_COMMAND} -E make_directory ${OUT_LOCAL}
-        COMMAND cat ${ALL_OUT_FASM} >> ${OUT_FASM_MERGED}
+        COMMAND cat ${ALL_OUT_FASM} > ${OUT_FASM_MERGED}
         COMMAND ${FASM_TO_BIT_CMD_FOR_TARGET_LIST}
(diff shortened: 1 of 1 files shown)


common/cmake/devices.cmake
```diff
@@ -1061,13 +1061,15 @@ function(ADD_BITSTREAM_TARGET)
         DEPENDS ${BITSTREAM_DEPS}
         COMMAND
           ${CMAKE_COMMAND} -E make_directory ${OUT_LOCAL}
+        COMMAND rm -f ${OUT_FASM_MERGED}
         COMMAND cat ${ALL_OUT_FASM} >> ${OUT_FASM_MERGED}
```
litghost, line 1065: Could just change `>>` to `>` and it would do the same thing without needing the `rm`?

acomodi: Unsure if this would break the multiple fasm files that get output with cat. Most probably you are right and we could just use >

litghost: It would break things if there were multiple targets that output to the same file, but that strategy would be a bug in the first place because if `--jobs` is greater than 1, then there would be a race condition.

Bottom line, just use `>` and drop the `rm -f`.  Even if `>>` was required, now that there is a `rm -f` first, the `>>` no longer does anything compared with just `>`.

acomodi: Ok done

andrewb1999 (approved):
LGTM

PR #1664, opened 2020-09-15
ice40: io constr: do not emit constraint that is not in packed netlist

acomodi:

This PR is to fix the IO constraints that are incorrectly emitted when the packed netlist (top.net) does not contain the requested IOs.

This fixes the issue found when running all_ice40 tests after the [new master+wip VTR upgrade](https://github.com/SymbiFlow/vtr-verilog-to-routing/pull/545) and described in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1541

### DIFF
--- a/ice40/icestorm.cmake
+++ b/ice40/icestorm.cmake
@@ -57,7 +57,8 @@ function(icestorm_setup)
     \${PYTHON3} \${PLACE_TOOL} \
     --map \${PINMAP} \
     --blif \${OUT_EBLIF} \
-    --pcf \${INPUT_IO_FILE}"
+    --pcf \${INPUT_IO_FILE} \
+    --net \${OUT_NET}"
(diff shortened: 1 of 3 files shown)


mithro:
This looks like a reasonable fix. Any idea why CI is failing?

acomodi:
I think I know how to fix the iCE40 CI error, but I am not sure about the XC7 failure. The changes in this PR should be unrelated and should not affect XC7 builds.

I'll investigate that. Turns out though that https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1541 got actually merged, but probably it would be good to add this fix as well.

acomodi:
Closing, superseded by https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1666

PR #1666, opened 2020-09-16
Update VTR

acomodi:
This PR contains the following:

- update of VTR version
- fixes in CMake to have correct parameters for the new VTR
- fixes to ice40 place constraints from https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1664

PR #1667, opened 2020-09-16
Transition to use extended map lookahead

acomodi:
This PR builds on top of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1666 and is intended to test the extended lookahead map.

### DIFF
--- a/xc/common/utils/prjxray_edge_library.py
+++ b/xc/common/utils/prjxray_edge_library.py
@@ -2255,7 +2255,8 @@ def create_edges(args):
                 tile_name = grid.tilename_at_loc(loc)
 
                 if tile_name in synth_tiles['tiles']:
-                    for pin in synth_tiles['tiles'][tile_name]['pins']:
+                    synth_tile = synth_tiles['tiles'][tile_name]
+                    for pin in synth_tile['pins']:
                         if pin['port_type'] not in ['input', 'output']:
                             continue
 
@@ -2263,14 +2264,38 @@ def create_edges(args):
                             tile_name, gridinfo.tile_type, pin['wire']
                         )
 
-                        if pin['port_type'] == 'input':
-                            # This track can output be used as a sink.
+                        is_input = pin['port_type'] == 'input'
+                        is_output = pin['port_type'] == 'output'
+
+                        assert is_input or is_output, pin
+
+                        if is_input:
+                            # This track can be used as a sink.
                             input_only_nodes |= set((node_pkey, ))
-                        elif pin['port_type'] == 'output':
-                            # This track can output be used as a src.
+                        elif is_output:
+                            # This track can be used as a src.
                             output_only_nodes |= set((node_pkey, ))
-                        else:
-                            assert False, pin
+
+                        # Adding all wires outside of the ROI relative to this synth tile
+                        # to the input/output only nodes, so that the corresponding nodes
+                        # have only outgoing or incoming edges
+                        for tile, wires in synth_tile['wires_outside_roi'
+                                                      ].items():
+                            tile_type = grid.gridinfo_at_tilename(
+                                tile
+                            ).tile_type
+
+                            for wire in wires:
+                                _, _, _, node_pkey = find_wire(
+                                    tile, tile_type, wire
+                                )
+
+                                if is_input:
+                                    # This track can be used as a sink.
+                                    input_only_nodes |= set((node_pkey, ))
+                                elif is_output:
+                                    # This track can be used as a src.
+                                    output_only_nodes |= set((node_pkey, ))
 
         create_and_insert_edges(
             db=db,
--- a/xc/common/utils/prjxray_edge_library.py
+++ b/xc/common/utils/prjxray_edge_library.py
@@ -2271,10 +2271,10 @@ def create_edges(args):
 
                         if is_input:
                             # This track can be used as a sink.
-                            input_only_nodes |= set((node_pkey, ))
+                            input_only_nodes.add(node_pkey)
                         elif is_output:
                             # This track can be used as a src.
-                            output_only_nodes |= set((node_pkey, ))
+                            output_only_nodes.add(node_pkey)
 
                         # Adding all wires outside of the ROI relative to this synth tile
                         # to the input/output only nodes, so that the corresponding nodes
@@ -2292,10 +2292,10 @@ def create_edges(args):
 
                                 if is_input:
                                     # This track can be used as a sink.
-                                    input_only_nodes |= set((node_pkey, ))
+                                    input_only_nodes.add(node_pkey)
                                 elif is_output:
                                     # This track can be used as a src.
-                                    output_only_nodes |= set((node_pkey, ))
+                                    output_only_nodes.add(node_pkey)
 
         create_and_insert_edges(
             db=db,
--- a/xc/common/cmake/arch_define.cmake
+++ b/xc/common/cmake/arch_define.cmake
@@ -66,7 +66,7 @@ function(ADD_XC_ARCH_DEFINE)
       --clock_modeling route \
       --place_delta_delay_matrix_calculation_method dijkstra \
       --place_delay_model delta_override \
-      --router_lookahead connection_box_map \
+      --router_lookahead extended_map \
       --check_route quick \
       --strict_checks off \
       --allow_dangling_combinational_nodes on \
--- a/xc/common/cmake/device_define.cmake
+++ b/xc/common/cmake/device_define.cmake
@@ -416,7 +416,7 @@ function(ADD_XC_DEVICE_DEFINE)
         --clock_modeling route
         --place_delay_model delta_override
         --place_delta_delay_matrix_calculation_method dijkstra
-        --router_lookahead connection_box_map
+        --router_lookahead extended_map
         --disable_errors check_unbuffered_edges:check_route:check_place
         --suppress_warnings sum_pin_class:check_unbuffered_edges:load_rr_indexed_data_T_values:check_rr_node:trans_per_R
         --route_chan_width 500
--- a/xc/common/utils/prjxray_create_synth_tiles.py
+++ b/xc/common/utils/prjxray_create_synth_tiles.py
@@ -214,6 +214,7 @@ def main():
     with DatabaseCache(args.connection_database, read_only=True) as conn:
         tile_in_use = set()
         num_synth_tiles = 0
+
         for roi, j in rois.items():
             if args.overlay:
                 synth_tiles['info'].append(j['info'])
@@ -261,6 +262,7 @@ def main():
                         'pins': [],
                         'loc': vpr_loc,
                         'tile_name': tile_name,
+                        'wires_outside_roi': {},
                     }
                     num_synth_tiles += 1
                     tile_pin_count[tile] = 0
@@ -281,6 +283,18 @@ def main():
                             tile_pin_count[tile],
                     }
                 )
+
+                if 'wires_outside_roi' in port:
+                    outside_roi = synth_tiles['tiles'][tile
+                                                       ]['wires_outside_roi']
+
+                    for tile_wire in port['wires_outside_roi']:
+
+                        tile_name, wire_name = tile_wire.split('/')
+                        if tile_name in outside_roi.keys():
+                            outside_roi[tile_name].append(wire_name)
+                        else:
+                            outside_roi[tile_name] = [wire_name]
 
                 tile_pin_count[tile] += 1
 


litghost:
At a glance it looks like the 50 and 100 lookahead computation times are lower with the extended map vs the connection box.

litghost:
Looks like a conflict on `environment.yml` when I merged the sv2v changes.

acomodi:
> At a glance it looks like the 50 and 100 lookahead computation times are lower with the extended map vs the connection box.

@litghost Indeed, I have seen that it is more than half of the time to compute the extended map w.r.t. connection box map.

litghost:
> > At a glance it looks like the 50 and 100 lookahead computation times are lower with the extended map vs the connection box.
> 
> @litghost Indeed, I have seen that it is almost half of the time to compute the extended map w.r.t. connection box map.

From the results, it looks like most routing times are down, except for baselitex arty 100.  Also the critical path delays for most designs meet the constraint (CPD < 20 ns).  Let's wait for the results from https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1666, but it looks like the extended map lookahead is definitely good enough to replace the connection box lookahead.

acomodi:
It seems like there an issue with fasm2bels or VPR on a ROI target (picosoc_basys3) when using the extended lookahead map, investigating:

```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/tests/9-soc/picosoc && diff -u /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/tests/9-soc/picosoc/picosoc_basys3/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/basys3_demo.bit.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/tests/9-soc/picosoc/picosoc_basys3/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_picosoc_basys3_vivado.bit.fasm
--- /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/tests/9-soc/picosoc/picosoc_basys3/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/basys3_demo.bit.fasm	2020-09-16 10:44:16.031415533 -0700
+++ /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/tests/9-soc/picosoc/picosoc_basys3/artix7-xc7a50t-basys3-roi-virt-xc7a50t-basys3-test/design_picosoc_basys3_vivado.bit.fasm	2020-09-16 10:48:59.694717134 -0700
@@ -65533,8 +65533,6 @@

 INT_R_X31Y62.LV0.LH12

-INT_R_X31Y63.LV0.LH12
-
 INT_R_X31Y64.LV0.LH12

 INT_R_X31Y65.LV0.LH12
@@ -65549,8 +65547,6 @@

 INT_R_X31Y80.LV0.LV18

-INT_R_X31Y81.LV0.LV18
-
 INT_R_X31Y82.LV0.LV18

 INT_R_X31Y83.LV0.LV18
@@ -65565,8 +65561,6 @@

 INT_R_X31Y98.LV0.LV18

-INT_R_X31Y99.LV0.LV18
-
 INT_R_X31Y100.LVB0.LV18

 INT_R_X31Y101.LV0.LV18
@@ -65584,8 +65578,6 @@

 INT_R_X31Y116.LVB0.LV18

-INT_R_X31Y117.LV0.LV18
-
 INT_R_X31Y119.NN6BEG3.LV18

 INT_R_X31Y121.LV0.LV18
@@ -65601,8 +65593,6 @@

 INT_R_X31Y130.NW6BEG3.LV18

-INT_R_X31Y135.LH0.LV18
-
 INT_R_X31Y139.WW4BEG3.LV18

 INT_R_X35Y71.SE6BEG0.LV0
@@ -65705,9 +65695,7 @@
 INT_R_X43Y43.NR1BEG0.LOGIC_OUTS18

 INT_R_X43Y44.LV0.NR1END0
-INT_R_X43Y44.NR1BEG0.LOGIC_OUTS18

-INT_R_X43Y45.LV0.NR1END0
 INT_R_X43Y45.NR1BEG0.LOGIC_OUTS18
 INT_R_X43Y45.SS6BEG0.SE6END0

@@ -65731,8 +65719,6 @@

 INT_R_X43Y62.LH0.LV18

-INT_R_X43Y63.LH0.LV18
-
 INT_R_X43Y64.LH0.LV18

 INT_R_X43Y65.LH0.LV18
@@ -66047,9 +66033,7 @@

 RIOB33_X43Y43.IOB_Y0.IN_TERM.NONE
 RIOB33_X43Y43.IOB_Y0.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL.SLEW.FAST
-RIOB33_X43Y43.IOB_Y0.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL_SSTL135_SSTL15.IN_ONLY
-RIOB33_X43Y43.IOB_Y0.LVCMOS25_LVCMOS33_LVTTL.IN
-RIOB33_X43Y43.IOB_Y0.PULLTYPE.NONE
+RIOB33_X43Y43.IOB_Y0.PULLTYPE.PULLDOWN
 RIOB33_X43Y43.IOB_Y1.IN_TERM.NONE
 RIOB33_X43Y43.IOB_Y1.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL.SLEW.FAST
 RIOB33_X43Y43.IOB_Y1.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL_SSTL135_SSTL15.IN_ONLY
@@ -66205,7 +66189,6 @@
 RIOI3_TBYTESRC_X43Y43.ILOGIC_Y0.IFF.SRTYPE.ASYNC
 RIOI3_TBYTESRC_X43Y43.ILOGIC_Y0.ISERDES.MODE.MASTER
 RIOI3_TBYTESRC_X43Y43.ILOGIC_Y0.ISERDES.NUM_CE.N1
-RIOI3_TBYTESRC_X43Y43.ILOGIC_Y0.ZINV_D
 RIOI3_TBYTESRC_X43Y43.ILOGIC_Y1.IDELMUXE3.P1
 RIOI3_TBYTESRC_X43Y43.ILOGIC_Y1.IFF.SRTYPE.ASYNC
 RIOI3_TBYTESRC_X43Y43.ILOGIC_Y1.ISERDES.MODE.MASTER
```

litghost:
I would assume this is a fasm2bels issue around constant routing, but hard to say for sure.

acomodi:
@litghost Not sure if this makes sense, but I have an idea why this is happening.

Basically, there is a connection within the ROI on an interconnect tile that seem to trigger make_routes to use a LH12 wire that gets outside of the ROI (the select white wire that gets to the bottom right in the image below that represents the broken dcp file).
This is a pretty unfortunate and improbable corner case, as the destination point of that  wire coincides with the destination of the missing wire that connects the missing IOB to the ROI.

In other words, make_routes correctly finds a route that connects two pips, but it happens that that is an inexistent route.
To be more precise, this would certainly lead to a non-working bitstream, as the long wire (LH12)  would have a conflict at its two ends (visible in the last image).


Screenshot from 2020-09-17 18-54-23


Screenshot from 2020-09-17 19-02-10

litghost:
> @litghost Not sure if this makes sense, but I have an idea why this is happening.
> 
> Basically, there is a connection within the ROI on an interconnect tile that seem to trigger make_routes to use a LH12 wire that gets outside of the ROI (the select white wire that gets to the bottom right in the image below that represents the broken dcp file).
> This is a pretty unfortunate and improbable corner case, as the destination point of that wire coincides with the destination of the missing wire that connects the missing IOB to the ROI.
> 
> In other words, make_routes correctly finds a route that connects two pips, but it happens that that is an inexistent route.
> To be more precise, this would certainly lead to a non-working bitstream, as the long wire (LH12) would have a conflict at its two ends (visible in the last image).
> 
> Screenshot from 2020-09-17 18-54-23
> 
> Screenshot from 2020-09-17 19-02-10

`make_routes` should never generate a route that doesn't exist in the underlying bitstream.  Given what you are writing, it should like the bitstream from VPR was broken?

acomodi:
@litghost let me rephrase.

The LH12 wire has three endpoints, one inside the ROI and the other two outside of the ROI as visible in the second image.

The problem is that in the output fasm from VPR both of the endpoints pips are enabled (one, the leftmost, is produced by VPR and the second, the rightmost, is part of the harness), but they are part of two different nets, but of course VPR does not know about that as it routes only within the ROI.

IMO, `make_routes` saw the leftmost pip enabled and saw also the middle endpoint of the LH12 enabled, and generated the (inexistent) connection.

litghost:
> @litghost let me rephrase.
> 
> The LH12 wire has three endpoints, one inside the ROI and the other two outside of the ROI as visible in the second image.
> 
> The problem is that in the output fasm from VPR both of the endpoints pips are enabled (one, the leftmost, is produced by VPR and the second, the rightmost, is part of the harness), but they are part of two different nets, but of course VPR does not know about that as it routes only within the ROI.
> 
> IMO, `make_routes` saw the leftmost pip enabled and saw also the middle endpoint of the LH12 enabled, and generated the (inexistent) connection.

Ok, so there are two bugs.  The first is that the routing import for the ROI shouldn't excluded the node containing the LH12 wire such that no input edges went to that CHAN except for the edge from the synthetic tile.

The second bug is a missing sanity check in fasm2bels about invalid bitstream that has a shorted wire.  This is mostly an additional error check (e.g. missing feature), rather than a defect in fasm2bels.

acomodi:
@litghost All right, IMO as this is a corner case affecting only ROI targets, solving these bugs is a lower priority task.

As I am unsure how long it would require to fix both the "bugs", it may would be better to file GH issues on both fasm2bels and here, and solve this in a second moment.

litghost:
> @litghost All right, IMO as this is a corner case affecting only ROI targets, solving these bugs is a lower priority task.
> 
> As I am unsure how long it would require to fix both the "bugs", it may would be better to file GH issues on both fasm2bels and here, and solve this in a second moment, and for the time being disable the diff_fasm test on the `picosoc_basys3` target.

I completely disagree.  If VPR is generating invalid bitstreams, we need to fix that.

acomodi:
> I completely disagree. If VPR is generating invalid bitstreams, we need to fix that.

Ok, I'll get this fixed on the VPR side, during the RR graph generation for ROI targets as this is the issue that is generating invalid bitstreams

acomodi:
@litghost I am quite unsure on how to proceed in solving the rr graph issue and eliminate the nodes that can escape the ROI, generating shorted connections with pre-existing harness routing.

With https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1667/commits/d0db3dcea355c2765038469657f18d4bd6ea13e5, I have tried to delete nodes that have edges with nodes outside the ROI, but I think that my implementation (and I believe also the idea) is quite wrong, in fact I CI went red due to the fact that I deleted nodes are unrelated to the problem I am trying to solve:

```
## Initializing router criticalities took 0.07 seconds (max_rss 711.1 MiB, delta_rss +0.0 MiB)
Warning 64: No routing path for connection to sink_rr 102804, retrying with full device bounding box
Warning 65: No routing path for connection to sink_rr 189328, retrying with full device bounding box
Cannot route from SYN-GND[0].GND[0] (RR node: 103530 type: SOURCE location: (30,27) class: 0 capacity: 1) to BLK-TL-BRAM_L[0].BRAM_FIFO18_ADDRBWRADDR0[0] (RR node: 189328 type: SINK location: (20,53) class: 18 capacity: 1) -- no possible path
Failed to route connection from '$false' to 'murax.system_ram.ram_symbol1.0.0.0' for net '$false' (#0)
```

Do you have some suggestions on how to proceed?

xc/common/utils/prjxray_routing_import.py
```diff
@@ -733,6 +734,68 @@ def import_graph_nodes(conn, graph, node_mapping, connection_box_map):
             assert False, side
 
 
+def remove_graph_nodes_outside_roi(conn, node_mapping, roi):
```
litghost, line 737: We shouldn't remove graph nodes outside the ROI.  There is already (potentially broken) logic for not adding edges to nodes used in the ROI.

litghost: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/6e6e8c7ce9e6f2f553ebf6610714e9b3b9e7d20c/xc/common/utils/prjxray_edge_library.py#L1120-L1125

litghost: My suspicion is that bidirectional edges are bypassing this logic, because the src can become the destination when creating the reverse edge.

acomodi: > We shouldn't remove graph nodes outside the ROI

Indeed, the naming of this function is actually wrong. The intent with this was to remove graph_nodes within the ROI that can potentially have connections with nodes outside of the ROI.

acomodi:
With https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1667/commits/e72d91340470ed19a6b5a10840d3d0ba83631675, the Vendor Tools failure seems to be fixed, but I am unsure whether this is because the corner case we hit prior to the commit does not show up due to changes to the RR graph that yield different routing.

One thing I see from the QoR results though is a degradation of CPD and runtime for the baselitex tests, which is strange due to the fact that the changes in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1667/commits/e72d91340470ed19a6b5a10840d3d0ba83631675 should affect ROI-only tests.

xc/common/utils/prjxray_edge_library.py
```diff
@@ -1149,6 +1149,91 @@ def make_connection(
...
+
+    for (track_pkey, ) in c.execute("""
+SELECT
+  pkey
+FROM
+  track
+WHERE
+  alive;"""):
+
+        write_cur.execute(
```
litghost, line 1177: Pull the constant source retrieval out of the loop?

xc/common/utils/prjxray_edge_library.py
```diff
@@ -1149,6 +1149,91 @@ def make_connection(
...
+        (dest_track_pkey, ) = write_cur.fetchone()
+        write_cur.execute(
+            """
+SELECT grid_x, grid_y FROM phy_tile WHERE pkey = (
+SELECT canon_phy_tile_pkey FROM track WHERE pkey = ?);""", (dest_track_pkey, )
+        )
+
+        tile_info = write_cur.fetchone()
+
+        if tile_info:
```
litghost, line 1210: We should be removing edges, not nodes.  Just because a node has an edge that is outside of the ROI doesn't mean it it should be killed.

If an **edge** (e.g. pip) exists outside of the ROI it should be removed, and if an **edge** connects **to** an output only node, or an **edge** connects **from** an input only node it should be removed.  The alive check will then kill any nodes that are unreachable or are dead ends.

acomodi: As far as I have seen, the edge to and from the IO only nodes is already taken care of.

The remaining problem would be removing the edges existing outside of the ROI. I thought about this, and it probably does not completely solve the issue of producing shorted connections.

For instance, an LH12 wire that has three endpoints can have two of the endpoints within the ROI and the last one outside of the ROI. This means that even though the edges present outside of the ROI are removed, the LH12 node will still be present as it has valid graph edges, therefore needs to be marked as alive, with the same potential of generating shorted wire.

xc/common/utils/prjxray_edge_library.py
acomodi: Done

xc/common/utils/prjxray_edge_library.py
litghost: > For instance, an LH12 wire that has three endpoints can have two of the endpoints within the ROI and the last one outside of the ROI. This means that even though the edges present outside of the ROI are removed, the LH12 node will still be present as it has valid graph edges, therefore needs to be marked as alive, with the same potential of generating shorted wire.

I've already talked about this.  In the case of the LH12, yes there are multiple endpoints.  However the entire node should be marked as "input only" (e.g. this is connected in the ROI to a source), or "output only" (e.g. this is connected in the ROI to a sink).  An "input only" node should only have edges that come from the node (except for the incoming edge from the SYN-IN).  An "output only" node should only have edges that come to the node (except for the outcoming edge to the SYN-OUT).  There can be no shorts if this is actually done.

acomodi: I am still kind of confused about this.

The problematic nodes that cannot be used are not related to SYN-IN, SYN-OUT or constant sources.
In `mark_track_liveness`, if a track is present in the "input_only_nodes" or "output_only_nodes" it will get added to the alive tracks, so adding the problematic nodes to the input only set or output only set won't get that node removed.

What I could do, is to have the edges creation process to avoid adding edges if the node is an "input only" node and the edge connects from that node (or vice-versa). The problem though is that I do not understand how to detect which are the problematic nodes prior to the creation of the graph edges.

Moreover, I do not yet understand why we can't just kill a track that is not related to synth nodes or constant source nodes and has edges outside the ROI, as it happens in the current implementation. After all, the node will still be killed if it has no output edges or no input edges during `mark_track_liveness` and it was marked as input- or output-only node.

litghost: Let's take a step back.  I thought the original bug was because a node that is part of the ROI harness was being include in the routing graph, and VPR used it for general routing.  All nodes that are part of the ROI harness should be modified such that the routing graph doesn't allow them to be used in a way that would break the ROI harness.  In the original example, I believe that the failure was a short was created because a LH12 wire had two drivers, one from VPR and one from the ROI harness?  If the relevant node was marked as either input_only or output_only, the relevant edges for that node would be excluded to prevent a misuse/short.  Am I missing something in that description?

acomodi: > I thought the original bug was because a node that is part of the ROI harness was being included in the routing graph, and VPR used it for general routing

As far as I understood yes, this is the main issue. The RR graph includes nodes that are being used in the harness (in this  particular situation, the node is `INT_R_X19Y135/LH12`, which can span out of the ROI).

>  I believe that the failure was a short was created because a LH12 wire had two drivers, one from VPR and one from the ROI harness?

Correct, the above-mentioned wire is used by VPR as a "bridge" to connect two nodes in the `INT_R_X19Y135` tile, while, the other end of the wire is used by the ROI harness and is connected to drive a SYN-IN.

>  If the relevant node was marked as either input_only or output_only, the relevant edges for that node would be excluded to prevent a misuse/short

I think that this can be done, but we would require information on which nodes are used in the ROI harness. I see that we can get this information from the design.json of the harness, which has a field called `wires_outside_roi`, which could potentially be used to get the wires that need to be marked as input_only or output_only.
The problem here is that the same wire, can be named differently in the `design.json`. To be more precise, the `INT_R_X19Y135/LH12` is not present in the json, but instead, `INT_R_X25Y135/LH6` is. Now, the latter is not a node, but a pip in the INT tile. The two though are the exact same wire.

In conclusion, I believe this issue could be "easily" solved by using the information on the wires outside the roi present in the json, but we would need to provide correct naming of the nodes used in the harness.

litghost: > I see that we can get this information from the design.json of the harness, which has a field called wires_outside_roi, which could potentially be used to get the wires that need to be marked as input_only or output_only.

Yes, and I believe it was intended to be used for that purpose.  The connection database can translate the wire -> node.

litghost (commented):
Code looks good, let's wait for results and make sure everything looks okay.

acomodi:
Preliminary analysis of the QoR results among the continuous CI on master and this PR's CI is the following:

Average route time increase: 17%
Average CPD increase: 0.1%

Good news is that CI builds do complete much faster compared to baseline. For instance, the longest current CI (200T vendor tools) takes on master ~10h20m, while, using the extended map, it takes ~6h30m.

A similar reduction appears in the install package size due to a reduction in the size of the lookahead:

- XC7 (compressed)
  - Baseline: 906 MB
  - Changes: 658 MB

- XC7 200T (compressed)
  - Baseline: 548 MB
  - Changes: 400 MB

[Here](https://github.com/SymbiFlow/symbiflow-arch-defs/files/5290985/xc7_qor_comparisons.zip) is the csv file that includes both the CI results of baseline and the changes. The same file is imported in sheets [here](https://docs.google.com/spreadsheets/d/1mh0ILIAlLpcp9qIETPV7bHA8FvCtVaPEnN1dqTDWdog/edit?usp=sharing)

mithro:
@acomodi How does CI run time complete faster while route time takes longer? It is because the map calculation is significantly faster?

acomodi:
@mithro Indeed, the main reason of faster CI is the Lookahead map calculation, which for what I have seen with local tests, takes up to half of the time required by the current master

litghost:
@acomodi Great work on this!  Exciting to see this merged.  Next steps:

1. Determine how to fix workaround in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1564/files#diff-0fa89048843d13236cff348a4ce8edc2L390-R394 .
2. Determine if penalty cost on edges is still as needed as before.  If so, implement the edge splitting work that was started before.

Sounds good?

PR #1671, opened 2020-09-23
Begin integration of new fasm2bels with interchange support

litghost:
This adds CMake logic to invoke the interchange and create DCPs with RapidWright, but leaves that logic dormant unless RAPIDWRIGHT_PATH is supplied pointing to an version of RapidWright with the interchange support present.

### DIFF
--- a/xc/common/cmake/vivado.cmake
+++ b/xc/common/cmake/vivado.cmake
@@ -492,21 +492,39 @@ function(PREPARE_RAPIDWRIGHT)
   #         location of the invoke_rapidwright.sh script which can be used to
   #         invoke RapidWright.
   add_custom_target(rapidwright)
+  set_target_properties(rapidwright PROPERTIES RAPIDWRIGHT_INSTALLED FALSE)
+
+  set(HAVE_RAPIDWRIGHT FALSE)
+  set(HAVE_JAVA FALSE)
+  set(HAVE_URAY_VIVADO FALSE)
   if(EXISTS "${RAPIDWRIGHT_PATH}" AND EXISTS ${RAPIDWRIGHT_PATH}/interchange AND EXISTS ${RAPIDWRIGHT_PATH}/scripts/invoke_rapidwright.sh)
+    set(HAVE_RAPIDWRIGHT TRUE)
+  endif()
+
+  if(${HAVE_RAPIDWRIGHT})
     find_program(JAVA java)
     if(EXISTS ${JAVA})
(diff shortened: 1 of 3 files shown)


litghost:
FYI @mithro

xc/common/cmake/vivado.cmake
```diff
@@ -464,3 +478,199 @@ function(ADD_VIVADO_PNR_TARGET)
...
+  set(HAVE_URAY_VIVADO FALSE)
+  if(EXISTS "${RAPIDWRIGHT_PATH}" AND EXISTS ${RAPIDWRIGHT_PATH}/interchange AND EXISTS ${RAPIDWRIGHT_PATH}/scripts/invoke_rapidwright.sh)
+    set(HAVE_RAPIDWRIGHT TRUE)
+  endif()
+
+  if(${HAVE_RAPIDWRIGHT})
+    find_program(JAVA java)
+    if(EXISTS ${JAVA})
+      set(HAVE_JAVA TRUE)
+    endif()
```
acomodi, line 508: Does java need to be pre-existing?

If so, wouldn't it be better (if doable) to have a Java package in conda, so that we can be sure the correct version is used?

acomodi (approved):
LGTM apart from a question on Java

xc/common/cmake/vivado.cmake
```diff
@@ -464,3 +478,199 @@ function(ADD_VIVADO_PNR_TARGET)
...
+  endif()
+
+  if(DEFINED ENV{URAY_VIVADO_SETTINGS} AND EXISTS "$ENV{URAY_VIVADO_SETTINGS}")
```
mkurc-ant, line 511: How `URAY_VIVADO_SETTINGS` is U-Ray specific? Is the same Vivado is used for xc7 targets?

xc/common/cmake/vivado.cmake
litghost: Because of make-env, java should be on the path, whether via conda or otherwise.  So from a CMake perspective, this is right.  Whether we should do an `apt-get install java` or `conda install java` is a different problem.

xc/common/cmake/vivado.cmake
litghost: RapidWright needs a more recent version of Vivado, and URAY_VIVADO_SETTINGS currently conforms to the minimum Vivado requirement.

PR #1673, opened 2020-09-23
Integrate interchange

litghost:


mithro (approved):
LGTM if CI is happy.

environment.yml
```diff
@@ -2,6 +2,7 @@ name: symbiflow_arch_def_base
 channels:
   - defaults
   - symbiflow
+  - conda-forge
```
acomodi, line 5: Just to mention this, I recall there was an effort to avoid using the conda-forge channel as it led to possible conflicts among packages.

I think there is no other way though to have the capnproto-java package installed though, so I guess this is fine.

acomodi (approved):
LGTM.

I have one question related to the vendor tools though: now that we are transitioning to use the interchange instead of the previous flow of fasm2bels, how are we going to test whether everything still works fine with the previous method that uses Vivado?
In other words, if the interchange flow passes, can we assume that the vivado flow passes as well for the same test?

I see a failure in the iddr_oddr tests in vendor tools:
```
ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester0.iddr_wrapper.the_iddrLIOI3_X0Y25_ILOGIC_X0Y26_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester1.iddr_wrapper.the_iddrLIOI3_X0Y27_ILOGIC_X0Y27_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester2.iddr_wrapper.the_iddrLIOI3_X0Y23_ILOGIC_X0Y23_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester3.iddr_wrapper.the_iddrLIOI3_X0Y21_ILOGIC_X0Y21_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
```

Probably an `IS_INVERTED` parameter needs to be added in fasm2bels for those tests.

litghost:
> I see a failure in the iddr_oddr tests in vendor tools:
> 
> ```
> ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester0.iddr_wrapper.the_iddrLIOI3_X0Y25_ILOGIC_X0Y26_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
> ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester1.iddr_wrapper.the_iddrLIOI3_X0Y27_ILOGIC_X0Y27_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
> ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester2.iddr_wrapper.the_iddrLIOI3_X0Y23_ILOGIC_X0Y23_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
> ERROR: [DRC REQP-76] c_cb_single_clk_inversion: On IDDR_2CLK cell tester3.iddr_wrapper.the_iddrLIOI3_X0Y21_ILOGIC_X0Y21_IDDR_2CLK the same clock signal drives the C and CB pins. Because of a hardware restriction, these pins need to have an opposite inversion with respect to one another.  Please add an inversion to one of these pins.
> ERROR: [Vivado 12-1345] Error(s) found during DRC. Bitgen not run.
> ```
> 
> Probably an `IS_INVERTED` parameter needs to be added in fasm2bels for those tests.

Ya, I fixed this in https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/25 , I need to update the submodule to get the fix.

litghost:
> I have one question related to the vendor tools though: now that we are transitioning to use the interchange instead of the previous flow of fasm2bels, how are we going to test whether everything still works fine with the previous method that uses Vivado?
> In other words, if the interchange flow passes, can we assume that the vivado flow passes as well for the same test?

To be clear, the interchange flow still uses Vivado, it just uses the interchange -> DCP instead of Verilog/XDC -> DCP -> bit.  This PR enables both the interchange and Verilog/XDC flows.  I believe in the future we'll change the Verilog/XDC -> DCP to just not do a diff fasm step, but it will still generate a bitstream.  This catches regressions in other areas.

environment.yml
litghost: If I add `capnproto` to the symbiflow channel we can avoid conda-forge.  The `capnproto` in `conda-forge` is pretty out of date anyways, so this is probably worthwhile anyways :/

litghost: FYI: https://github.com/SymbiFlow/conda-packages/pull/137

PR #1679, opened 2020-09-25
Fix missing CLKFBOUT_PHASE update

rw1nkler:
This PR:
- adds the `CLKFBOUT_PHASE` parameter to the list of updated PLL parameters.
- improves formatting in the `utils.tcl` script (previously mix of tabs and spaces)
- updates the ibex.patch

After the update, Ibex works on HW.

This issue has been tracked on our burndown list. After merging, I will update the list.


tmichalak (approved):
LGTM

PR #1687, opened 2020-10-01
Add FF-LUT pack patterns.

litghost:


mithro (commented):
Any idea how was this working without these?

litghost:
> Any idea how was this working without these?

These aren't required, they just help the packer to pack tighter, and results in some CPD improvements.

mkurc-ant (commented):
Shouldn't these pack patterns have specialized names? Like "**A**LUT5_TO_FF_FDSE" ?

litghost:
> Shouldn't these pack patterns have specialized names? Like "**A**LUT5_TO_FF_FDSE" ?

They specially should not be specialized.  pack pattern -> molecule happens pre cluster placement.

PR #1688, opened 2020-10-01
Fixup IOB definition

litghost:


xc/xc7/techmap/iobs.v
```diff
@@ -86,11 +79,67 @@ module OBUFTDS (
...
+endmodule
+
+module OBUFT (
+    (* iopad_external_pin *)
+    output O,
+    input I,
+    input T
+);
+    parameter CAPACITANCE = "DONT_CARE";
+    parameter integer DRIVE = 12;
```
mkurc-ant, line 121: If this one is `integer` then it should be like that in all other IOBs too

xc/xc7/yosys/synth.tcl
```diff
@@ -30,6 +30,10 @@ if { $::env(USE_ROI) == "TRUE" } {
     # Start flow after library reading
     synth_xilinx -vpr -flatten -abc9 -nosrl -noclkbuf -nodsp -iopad -nowidelut -run prepare:check
 }
+
+# Check that post-synthesis cells match libraries.
+hierarchy -check -auto-top
```
mkurc-ant, line 35: I think there shouldn't be `-auto-top` as this option may cause changing the top module.

xc/xc7/techmap/cells_sim.v
```diff
@@ -2012,3 +2012,12 @@ module PS7_VPR (
 
 endmodule
 
+module BANK();
```
mkurc-ant, line 2015: I'd add `(* keep *)` here (it should work on cell defintions) to make Yosys never prune `BANK` cells.

mkurc-ant (changes requested):
Looks ok, some minor changes requested.

xc/xc7/techmap/cells_sim.v
litghost: It already have KEEP?  See line 2017 and 2018?

litghost: Keep in mind that `(* keep *) is for cell instance declarations, not cell definitions.

xc/xc7/yosys/synth.tcl
litghost: Sure, fixed

xc/xc7/techmap/iobs.v
litghost: Removed the integer from that location to math the others.

litghost:
Fixes in from review.

litghost:
@mkurc-ant / @acomodi / @mithro This PR needs a review

acomodi (approved):
LGTM

PR #1689, opened 2020-10-01
Fix BRAM write enable ports.

litghost:


mithro:
I'm pretty sure these should be logically equivalent?

litghost:
> I'm pretty sure these should be logically equivalent?

Not at all.  WEA[0] is the lower byte, and WEA[1] is the upper byte.

mkurc-ant (approved):
Verified that with XC7 Memory resources user guide. LGTM

acomodi (approved):
LGTM, merging

xc/xc7/techmap/cells_map.v
```diff
@@ -1533,8 +1533,10 @@ end
   if(WRITE_WIDTH_A < 18) begin
       assign WEA_WIDE = {4{WEA[0]}};
   end else if(WRITE_WIDTH_A == 18) begin
```
mithro, line 1535: I think it would have been clearer if you ordered on the left hand side like;
```verilog
      assign WEA_WIDE[3] = WEA[1];
      assign WEA_WIDE[2] = WEA[0];
      assign WEA_WIDE[1] = WEA[1];
      assign WEA_WIDE[0] = WEA[0];
```

PR #1698, opened 2020-10-07
Add never prune idelayctrl

acomodi:
This PR enables the never_prune attribute to the IDELAYCTRL so that it does not get pruned by VPR if no output signal is provided.


litghost (approved):
LGTM

PR #1699, opened 2020-10-07
Generalize libxml2 package spec now that 2.9.10 is released.

litghost:
This allows using the libxml2 package from anaconda rather than from
symbiflow.

After this is merged, we can remove the libxml2 job from https://github.com/SymbiFlow/conda-packages

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -14,7 +14,8 @@ dependencies:
   - nodejs
   - libiconv
   - iverilog
-  - libxml2=2.9.10=20200629_180127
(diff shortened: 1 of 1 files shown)


mithro (approved):
LGTM!

Merge if green.

acomodi (approved):
LGTM

litghost:
> I see that XC7 took 10 hours to complete and the reason being the baselitex_arty100T which took ~30000 seconds.

My best guess is something we hit memory pressure and start paging.  Something that might be worth adding is a memory/cpu status print like we have on VTR and prjxray,

PR #1674, opened 2020-09-23
Change how CARRY4 modelling is handled

litghost:
This PR attempts to fix #1597 by change the strategy used to decongest SLICEL CARRY4 outputs.  Previously, all ALU CO uses were modeled as a `O ^ S`, but yosys's abc9 pass is now good enough to undo this change.  So rather than continue to try to fight abc9 as it gets better at merging logic, this PR uses a new CARRY4 congestion avoidance strategy.

Changes:
 - Yosys arith mapping using upstream CARRY4 mapping
 - New pass (implemented as a Python3 script) de-congests O + CO carry rows by using a LUT equation from the inputs
 - CARRY4_VPR is modelled as a blackbox instead of a whitebox to prevent abc9 from modifying/swapping outputs during cleanup optimization

PR #1705, opened 2020-10-08
Make sure core dumps are recorded

litghost:
Attempting to get a bead on segfaults seen in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1674


mithro (approved):
LGTM.

PR #1711, opened 2020-10-14
Fix unroutability issues with new VPR

acomodi:
The newest VPR has fixed an issue with IPIN/OPIN RR nodes for which multiple RR nodes were emitted for the same site pin, due to the possible multiple side locations of the pin.

Therefore, with the new VPR, we cannot count on the presence of multiple nodes for the same pin during the routing import, to create the correct connections between the interconnection nodes and the site nodes.

This PR fixes this issue, assigning the same node ID to the directional graph nodes allowing to have multiple CHAN nodes connected to an IPIN/OPIN.

The corresponding VPR issue is the following one: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1571.

This PR is expected to fail as with the current version of VPR, multiple nodes corresponding to the same site pin are emitted during the creation of the virtual RR graph.

### DIFF
--- a/xc/common/utils/prjxray_routing_import.py
+++ b/xc/common/utils/prjxray_routing_import.py
@@ -771,39 +771,38 @@ AND
             right_graph_node_pkey
         ) = result
 
-        side = node.loc.side
-        if side == tracks.Direction.LEFT:
-            assert left_graph_node_pkey is not None, (tile_type, pin_name)
-            node_mapping[left_graph_node_pkey] = node.id
+        # VPR emits only one node for each site pin, instead of one node for each
+        # side location of a pin.
+        #
+        # If the directional graph nodes are present for a specific tile wire,
+        # the same node ID is assigned to the directional graph node to prevent
+        # VPR failing to find a route.
+        if left_graph_node_pkey is not None:
+            node_mapping[left_graph_node_pkey] = (node.id, node.type)
 
             update_connection_box(
                 conn, graph, left_graph_node_pkey, node_idx, connection_box_map
             )
-        elif side == tracks.Direction.RIGHT:
-            assert right_graph_node_pkey is not None, (tile_type, pin_name)
-            node_mapping[right_graph_node_pkey] = node.id
+        if right_graph_node_pkey is not None:
+            node_mapping[right_graph_node_pkey] = (node.id, node.type)
 
             update_connection_box(
                 conn, graph, right_graph_node_pkey, node_idx,
                 connection_box_map
             )
-        elif side == tracks.Direction.TOP:
-            assert top_graph_node_pkey is not None, (tile_type, pin_name)
-            node_mapping[top_graph_node_pkey] = node.id
+        if top_graph_node_pkey is not None:
+            node_mapping[top_graph_node_pkey] = (node.id, node.type)
 
             update_connection_box(
                 conn, graph, top_graph_node_pkey, node_idx, connection_box_map
             )
-        elif side == tracks.Direction.BOTTOM:
-            assert bottom_graph_node_pkey is not None, (tile_type, pin_name)
-            node_mapping[bottom_graph_node_pkey] = node.id
+        if bottom_graph_node_pkey is not None:
+            node_mapping[bottom_graph_node_pkey] = (node.id, node.type)
 
             update_connection_box(
                 conn, graph, bottom_graph_node_pkey, node_idx,
                 connection_box_map
             )
-        else:
-            assert False, side
 
 
 def import_tracks(conn, alive_tracks, node_mapping, graph, default_segment_id):
@@ -871,15 +870,17 @@ SELECT grid_x, grid_y FROM phy_tile WHERE pkey = (
             y_high=y_high,
         )
         assert graph_node_pkey not in node_mapping
-        node_mapping[graph_node_pkey] = graph.add_track(
-            track=track,
-            segment_id=segment_id,
-            ptc=ptc,
-            timing=graph2.NodeTiming(
-                r=resistance,
-                c=capacitance,
-            ),
-            canonical_loc=canonical_loc
+        node_mapping[graph_node_pkey] = (
+            graph.add_track(
+                track=track,
+                segment_id=segment_id,
+                ptc=ptc,
+                timing=graph2.NodeTiming(
+                    r=resistance,
+                    c=capacitance,
+                ),
+                canonical_loc=canonical_loc
+            ), node_type
         )
 
 
@@ -995,9 +996,10 @@ WHERE
                 tuple(synth_tile['loc']), pin_name
             )
 
+            track_node_id, _ = node_mapping[track_node]
             if pin['port_type'] == 'input':
                 graph.add_edge(
-                    src_node=node_mapping[track_node],
+                    src_node=track_node_id,
                     sink_node=pin_node[0][0],
                     switch_id=delayless_switch,
                     name='synth_{}_{}'.format(tile_name, pin['wire']),
@@ -1005,7 +1007,7 @@ WHERE
             elif pin['port_type'] in ['VCC', 'GND', 'output']:
                 graph.add_edge(
                     src_node=pin_node[0][0],
-                    sink_node=node_mapping[track_node],
+                    sink_node=track_node_id,
                     switch_id=delayless_switch,
                     name='synth_{}_{}'.format(tile_name, pin['wire']),
                 )
@@ -1079,6 +1081,8 @@ def get_number_graph_edges(conn, graph, node_mapping):
     cur = conn.cursor()
     cur.execute("SELECT count() FROM graph_edge;" "")
 
+    nodes_set = set()
+
     for src_graph_node, dest_graph_node in cur.execute("""
 SELECT
   src_graph_node_pkey,
@@ -1091,6 +1095,20 @@ FROM
 
         if dest_graph_node not in node_mapping:
             continue
+
+        src_node, src_node_type = node_mapping[src_graph_node]
+        sink_node, sink_node_type = node_mapping[dest_graph_node]
+
+        pin_node_types = [graph2.NodeType.IPIN, graph2.NodeType.OPIN]
+
+        src_node_is_site_pin = src_node_type in pin_node_types
+        sink_node_is_site_pin = sink_node_type in pin_node_types
+
+        if src_node_is_site_pin ^ sink_node_is_site_pin:
+            if (src_node, sink_node) in nodes_set:
+                continue
+            else:
+                nodes_set.add((src_node, sink_node))
 
         num_edges += 1
 
@@ -1114,6 +1132,8 @@ def import_graph_edges(conn, graph, extra_features, node_mapping):
 
     switch_name_map = {}
 
+    nodes_set = set()
+
     print('{} Importing edges from database.'.format(now()))
     with progressbar_utils.ProgressBar(max_value=num_edges) as bar:
         for idx, (src_graph_node, dest_graph_node, switch_pkey, phy_tile_pkey,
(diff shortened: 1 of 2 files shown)


acomodi:
@litghost By removing the `pres_fac 20.0` change and restoring it, XC7 CI went green. I have also tested it locally and everything passed, as well as vendor tools. I think that the reason CI went red was the pres_fac change.

litghost:
> @litghost By removing the `pres_fac 20.0` change and restoring it, XC7 CI went green. I have also tested it locally and everything passed, as well as vendor tools. I think that the reason CI went red was the pres_fac change.

That makes sense.  I'm going to work on fixing the vendor tool tests today, and after that we can re-run the CI here.

xc/common/utils/prjxray_routing_import.py
```diff
@@ -698,39 +698,38 @@ def import_graph_nodes(conn, graph, node_mapping, connection_box_map):
             right_graph_node_pkey
         ) = result
 
-        side = node.loc.side
-        if side == tracks.Direction.LEFT:
-            assert left_graph_node_pkey is not None, (tile_type, pin_name)
-            node_mapping[left_graph_node_pkey] = node.id
+        # VPR emits only one node for each site pin, instead of one node for each
```
litghost, line 701: This code should consult with input arch to determine which sides the node is on.  See https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1577

litghost: This can be tracked in a follow up issue, it doesn't have to be fixed in this PR.  Currently upstream VPR is broken in this area, so it might blow up at any time :(

xc/common/utils/prjxray_routing_import.py
```diff
@@ -1019,6 +1023,20 @@ def get_number_graph_edges(conn, graph, node_mapping):
         if dest_graph_node not in node_mapping:
             continue
 
+        src_node, src_node_type = node_mapping[src_graph_node]
```
litghost, line 1026: Need a comment on what this is going?

litghost: Consider factoring in a function?

xc/common/utils/prjxray_routing_import.py
```diff
@@ -1061,6 +1081,24 @@ def import_graph_edges(conn, graph, node_mapping):
...
 
+            src_node, src_node_type = node_mapping[src_graph_node]
+            sink_node, sink_node_type = node_mapping[dest_graph_node]
+
+            pin_node_types = [graph2.NodeType.IPIN, graph2.NodeType.OPIN]
+
+            src_node_is_site_pin = src_node_type in pin_node_types
+            sink_node_is_site_pin = sink_node_type in pin_node_types
+
+            # It may happen that a same CHAN <-> PIN edge is generated and this is unaccepted
```
litghost, line 1092: This is bad, because that means that the underlying rr graph is less connected than the true hardware routing graph.  Also this code is too pessimistic.  VPR supports allowing multiple `CHAN <-> PIN` edges if a different switch is used.  So you need to account for that.

litghost (changes requested):
Comments in.  This fix is good, but actually points out a break in the VPR data model, I've filled an issue here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1577

xc/common/utils/prjxray_routing_import.py
acomodi: I have added an assertion to check whether the switch is the same for the same `CHAN <-> PIN` pair.

This was meant to avoid the error generated [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/37095c742471af2e2c8d90faf10ab13156a47942/vpr/src/route/check_rr_graph.cpp#L126-L138)

xc/common/utils/prjxray_routing_import.py
acomodi: Factored in a function and added a docstring to it

xc/common/utils/prjxray_routing_import.py
```diff
@@ -999,17 +1001,62 @@ def get_pip_wire_names(pip_pkey):
...
 
+def check_duplicate_chan_pin_edges(
+        nodes_dict, node_mapping, src_graph_node, dest_graph_node, switch_pkey
+):
+    """
+    This function checks whether there are multiple edges between the same
+    (CHAN, PIN) pair with the same switch.
+
+    It may happen that a same CHAN <-> PIN edge is generated and this is unaccepted
+    by VPR, as it allows only multiple edges between CHAN nodes.
```
litghost, line 1012: The error you linked now accommodates multiple CHAN <-> PIN edges?  So this comment is wrong?

xc/common/utils/prjxray_routing_import.py
litghost: You still haven't fixed this?

xc/common/utils/prjxray_routing_import.py
acomodi: Hmm, actually I think I found a bug in VPR, making this whole check useless. Testing it now

xc/common/utils/prjxray_routing_import.py
acomodi: Shouldn't this already be checked with the left/right/top/bottom graph node pkeys?

litghost: That checks the wrong direction, right?  The database is based on the arch-defs side, we need something from the rr graph side to line up.  Once https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1577 is fixed, this might revert to just using something from the rr graph.  Maybe factor the check into something like "is node X on side Y", and then it can be updated once https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1577 is fixed?

acomodi: I have added a check to verify the correctness of the side for a given node, using the arch.xml as input to get the side data.

litghost: Great.

xc/common/utils/prjxray_routing_import.py
```diff
@@ -1237,6 +1253,46 @@ def find_constant_network(graph):
...
+                side = Sides.TOP
+            elif side == "bottom":
+                side = Sides.BOTTOM
+
+            pins = loc.text.split()
+            for pin in pins:
+                pin_name = pin.split(".")[-1]
+
+                if pin_name not in pin_sides_dict[tile_name]:
+                    pin_sides_dict[tile_name][pin_name] = list()
```
litghost, line 1289: Should use a set rather than a list here?

xc/common/utils/prjxray_routing_import.py
```diff
@@ -1237,6 +1253,46 @@ def find_constant_network(graph):
...
+            elif side == "right":
+                side = Sides.RIGHT
+            elif side == "top":
+                side = Sides.TOP
+            elif side == "bottom":
+                side = Sides.BOTTOM
+
+            pins = loc.text.split()
+            for pin in pins:
+                pin_name = pin.split(".")[-1]
```
litghost, line 1286: Please add a comment showing an example of pin

xc/common/utils/prjxray_routing_import.py
```diff
@@ -1237,6 +1253,46 @@ def find_constant_network(graph):
     return synth_tiles
 
 
+def get_pin_sides_dict(arch_file):
+    """
+    This function is used to generate a dictionary containing the sides of each pin
+    of a physical tile.
+
+    This is necessary to have a correct location of nodes when a pin is present in multiple
+    sides of a tile.
```
litghost, line 1262: Document the return type?

xc/common/utils/prjxray_routing_import.py
```diff
@@ -1311,6 +1367,10 @@ def main():
         '--vpr_capnp_schema_dir',
         help='Directory container VPR schema files',
     )
+    parser.add_argument(
+        '--arch_file',
+        help='Architecture XML file',
```
litghost, line 1372: Set `required=True`?

xc/common/utils/prjxray_routing_import.py
acomodi: Done

xc/common/utils/prjxray_routing_import.py
acomodi: Done

xc/common/utils/prjxray_routing_import.py
acomodi: Done

xc/common/utils/prjxray_routing_import.py
acomodi: Yes, makes sense, fixed.

litghost (changes requested):
CI is pretty red?

acomodi:
> CI is pretty red?

Yes, verified also locally. The problem here is that the rr_graph_check fails as we now have the IPIN/OPIN nodes with multiple connections: https://github.com/SymbiFlow/vtr-verilog-to-routing/blob/f1a3bcc2a8aa93d88cb26ca7f1853962fff57c32/vpr/src/route/check_rr_graph.cpp#L126-L132

I am unsure how to proceed. Possible ways are:
- Update the VTR master+wip (unless we re-introduce the non-averaging bug during the indexed data calculation, we would see the increased run-time).
- Re-add the code to add edges only once if there are more then one `CHAN <-> PIN` connections.

acomodi:
In particular, the issue seems to affect IO Pads.

```
TILE: RIOPAD_M, PIN: IOI_OLOGIC0_CLKDIV, NODE ID: 289862, X: 114, Y: 111                                                                                                                                                                                                        
     GRAPH_NODE: 2164301                                                                                                                                                                                                                                                                 
     GRAPH_NODE: 2164300                                                                                                                                                                                                                                                        
     GRAPH_NODE: 2164302   
```

This pin node results in having three corresponding graph nodes (sides locations).
Each of the graph nodes ends up having the same RR node, and, during the edge import, the same RR node gets connected to a CHAN multiple times, generating the check_rr_graph_issue.

litghost:
>     * Update the VTR master+wip (unless we re-introduce the non-averaging bug during the indexed data calculation, we would see the increased run-time).

How close are you to debugging the new base cost calculation?  If it is going to be a bit, let's recut master+wip and put in this fix.

acomodi:
@litghost I have updated the new base cost calculation results [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1576#issuecomment-720518655). I think that it will require some more iterations to land, therefore I believe we might as well proceed by updating master+wip, possibly with a temporarily `wip` branch with the new base_cost calculation.

acomodi:
@litghost All CIs went green. This fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1768

litghost:
Ibex route time dropped, and CPD increased, 

before:
```
path | pack time (sec) | place time (sec) | route time (sec) | t_crit (ns) | Fmax (MHz)
-- | -- | -- | -- | -- | --
/tmpfs/src/github/symbiflow-arch-defs-continuous-xc7/build/xc/xc7/tests/soc/ibex/ibex_arty/artix7-xc7a50t-virt-xc7a50t-test | 35.16 | 34.79 | 98.66 | 27.7936 | 35.9795
```

after:
```
path | pack time (sec) | place time (sec) | route time (sec) | t_crit (ns) | Fmax (MHz)
-- | -- | -- | -- | -- | --
/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/xc/xc7/tests/soc/ibex/ibex_arty/artix7-xc7a50t-virt-xc7a50t-test | 35.65 | 36.69 | 85.54 | 30.7268 | 32.5449
```

Theories?

acomodi:
@litghost I'll need to look a closer look to this one.

One possibility might be that the different way we connect site PINs to CHAN could have caused this?

I cannot think about other reasons, as, apart from the different RR graph generated produced by VPR with one index for all the sides, there shouldn't have been other changes that influence the RR graph and the router.

litghost (commented):
> I cannot think about other reasons, as, apart from the different RR graph generated produced by VPR with one index for all the sides, there shouldn't have been other changes that influence the RR graph and the router.

That's why I'm pointing this out.  It surprised me :(

acomodi:
@litghost By building the ibex test locally with a clean repo, using this PR, the ibex CPD I get is `27.887`.

acomodi:
@litghost, with the bug with the non-deterministic ibex sv2v translation (fixed in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1782) and https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1776 We cannot rely too much on results between different builds when comparing LiteX and Ibex designs.

All other tests are deterministic, therefore, from https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1711#issuecomment-726248546 we can take as valid only the `ddr_uart` results.

We could compare other designs (e.g. scalable_proc) and, if everything looks fine and the code in this PR is ok, we could merge this, as it keeps holding the VtR+SymbiFlow CI going green.

PR #1719, opened 2020-10-22
Initial GH actions activation

acomodi:
GH actions will be used to get the latest successfull symbiflow install package.

The GH workflow works only if the yml configuration file is added to the master branch, where it will get to be effective on any status change for commits within the master branch.

Moreover, currently the GCS upload of the generated link is not present at the moment.

### DIFF
--- a/.github/workflows/checkstatus.yml
+++ b/.github/workflows/checkstatus.yml
@@ -5,7 +5,6 @@ on:
 
 jobs:
   Check-statuses:
-    if: ${{ contains(github.event.branches.*.name, 'master') == 0 }}
     runs-on: ubuntu-latest
 
     steps:
@@ -14,18 +13,14 @@ jobs:
         curl -s --url https://api.github.com/repos/${{ github.repository }}/commits/${{ github.event.sha }}/status \
         --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | head -n 2 | tail -n 1 | grep "success"
     - name: Set up Python
-      if: ${{ success() }}
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
     - name: Install STPM
-      if: ${{ success() }}
       run: pip install git+https://github.com/SymbiFlow/symbiflow-tools-package-manager#egg=stpm
     - name: Get latest artifact name
-      if: ${{ success() }}
       run: mkdir install && symbiflow_get_latest_artifact_url > install/latest
     - name: Upload artifact URL file
-      if: ${{ success()  }}
       uses: weslenng/gcp-storage-sync@master
       env:
         GCP_SERVICE_ACCOUNT_KEY_FILE: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY_FILE  }}
--- a/.github/workflows/checkstatus.yml
+++ b/.github/workflows/checkstatus.yml
@@ -16,10 +16,10 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
-    - name: Install STPM
-      run: pip install git+https://github.com/SymbiFlow/symbiflow-tools-package-manager#egg=stpm
+    - name: Install STDM
+      run: pip install git+https://github.com/SymbiFlow/symbiflow-tools-data-manager#egg=stdm
     - name: Get latest artifact name
-      run: mkdir install && symbiflow_get_latest_artifact_url > install/latest
+      run: mkdir install && get_latest_artifact_url > install/latest
     - name: Upload artifact URL file
       uses: weslenng/gcp-storage-sync@master
       env:
--- a/.github/workflows/checkstatus.yml
+++ b/.github/workflows/checkstatus.yml
@@ -15,7 +15,7 @@ jobs:
         --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | head -n 2 | tail -n 1 | grep "success"
     - name: Set up Python
       if: ${{ success() }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.8
     - name: Install STPM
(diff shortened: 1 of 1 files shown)


mithro:
@umarcor - Can you check this over for for GitHub action smells?

.github/workflows/checkstatus.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check status
+      run: |
+        curl -s --url https://api.github.com/repos/${{ github.repository }}/commits/${{ github.event.sha }}/status \
+        --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | head -n 2 | tail -n 1 | grep "success"
+    - name: Set up Python
+      if: ${{ success() }}
+      uses: actions/setup-python@v1
```
litghost, line 18: Is this something that github actions provides?  Why `@v1`?  When should `@v1` be changed?

kgugala: it's this one https://github.com/actions/setup-python and there is never version. @acomodi why do we use v1?

acomodi: No reason actually, I will switch to version 2

acomodi:
I have added also the GCP upload step, we need to set the following GH secrets to have the file with the link uploaded to GCS:
- `GCP_SERVICE_ACCOUNT_KEY_FILE`: this enables the authentication to GCP
- `GCP_STORAGE_BUCKET`: This is needed to identify the bucket we want to upload to.

https://github.com/marketplace/actions/gcp-storage-sync

umarcor:
Hi! I'm not sure I understand what's going on here. @acomodi, would you mind explaining what are we trying to achieve from a higher level perspective? Is the purpose to "retrieve the latest artifact from somwhere and push it to Google Cloud Platform (GCP)?

- Step "Check status" feels weird to me. It seems that you are using the API for checking the status of the same commit the workflow is run for.
- You should not need to define `if: ${{ success() }}` for each step after the first one. That should be the default (implicit behaviour). Moreover, I'd say it should be `if: success` or `if: success()`.
- What dows "STPM" stand for? Is it "Symbiflow Tools Package Manager"?
- What latest artifact do you want to get? Where is `symbiflow_get_latest_artifact_url` defined? Does it come from STPM and that's why you installed it?

acomodi:
Hi @umarcor, yes, the main purpose is to retrieve the latest package and push it on GCP:

> Step "Check status" feels weird to me. It seems that you are using the API for checking the status of the same commit the workflow is run for.

That's true, but we don't check just one status, but combined status of all checks, that is set to success only if every check passed.

> You should not need to define if: ${{ success() }} for each step after the first one. That should be the default (implicit) behaviour. Moreover, I'd say it should be if: success or if: success().

Fixed.

> What does "STPM" stand for? Is it "Symbiflow Tools Package Manager"?

Indeed, this is a script used to get packages generated from SymbiFlow CIs.

> What latest artifact do you want to get? Where is symbiflow_get_latest_artifact_url defined? Does it come from STPM and that's why you installed it?

We want to get one the artifact of one of the CI, namely `Xilinx Series 7 - Install`. And the symbiflow_get_latest_artifact_url comes indeed from STPM.

umarcor:
@acomodi thanks for clarifying. This is a new use case for me, and it's always enriching to learn.

> That's true, but we don't check just one status, but combined status of all checks, that is set to success only if every check passed.

I understand. You want to wait for Travis and Kokoro to run and have this workflow executed when and only when all of them are done.

Regarding constraining to `master`, there should be a more idiomatic way of doing it. As you see in https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration, for `push` and `pull_request` events, you can filter the branches. My guess is that the same approach should work for `status` events.

But, according to the note in https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#status, either your current implementation or filtering the event are NOT necessary. Events of type status do work with the default branch only. Hence, any further filtering is redundant.

Apart from that, `${{ }}` is not required in `if` fields. Not to be applied here, but for future uses.

> > You should not need to define if: ${{ success() }} for each step after the first one. That should be the default (implicit) behaviour. Moreover, I'd say it should be if: success or if: success().
> 
> Fixed.

I think you can remove the first one too. If "Check status" exits with non-zero, the workflow is finished there, by default.

> > What does "STPM" stand for? Is it "Symbiflow Tools Package Manager"?
> 
> Indeed, this is a script used to get packages generated from SymbiFlow CIs.

Just to be sure, with "SymbiFlow CIs" you mean the ones named "kokoro"? That is, neither Travis nor GitHub Actions?

> We want to get one the artifact of one of the CI, namely `Xilinx Series 7 - Install`. And the symbiflow_get_latest_artifact_url comes indeed from STPM.

I would say that `symbiflow_get_latest_artifact_url` is not the best naming if the target is to download `Xilinx Series 7 - Install` only. Anyway, this comment is absolutely out of the scope of this PR.

---

I have not tried this, but tool `hub` is available in GitHub Actions by default. According to https://hub.github.com, `hub ci-status --verbose` might be equivalent to the API check.

acomodi:
> Events of type status do work with the default branch only. Hence, any further filtering is redundant.

@umarcor Ok, I have removed the filering on the branch (as well as the `if: success`)

> Just to be sure, with "SymbiFlow CIs" you mean the ones named "kokoro"? That is, neither Travis nor GitHub Actions?

Probably I chose wrong words. I mean all CIs that run on master (the ones named Kokoro and also Travis)

> I would say that symbiflow_get_latest_artifact_url is not the best naming if the target is to download Xilinx Series 7 - Install only. Anyway, this comment is absolutely out of the scope of this PR.

Actually this tool will be more of a general purpose script to get latest artifacts of different CI builds. For the time being it defaults to fetch the `Xilinx Series 7 - Install` artifact.

PR #1731, opened 2020-10-28
Add environment configuration to install tarball

acomodi:
This PR also enables the 200T device installation within the other device's installation, so to have all of the devices in one package only.

### DIFF
--- a/.github/kokoro/install-200t.sh
+++ b/.github/kokoro/install-200t.sh
@@ -21,6 +21,7 @@ echo "----------------------------------------"
 	export VPR_NUM_WORKERS=${CORES}
 	ninja -j${MAX_CORES} install
 	popd
+	cp environment.yml install/
 )
 echo "----------------------------------------"
 
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -18,6 +18,7 @@ echo "----------------------------------------"
 	export VPR_NUM_WORKERS=${CORES}
 	ninja -j${MAX_CORES} install
 	popd
+	cp environment.yml install/
 )
 echo "----------------------------------------"
 


kgugala:
Do we want to restore a200t installation together with the rest of the xc7? I thought we will rather split all the architectures into separate packages.

acomodi:
Makes sense to split all different architectures in sub-packages. That is IMO the best way to proceed. For the time being I wanted to see how much time/size requires to have the 200T included in the same package.

Probably we need to modify also the STPM and the GH actions in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1719 to get the latest devices and generate one file with a link for each different device. So that in the end we would have the following:

- a link to the binary toolchain (this could potentially end up in a conda package and it will include all the bash scripts to run the toolchain)
- a set of links to the various devices tarballs

xc/xc7/tests/install_test/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
 set(INSTALLATION_DIR_BIN "${CMAKE_INSTALL_PREFIX}/bin")
 
-if(${INSTALL_DEVICE} STREQUAL "xc7a200t")
-add_test(NAME binary_toolchain_test
+add_test(NAME binary_toolchain_test_200t
```
mithro, line 3: Is there a `binary_toolchain_test_100t`? Should this also be `xc7a200t`?

mithro:
I think we should probably have an individual package per part, or a combined package which has all the parts. Having one package with `a50t` + `a100t` and another with `a200t`.

xc/xc7/tests/install_test/Makefile
```diff
@@ -17,17 +16,13 @@ ${BUILDDIR}:
 	mkdir ${BUILDDIR}
 
 ${BUILDDIR}/${TOP}.eblif: | ${BUILDDIR}
-ifeq (,$(wildcard ${XDC}))
```
litghost, line 20: Why is the XDC stuff gone?

litghost:
We should probably have a tarball for each part class (e.g. 50T, 100T, 200T).

xc/xc7/tests/install_test/Makefile
acomodi: I changed the test to use XDC only

mithro:
> We should probably have a tarball for each part class (e.g. 50T, 100T, 200T).
Agreed!

My sentence should have ended with the following;
> I think we should probably have an individual package per part, or a combined package which has all the parts. Having one package with `a50t` + `a100t` and another with `a200t` **is weird**.

HackerFoo (approved):

acomodi:
This PR seems to break the install CI. Unsure what is the reason as the log is empty. Maybe it takes too much time/memory/disk space?

acomodi:
I have modified this PR to only add the environment file to the tarball. The previous changes in this PR have been moved [here](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1747) alongside with a restructure of the install packages generated by CI

PR #1726, opened 2020-10-26
Review getting started, point to symbiflow-examples

mgielda:
Addressing: https://github.com/SymbiFlow/symbiflow-docs/issues/367

PR #1735, opened 2020-10-30
Adjust VPR flags to reduce runtime

HackerFoo:
With these settings, `baselitex` on the 50t is 24% faster on pack, place and route, while `ibex` is 16% faster, with `scalable_proc` tests running faster as well.


xc/common/cmake/arch_define.cmake
```diff
@@ -75,7 +75,10 @@ function(ADD_XC_ARCH_DEFINE)
       --incremental_reroute_delay_ripup off \
       --base_cost_type delay_normalized_length_bounded \
       --bb_factor 10 \
-      --initial_pres_fac 4.0 \
+      --acc_fac 0.7 \
+      --astar_fac 1.8 \
```
litghost, line 79: An A* factor of 1.8 likely has a very bad effect on QoR.  What do your results show on CPD with these changes?

HackerFoo: The difference is very small:
baselitex: [48.1 MHz](https://hydra.vtr.tools/build/40238/download/1/meta.json) -> [47.9 MHz](https://hydra.vtr.tools/build/41245/download/1/meta.json) sys_clk, clk200_clk slowed to 202 MHz, but still meets timing. 150 seconds faster.
ibex: [32.8 MHz](https://hydra.vtr.tools/build/40172/download/1/meta.json) -> [30.7 MHz](https://hydra.vtr.tools/build/41220/download/1/meta.json) clk. 50 seconds faster.

litghost: Well that means the lookahead is doing a good job.  Just keep in mind that increasing A* could result in dramatic QoR degradation if the lookahead mispredicts.

litghost:
CI is red, looks like flags were not deduped correctly.

Also I'd like to see what the runtime / QoR trade off looked like.

HackerFoo:
Comparing the last run to #1726 (old) for baselitex:

```
old, pack:  The entire flow of VPR took 147.20 seconds (max_rss 3498.1 MiB)
old, place:  The entire flow of VPR took 185.75 seconds (max_rss 3471.0 MiB)
old, route:  The entire flow of VPR took 488.92 seconds (max_rss 3471.1 MiB)
new, pack:  The entire flow of VPR took 152.99 seconds (max_rss 3499.8 MiB)
new, place:  The entire flow of VPR took 175.32 seconds (max_rss 3472.9 MiB)
new, route:  The entire flow of VPR took 293.65 seconds (max_rss 3472.3 MiB)

old CPD:
  sys_clk to sys_clk CPD: 18.2112 ns (54.9112 MHz)
  clk200_clk to clk200_clk CPD: 5.77223 ns (173.243 MHz)

new CPD:
  sys_clk to sys_clk CPD: 18.2289 ns (54.8581 MHz)
  clk200_clk to clk200_clk CPD: 5.66468 ns (176.533 MHz)
```

litghost:
> Also I'd like to see what the runtime / QoR trade off looked like.

Let me be more specific.  Much like inner num for the placer, A* is a runtime versus quality trade off.  The lookahead quality determines the sharpness of the trade off. I'd like to see how close to the edge we are with an A* of 1.8.  Also at some point as A* gets higher, the router may actually slow down again as the router excessively trusts the lookahead.

So I'm basically asking for two graphs, with A* on the x-axis of both.  On the y-axis of one is the runtime, and the other is CPD.  A third possible graph would be iterations to convergence on the y-axis.

How many circuits are you testing with?

acomodi:
I think that if we get good improvements in run-time at a slight CPD cost, it might be worth adding these flags.

One thing though would be to compare the xc7_qor produced by this PR's Kokoro CI and the one produced by the current master.

litghost:
> One thing though would be to compare the xc7_qor produced by this PR's Kokoro CI and the one produced by the current master.

xc7 QoR looks really good.  @HackerFoo can we add the curves to Colab?  The results from the arch-defs show this appears to be a solid point, at least per the circuits in arch-defs.  Also I think we should be prepared to show this results this Thursday and see if we can get some insight from Vaughn.

xc/common/cmake/arch_define.cmake
```diff
@@ -75,7 +75,10 @@ function(ADD_XC_ARCH_DEFINE)
       --incremental_reroute_delay_ripup off \
       --base_cost_type delay_normalized_length_bounded \
       --bb_factor 10 \
-      --initial_pres_fac 4.0 \
+      --acc_fac 0.7 \
```
litghost, line 78: Let's add a comment pointing to documentation on how these parameters were chosen.  E.g. we swept the following circuits, over the following parameters, etc.

xc/common/cmake/arch_define.cmake
```diff
@@ -65,7 +65,7 @@ function(ADD_XC_ARCH_DEFINE)
       --router_heap bucket \
       --clock_modeling route \
       --place_delta_delay_matrix_calculation_method dijkstra \
-      --place_delay_model delta_override \
+      --place_delay_model delta \
```
litghost, line 68: Given that the results look good, this is fine, but this a little surprising.  We should talk to Vaughn about this.

HackerFoo:
I'm running a sweep and also documenting how I do this [here](https://colab.research.google.com/drive/1X91RGZnvlC7dBjJJUbS7JfqCbPCzJ3Xb?usp=sharing).

xc/common/cmake/arch_define.cmake
HackerFoo: Should this go somewhere in `docs/`, or as comments in this file?

litghost: A comment pointing to more details docs is probably good here.

HackerFoo:
I've added a scatter plot of runtime vs. max frequency for each of ibex, baselitex, and bram-n3.

xc/common/cmake/arch_define.cmake
vaughnbetz: Ok with me, given discussion today.

xc/common/cmake/arch_define.cmake
vaughnbetz: Agreed, comment that is a pointer to a doc is probably good here.

xc/common/cmake/arch_define.cmake
vaughnbetz: Good to tune on more designs in the longer term, as 3 designs is a small number for fine tuning.

vaughnbetz: Default initial_pres_fac is 0.5; pres_fac_mult is 1.3.

vaughnbetz: So you pay more attention to congestion early, which makes sense that it helps runtime. You ramp the cost of congestion more slowly, which would probably partially compensate for a higher initial cost of congestion (may mitigate  quality losses). So the congestion schedule is just different, in a way that seems reasonable.

xc/common/cmake/arch_define.cmake
HackerFoo: I've added documentation inline, and a link to the detailed report.

xc/common/cmake/arch_define.cmake
```diff
@@ -61,11 +61,24 @@ function(ADD_XC_ARCH_DEFINE)
...
+  # - acc_fac: Lowering this to 0.7 slightly improves runtime 1-4% (default: 1)
+  # - astar_fac: Increasing this to 1.8 reduces runtime 2-15% (default: 1.2)
+  # - initial_pres_fac: Setting this to 2.828 reduces runtime 10-20% from the default,
+  #   and up to 3% over the previously used value of 4 (default: 0.5)
+  # - place_delay_model: `delta` is up to 5% faster than `delta_override` with slight to no
+  #   degradation in QoR (default: delta)
+  # - pres_fac_mult: A slightly lower value of 1.2 performs better on average by ~5% for some designs (default: 1.3)
+  # Based on analysis performed on hydra.vtr.tools for the ibex, baselitex, and bram-n3 designs.
+  # These changes did not have a measurable effect on QoR for these designs.
+  # More details can be found in the report: https://colab.research.google.com/drive/1X91RGZnvlC7dBjJJUbS7JfqCbPCzJ3Xb
```
litghost, line 76: Thanks for the additional comments.  My confusion at this point is how from the data you posted did you select these particular settings for `acc_fac`/`astar_fac`/`initial_pres_fac`/`pres_fac_mult`.

HackerFoo: I started with parameters that resulted in the shortest runtime for each project, and I keep the parameters that agree. Then, for the remaining parameters, I try values and check that they rank highly among all projects. I use the statistics to check that the selected parameters rank well with mean and minimum values, and that fmax doesn't suffer. I also look at the scatter plot to get an idea of the larger population of configurations, because I want to avoid parameter values that I don't think will generalize.

It's not very scientific, and I'm not sure how to automate it yet. Maybe ML :)

litghost: I'm not interested in full automation here, as I agree that is not easy.  This comment was not intended to get a reply here, but to modify the comment and/or Colab to capture the process.

Are there some short hands you can add to the Colab for the process of boiling it down?  Writing down what you did in a process is probably the minimum here, but ideally there'd be a way for others to follow what you've done.

HackerFoo: I've expanded on this a bit in the [Colab](https://colab.research.google.com/drive/1X91RGZnvlC7dBjJJUbS7JfqCbPCzJ3Xb?usp=sharing).

HackerFoo:
@litghost I'd like to get this PR merged. Can we discuss any further process documentation in the [colab document](https://colab.research.google.com/drive/1X91RGZnvlC7dBjJJUbS7JfqCbPCzJ3Xb?usp=sharing)?

xc/common/cmake/arch_define.cmake
HackerFoo: Thanks. I realized this isn't that bad to automate -- just sort by the minimum percent improvement across some representative projects. I've added this table to the colab, and found some better parameters that give a 15% runtime improvement on both `baselitex` and `ibex`. Thanks for the discussion!

litghost: >  I've added this table to the colab, and found some better parameters that give at least 15% runtime improvement on both baselitex and ibex

That's good news! Given that the Colab page now really has code in it to select parameters, we should probably manage it like code, e.g. check it in.  Can you export the current Colab page (e.g. the `.ipynb`) and add it to this PR?

litghost: I like the "sort by improvement" approach.  Should that table also have relative CPD effect column for quickly evaluating the potential cost to QoR against the benefit to runtime?  I think with that, it is straight forward enough to pick a point with reasonable improvement with minimal to no QoR cost.

HackerFoo: I added a "cpd_improvement" column. The parameters I chose increased CPD by 5%. I chose not to modify `--first_iter_pres_fac` because it made `xc/xc7/tests/counter/counter_basys3` unroutable .

HackerFoo: I've checked in the colab at `utils/ipynb/Parameter_Sweep_using_fpga_tool_perf.ipynb`.

litghost:
> I'm running a sweep and also documenting how I do this [here](https://colab.research.google.com/drive/1X91RGZnvlC7dBjJJUbS7JfqCbPCzJ3Xb?usp=sharing).

The instructions on the Colab page and the nix-symbiflow README don't appear to provide enough instructions to actually re-run the results.

As a concrete example, the Colab page says `edit to point at your fork`.  That is not meaningful to me reading the Colab page.  The linked `.nix` doesn't appear to specify a fork.  

To be clear, someone new to the team should be able to re-run the data gathering phase and regenerate the data on the Colab page from the instructions contained within.  I don't believe the current instructions meet that benchmark.

HackerFoo:
@litghost I've added more detailed instructions to the colab.

litghost:
> @litghost I've added more detailed instructions to the colab.

New instructions look good.  Please update the Colab copy in git.  Last thing (besides looking at CI results) is to change the references for https://github.com/HackerFoo/nix-symbiflow to https://github.com/Symbiflow/nix-symbiflow .  You need to add DCO checks to that repo, and add DCO on your commits.

HackerFoo:
> New instructions look good. Please update the Colab copy in git. Last thing (besides looking at CI results) is to change the references for https://github.com/HackerFoo/nix-symbiflow to https://github.com/Symbiflow/nix-symbiflow . You need to add DCO checks to that repo, and add DCO on your commits.

@litghost I've updated the Colab in git, and changed the references to https://github.com/SymbiFlow/nix-symbiflow, which has DCO checks.

I'm re-running the "Xilinx Series 7 - Install (Presubmit)" test which failed due to an [infrastructure failure](https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1768#issuecomment-726025066).

Assuming there are no problems with that, is this PR okay to merge?

HackerFoo:
Runtime is 24% faster for ibex (2% higher CPD), 10% faster for litex (7% lower CPD.)

litghost:
> Runtime is 24% faster for ibex (2% higher CPD), 10% faster for litex (7% lower CPD.)

So the previous settings from ead80ae0 were Pareto improvement on `geomean(route_time)` and `geomean(CPD)` over master.  That is not the case on the latest settings.  I believe the settings from ead80ae0 were a better point?

HackerFoo:
@litghost Which designs are worse? I can revert the settings to https://github.com/SymbiFlow/symbiflow-arch-defs/commit/ead80ae0bded1d674af70d4e9bc83c58efb1cca8.

litghost:
ibex and ddr_uart_arty both show significant change, but wide array of designs are worse at that design point.  Mean percentage change is 4% worse, geomean CPD is 3% worse.  The other point had a geomean CPD change of less than 0.2 %.

HackerFoo:
@litghost Okay, I've reverted the settings. Anything else before I merge this?

litghost:
> @litghost Okay, I've reverted the settings. Anything else before I merge this?

Just waiting for green.  Please rebase on master to grab the other fixes.

litghost (approved):
LGTM, merge once green.  Recommend rebasing on master.

HackerFoo:
Vendor tool tests are failing due to unrelated compilation errors.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 cairosvg
 gitpython
-hilbertcurve
+hilbertcurve==1.0.5
```
litghost, line 3: @HackerFoo Please file an issue with upstream `hilbertcurve`, and create an issue to remove the pin once upstream `hilbertcurve` is fixed.

HackerFoo: I don't think the issue is upstream. The API changed.

litghost: The PR comment in https://github.com/galtay/hilbertcurve/pull/25#issue-519041227 indicated that that specific PR was supposed to be backwards compatible.  It is unclear if the API interface was supposed to change here, or if it was an accident.

litghost: My comment was basically to post an issue upstream showing that from 1.0.5 to 2.0.x that this API no longer existed/worked, and determine if that break was intentional?

litghost: Although the notes from 2.0.3 do say `New API`, which is foreboding.

Regardless, we should likely create an issue to remove the pin in the future to avoid being in a situation where we have a very stale dependency.  This isn't equivalent, but a numpy pin on prjxray eventually resulted in the pip install for numpy to require building it instead of using pip binary caches.  Ideally we'd like to avoid something like that.

HackerFoo: I propose removing the dependency and using VPR's RR node reordering option: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1773

litghost: Sure.  Can you please open a PR to that effect?

HackerFoo: Yeah, I'll assign the issue to me.

PR #1742, opened 2020-11-01
ci/check-status: set-output name=skip

### LINKED ISSUE #1741: Github action 'checkstatus.yml' fails while status pending
When the status checks are still pending, the github action fails and causes you to get emailed...

https://github.com/SymbiFlow/symbiflow-arch-defs/runs/1337290559?check_suite_focus=true
```
  curl -s --url https://api.github.com/repos/SymbiFlow/symbiflow-arch-defs/commits/b31e4a4336d37ea0f2f48b485eeca697da48c9e2/status \
  --header 'authorization: Bearer ***' | head -n 2 | tail -n 1 | grep "success"
  shell: /bin/bash -e ***0***
Error: Process completed with exit code 1.
```

Maybe it should be something like;
```bash
if curl -s --url https://api.github.com/repos/SymbiFlow/symbiflow-arch-defs/commits/b31e4a4336d37ea0f2f48b485eeca697da48c9e2/status --header 'authorization: Bearer ***' | head -n 2 | tail -n 1 | grep "success"; then
  mkdir install && symbiflow_get_latest_artifact_url > install/latest
fi
```

umarcor:
Close #1741

This is a small enhancement to #1719. Currently, if step 'Check status' fails, the CI run is reported as a failure. This PR sets an output variable (`skip`) instead. That variable is used as the `if` condition in all the remaining steps. As a result, CI runs will all be green, unless there is some issue with any of the steps after the first one.

### DIFF
--- a/.github/workflows/checkstatus.yml
+++ b/.github/workflows/checkstatus.yml
@@ -4,26 +4,41 @@ on:
   status:
 
 jobs:
+
   Check-statuses:
     runs-on: ubuntu-latest
-
     steps:
+
     - name: Check status
-      run: |
-        curl -s --url https://api.github.com/repos/${{ github.repository }}/commits/${{ github.event.sha }}/status \
-        --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | head -n 2 | tail -n 1 | grep "success"
+      id: check
+      run: >-
+        curl -fsSL
+        -H 'Accept: application/vnd.github.v3+json'
+        -H 'authorization: Bearer ${{ github.token }}'
+        https://api.github.com/repos/${{ github.repository }}/commits/${{ github.event.sha }}/status
+        | head -n 2
+        | tail -n 1
+        | grep "success"
+        || echo '::set-output name=skip::true'
+
     - name: Set up Python
+      if: steps.check.outputs.skip != 'true'
       uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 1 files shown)


umarcor:
@mithro this PR fixes #1741.

.github/workflows/checkstatus.yml
```diff
@@ -4,26 +4,41 @@ on:
...
+        -H 'Accept: application/vnd.github.v3+json'
+        -H 'authorization: Bearer ${{ github.token }}'
+        https://api.github.com/repos/${{ github.repository }}/commits/${{ github.event.sha }}/status
+        | head -n 2
+        | tail -n 1
+        | grep "success"
+        || echo '::set-output name=skip::true'
+
     - name: Set up Python
+      if: steps.check.outputs.skip != 'true'
```
mithro, line 25: Is no way to wrap all the remaining steps in the if?

mithro (approved):
LGTM!

.github/workflows/checkstatus.yml
umarcor: Unfortunately, I don't think so: https://stackoverflow.com/questions/60589373/how-to-force-to-exit-in-github-actions-step

It would be possible to define two jobs (instead of one job with multiple steps). However, then we'd need to pass some parameters from one job to another. I think that'd be too contrived...

PR #1743, opened 2020-11-01
ci: use custom Action 'CheckStatus'

### LINKED ISSUE #1738: Give github actions 'checkstatus.yml' a better name
Currently it has;
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/c5ca99881224ff55e329cd6c44882f9d567ee593/.github/workflows/checkstatus.yml#L1

"Check all tests" could mean many things and it doesn't even describe what this github action actually does. I think a better name would be something like "Update latest artifacts redirect on successful CI completion" or something similar.

### LINKED ISSUE #1741: Github action 'checkstatus.yml' fails while status pending

umarcor:
Close #1741
Close #1742
Close #1738

Instead of using `curl`, `setup-python` and two additional steps, in this PR a custom Action named `CheckStatus` is added to this repository. That's a Docker based GitHub Action. In a `python:alpine` container, script `CheckStatus.py`, the entrypoint shell script and `symbiflow-tools-data-manager` are installed.

`CheckStatus` uses [PyGithub](https://github.com/PyGithub/PyGithub), a Python package for interacting with the GitHub API. [get_combined_status](https://pygithub.readthedocs.io/en/latest/github_objects/Commit.html?highlight=status#github.Commit.Commit.get_combined_status) is used for retrieving the same content as with `curl`. Since, `CheckStatus` is written in Python, it is trivial to iterate over [statuses](https://pygithub.readthedocs.io/en/latest/github_objects/CommitCombinedStatus.html#github.CommitCombinedStatus.CommitCombinedStatus.statuses) for printing the current status of each check. See for example https://github.com/umarcor/symbiflow-arch-defs/runs/1337747107?check_suite_focus=true#step:3:92.

> NOTE: Action CheckStatus is based on [eine/tip](https://github.com/eine/tip).

WRT #1738, Action `CheckStatus` has a description explaining what it does.

Possible enhancements:

- Add colour to `pending`, `success`, `failure`, etc.
- Show `X of Y`: https://pygithub.readthedocs.io/en/latest/github_objects/CommitCombinedStatus.html#github.CommitCombinedStatus.CommitCombinedStatus.total_count.

---

Unfortunately, I could not have this workflow properly triggered in my fork. Even though I enabled Travis (https://travis-ci.com/github/umarcor/symbiflow-arch-defs/builds), the `statuses` feature of the fork seems to be empty:

```sh
# curl -fsSL -H 'Accept: application/vnd.github.v3+json' -H 'authorization: Bearer ***' https://api.github.com/repos/umarcor/symbiflow-arch-defs/commits/15f9863ab58da8ea82a4dac3886b110a7d8bdea7/status
{
  "state": "pending",
  "statuses": [

  ],
  "sha": "15f9863ab58da8ea82a4dac3886b110a7d8bdea7",
  "total_count": 0,
  "repository": {
...
```

According to https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#statuses, statuses are for external services only. Hence, I would expect the feature not to work with GitHub Actions workflows, but it should work with either Travis or Kokoro. On the other hand, [Check API](https://docs.github.com/en/free-pro-team@latest/rest/reference/checks) is mentioned. According to https://blog.travis-ci.com/2018-09-27-deprecating-github-commit-status-api-for-github-apps-managed-repositories, Travis deprecated the `statuses` API in September 2018, and recomends using `check-runs` instead.

My guess is that this repository was created before Travis deprecated `statuses`. That's why it works, but I cannot reproduce in my fork. Using `check-runs` seems to work both in my fork and here. Therefore, I'd suggest changing the API endpoint. We can do it either before or after merging this PR.

Note that changing the API entrypoint means changing the event type in the workflow YAML file too. See  https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#check_run and https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#check_suite.

---

Events of type `statuses` or `check_run`/`check_suite` should provide the same payload that we are retrieving from the API (either through curl or PyGitHub). [github.event](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context) (which is already used in the workflow to retrieve the SHA), should contain some more relevant data. We should be able to pass that content to CheckStatus as JSON: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#example-printing-context-information-to-the-log-file.

/cc @mithro @acomodi @kgugala


mithro:
> My guess is that this repository was created before Travis deprecated statuses. That's why it works, but I cannot reproduce in my fork. Using check-runs seems to work both in my fork and here. Therefore, I'd suggest changing the API endpoint. We can do it either before or after merging this PR.

Let's do this bit first?

.github/CheckStatus/CheckStatus.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
```
mithro, line 1: Need copyright header.

mithro (approved):
Generally LGTM, but let's wait for @acomodi and @kgugala to respond before merging.

acomodi (changes requested):
In general LGTM, I think it is definitely worth to have more control on the APIs response to print out for instance information on which CI passed or not.

I'd change the camel case to use dash (e.g. check-status) to be consistent with other GH actions (e.g. gcp-storage-sync, etc).

> Therefore, I'd suggest changing the API endpoint. We can do it either before or after merging this PR.

I'd say that if the check-runs API works better for Travis, we should already transition using that. The only question related to this is whether the Kokoro CIs statuses as  well can be inspected with teh check-runs API. If not, I think we should stick with statuses API instead

.github/CheckStatus/CheckStatus.py
umarcor: Which should it be? I didn't find copyright headers in this repo, which I could use as a reference.

umarcor:
> I'd change the camel case to use dash (e.g. check-status) to be consistent with other GH actions (e.g. gcp-storage-sync, etc).

Done.

> Let's do this bit first?

> I'd say that if the check-runs API works better for Travis, we should already transition using that. The only question related to this is whether the Kokoro CIs statuses as well can be inspected with teh check-runs API. If not, I think we should stick with statuses API instead

I did some further tests using `check_suite`. Unfortunately, that's not supported in https://github.com/PyGithub/PyGithub/issues/1063 (yet). As an alternative, I tried using the data in `github.event`. See https://github.com/umarcor/symbiflow-arch-defs/blob/master/.github/workflows/checkstatus.yml#L15-L16. However, I need to get more familiar with it, before proposing a robust solution.

acomodi (approved):
LGTM

PR #1747, opened 2020-11-02
Restructure install packages

acomodi:
This PR builds on top of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1731 and it performs the install step on all the devices.

It generates different packages for each device, as well as a package containing the common required scripts/utils (e.g. toolchain_wrappers, yosys scripts, python utils, etc):

The final content uploaded to GCS should look like the following:
- symbidlow-arch-defs-<commit-sha>.tar.xz
- xc7a50t.tar.xz
- xc7a100t.tar.xz
- ...

### DIFF
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -51,9 +51,8 @@ echo "----------------------------------------"
 	tar cfJv symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share/symbiflow/techmaps share/symbiflow/scripts environment.yml
 	for device in $(ls install/share/symbiflow/arch)
 	do
-		tar cfJv $device.tar.xz -C install $device
+		tar cfJv $device.tar.xz -C install/share/symbiflow/arch $device
 	done
-	source env/conda/bin/activate symbiflow_arch_def_base
 )
 echo "----------------------------------------"
 
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -49,6 +49,7 @@ echo "----------------------------------------"
 	du -ah install
 	export GIT_HASH=$(git rev-parse --short HEAD)
 	tar cfJv symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share/symbiflow/techmaps share/symbiflow/scripts environment.yml
+	tar cfJv symbiflow-arch-defs-benchmarks-${GIT_HASH}.tar.xz -C install benchmarks
 	for device in $(ls install/share/symbiflow/arch)
 	do
 		tar cfJv symbiflow-$device.tar.xz -C install share/symbiflow/arch/$device
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -48,8 +48,12 @@ echo "----------------------------------------"
 	rm -rf build
 	du -ah install
 	export GIT_HASH=$(git rev-parse --short HEAD)
-	tar cfJv symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share environment.yml
-	tar cfJv symbiflow-arch-defs-benchmarks-${GIT_HASH}.tar.xz -C install benchmarks
+	tar cfJv symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share/symbiflow/techmaps share/symbiflow/scripts environment.yml
+	for device in $(ls install/share/symbiflow/arch)
+	do
+		tar cfJv $device.tar.xz -C install $device
+	done
+	source env/conda/bin/activate symbiflow_arch_def_base
 )
 echo "----------------------------------------"
 
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -51,7 +51,7 @@ echo "----------------------------------------"
 	tar cfJv symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share/symbiflow/techmaps share/symbiflow/scripts environment.yml
 	for device in $(ls install/share/symbiflow/arch)
 	do
-		tar cfJv symbiflow-$device.tar.xz -C install/share/symbiflow/arch $device
+		tar cfJv symbiflow-$device.tar.xz -C install share/symbiflow/arch/$device
 	done
 )
 echo "----------------------------------------"
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -52,7 +52,7 @@ echo "----------------------------------------"
 	tar cfJv symbiflow-arch-defs-benchmarks-${GIT_HASH}.tar.xz -C install benchmarks
 	for device in $(ls install/share/symbiflow/arch)
 	do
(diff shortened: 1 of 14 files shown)


xc/common/cmake/install.cmake
```diff
@@ -56,20 +56,20 @@ function(DEFINE_XC_TOOLCHAIN_TARGET)
...
 
   install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/utils/lib/parse_pcf.py
-          DESTINATION bin/python/lib)
+          DESTINATION bin/python-utils/lib)
 
 
   # install prjxray techmap
   install(DIRECTORY ${symbiflow-arch-defs_SOURCE_DIR}/xc/${FAMILY}/techmap
-          DESTINATION share/symbiflow/techmaps/${FAMILY}_vpr)
+          DESTINATION share/techmaps/${FAMILY}_vpr)
```
litghost, line 68: Why did we remove this subdir?

litghost: This change didn't get propagated to the scripts?

xc/xc7/tests/install_test/counter.v
```diff
@@ -11,11 +11,8 @@ module top (
     localparam BITS = 4;
     localparam LOG2DELAY = 22;
 
-    wire clk_to_bufg;
```
litghost, line 14: Why change this?  If we are doing this, also remove the BUFG completely?

xc/common/cmake/install.cmake
```diff
@@ -56,20 +56,20 @@ function(DEFINE_XC_TOOLCHAIN_TARGET)
             ${symbiflow-arch-defs_SOURCE_DIR}/utils/vpr_io_place.py
             ${symbiflow-arch-defs_SOURCE_DIR}/utils/vpr_place_constraints.py
             ${symbiflow-arch-defs_SOURCE_DIR}/utils/eblif.py
-          DESTINATION bin/python
+          DESTINATION bin/python-utils
```
litghost, line 59: Why did we rename this?

xc/xc7/tests/install_test/counter.v
acomodi: Well, I wanted to keep the BUFG to be able to apply the XDC timing constraints of the `bufg` net

xc/common/cmake/install.cmake
acomodi: The idea behind this PR is also to get the install package to be installed via conda.
The `python` directory within the `bin` would generate a conflict with the pre-existing conda `python`. In this way, no conflict is generated. I could as well move all the python scripts within the `share/symbiflow/scripts/` directory though.

xc/common/cmake/install.cmake
acomodi: This was a leftover, Fixed.

acomodi:
The install CI on this PR always ends up in a failure which does not allow us to verify the cause, as the log is empty. I am unsure how to debug this.

xc/xc7/tests/install_test/counter.v
litghost: The SDC clock propagation should be working, so the BUFG shouldn't be needed?

xc/common/cmake/install.cmake
litghost: > I could as well move all the python scripts within the `share/symbiflow/scripts/` directory though.

Ya, please move everything under `share/symbiflow`.

xc/xc7/tests/install_test/counter.v
acomodi: We are not yet auto-inferring the BUFG, therefore it still needs to be explicitly instantiated in the design

litghost: Ah, ok.  We should add that to the burndown list.

HackerFoo (approved):

litghost:
> The install CI on this PR always ends up in a failure which does not allow us to verify the cause, as the log is empty. I am unsure how to debug this.

I think I figured out the issue.  Fix is coming in 1275d6c7caed0f07bcc81822815f0852b5b9c5ec .  Basically kokoro has a 100 GB max on the entire workspace, regardless of the artifact filter.  I believe we were pushing up against that limit.

litghost:
Please rebase on master.

acomodi:
CI has finished and we have the following packages:
- symbiflow toolchain
- device packages.

Next steps are:
1. add functionality to https://github.com/SymbiFlow/symbiflow-tools-data-manager to get the latest links of all the packages, as it now it only returns one of them
2. modify the GH workflow to produce one link for each package
3. generate a conda package for the symbiflow toolchain
4. think on how best to download and install a required device package

acomodi:
I have rebased on top of master and included also fixes in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1828. 

I have also disabled the 200T install CI from https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1824

.github/kokoro/install.sh
```diff
@@ -48,8 +48,12 @@ echo "----------------------------------------"
 	rm -rf build
 	du -ah install
 	export GIT_HASH=$(git rev-parse --short HEAD)
-	tar cfJv symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share environment.yml
+	tar cfJv symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share/symbiflow/techmaps share/symbiflow/scripts environment.yml
 	tar cfJv symbiflow-arch-defs-benchmarks-${GIT_HASH}.tar.xz -C install benchmarks
+	for device in $(ls install/share/symbiflow/arch)
+	do
+		tar cfJv symbiflow-$device.tar.xz -C install share/symbiflow/arch/$device
```
litghost, line 55: Can you explain why this is `symbiflow-<device>` rather than `symbiflow-<git hash>-<device>`?

litghost (commented):
Minor comments.  I'll take a look at the output products once done too.

.github/kokoro/install.sh
acomodi: One reason would be that it would make easier to generate the latest link name in the GH actions script (https://github.com/SymbiFlow/symbiflow-arch-defs/blob/a0dfd328c974930671f6fb9e4a47f3959547c611/.github/check-status/check-status.py#L33), but I can fix this to be `symbiflow-arch-defs-<device>-<git hash>` and make the latest link generation capable of extracting the device name

litghost:
> Minor comments. I'll take a look at the output products once done too.

Outputs look good.  Lets rename the filenames to include the git hash, and I think this is ready to go!

acomodi:
> Outputs look good. Lets rename the filenames to include the git hash, and I think this is ready to go!

@litghost Done, I have also included the latest version of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1828

.github/kokoro/install.sh
acomodi: Done

litghost (approved):
LGTM, merge once green.

PR #1749, opened 2020-11-03
check-status: remove variable from yml description of the gh action

acomodi:


acomodi:
Merging this PR as it only touches the GH actions, which is non-related to any part of the CIs build flow.

PR #1751, opened 2020-11-04
ci/check-status: see content

umarcor:
This PR moves the branch condition to the last step, so that we can see what's the event/status data when the runs are skipped (most of the time). The JSON data is printed in a separate step.

### DIFF
--- a/.github/workflows/checkstatus.yml
+++ b/.github/workflows/checkstatus.yml
@@ -6,7 +6,6 @@ on:
 jobs:
 
   Check-statuses:
-    if: ${{ contains(github.event.branches.*.name, 'master') == 1 }}
     runs-on: ubuntu-latest
     steps:
 
@@ -19,7 +18,7 @@ jobs:
         sha: ${{ github.event.sha }}
 
     - name: Upload artifact URL file
-      if: steps.check.outputs.skip != 'true'
+      if: (steps.check.outputs.skip != 'true') && contains(github.event.branches.*.name, 'master')
       uses: weslenng/gcp-storage-sync@master
       env:
         GCP_SERVICE_ACCOUNT_KEY_FILE: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY_FILE }}
--- a/.github/workflows/checkstatus.yml
+++ b/.github/workflows/checkstatus.yml
@@ -11,6 +11,9 @@ jobs:
 
     - uses: actions/checkout@v2
 
+    - name: Print GitHub event
+      run: echo '${{ toJson(github.event) }}'
+
     - uses: ./.github/check-status
       id: check
       with:


acomodi:
With the current status, I have verified that only commits on the master default branch trigger the GH workflow. If we want to investigate further what the event contain, we can merge this PR, but this should be only temporary.

Basically, there is no way to not trigger the GH workflow if a commit status changes in a PR, but at least, cancelling the job at the very beginning let us have two types of CI results, cancelled or successful. It is then trivial to get to filter out the cancelled jobs and inspect the successful ones (which are only run on master).

A kind of unrelated note is that IMO, we would need to filter out the status of specific Kokoro CIs, as their result is unstable and does not depend only on the status of the symbiflow-arch-defs repository. In particular, the `VtR + Symbiflow-arch-defs` CI is pretty unstable, preventing the GH action to successfully upload the latest valid package, even though all the relevant CIs turned out to be green. @litghost @mithro Thoughts on this?

umarcor:
> With the current status, I have verified that only commits on the master default branch trigger the GH workflow. If we want to investigate further what the event contain, we can merge this PR, but this should be only temporary.

My understanding is that only commits on the default branch do trigger the workflow, however, for each push to master, several events are produced in a chain. Each time one of the tests changes from `queued` to `pending` to `success`/`failure` an event is created. That's why we can see 8-12 executions almost at the same time: https://github.com/SymbiFlow/symbiflow-arch-defs/actions.

The purpose of this PR is to see the payload of all the n-1 events which contain some `queued` or `pending` job/test. For example, compare https://github.com/SymbiFlow/symbiflow-arch-defs/runs/1355929480?check_suite_focus=true and https://github.com/SymbiFlow/symbiflow-arch-defs/runs/1355928556?check_suite_focus=true. Those are different runs, a few seconds apart, but produce exactly the same result. My understanding is that the state of the check suite was different when the events were created, but a few seconds later (when the API calls were done) the state was the same.

> Basically, there is no way to not trigger the GH workflow if a commit status changes in a PR, but at least, cancelling the job at the very beginning let us have two types of CI results, cancelled or successful. It is then trivial to get to filter out the cancelled jobs and inspect the successful ones (which are only run on master).

We can programmatically cancel non-master jobs in the Python script. That would preserve the behaviour you expect, while allowing us to see some content before cancelling.

> A kind of unrelated note is that IMO, we would need to filter out the status of specific Kokoro CIs, as their result is unstable and does not depend only on the status of the symbiflow-arch-defs repository. In particular, the `VtR + Symbiflow-arch-defs` CI is pretty unstable, preventing the GH action to successfully upload the latest valid package, even though all the relevant CIs turned out to be green. @litghost @mithro Thoughts on this?

The purpose of this PR is to better understand this context. If the event payload specifies that Kokoro produced it (and not Travis) and which test it was, we can use that information instead of calling the API. Anyway, we can always filter the JSON that we are currently getting from the API.

PR #1758, opened 2020-11-06
Add QuickLogic support

rw1nkler:
This PR adds support for QuickLogic devices. The codebase comes from the main [QuickLogic repository](https://github.com/QuickLogic-Corp/symbiflow-arch-defs/tree/quicklogic-upstream-rebase),  with additional changes that are required for proper integration with the SymbiFlow mainline.

At this point, these issues need to be addressed to provide the QuickLogic support:
- [ ] Copy the repository included in this PR as submodules to SymbiFlow
- [ ] Set up CI for QuickLogic devices
- [ ] Create Yosys plugin that will contain the changes introduced in https://github.com/SymbiFlow/yosys/pull/89
- [ ] Add all the remaining changes to SymbiFlow v2x

I believe that the integration can be speeded up by replacing the `Yosys` and `v2x` conda packages in the CI.
In this PR I proposed the necessary changes for this solution.


rw1nkler:
@litghost, can you provide any information on what should be added to integrate the QuickLogic support with SymbiFlow?
Probably, your action is also required to set up proper CI. Do you think that the solution with overwriting certain packages temporarily in the CI is a good approach?

quicklogic/common/cmake/quicklogic_fasm.cmake
```diff
@@ -0,0 +1,44 @@
...
+    OUTPUT ${QUICKLOGIC_FASM_OUTPUT}
+    DEPENDS ${DEPS} ${QLFASM}
+    COMMAND qlfasm ${CMD_ARGS} ${QUICKLOGIC_FASM_FILE} ${QUICKLOGIC_FASM_OUTPUT}
+    )
+  add_file_target(FILE ${QUICKLOGIC_FASM_OUTPUT} GENERATED)
+
+  get_rel_target(REL_QUICKLOGIC_FASM_NAME quicklogic_fasm ${QUICKLOGIC_FASM_NAME})
+  add_custom_target(
+    ${REL_QUICKLOGIC_FASM_NAME}
+    DEPENDS ${QUICKLOGIC_FASM_OUTPUT}
```
litghost, line 42: This should use `add_file_dependency`, not directly depend on the output filename.

litghost (changes requested):
Anytime an `add_custom_command` depends on something that has `add_file_target`, it needs to use `add_file_dependency`, not just depending on the filename.

common/cmake/devices.cmake
```diff
@@ -1365,6 +1365,26 @@ function(ADD_FPGA_TARGET)
     # as targets not defining it should not use TECHMAP_PATH ENV variable
     get_target_property(YOSYS_TECHMAP ${ARCH} YOSYS_TECHMAP)
 
+    # Device type specific cells and techmap
+    get_target_property(YOSYS_DEVICE_CELLS_SIM ${DEVICE_TYPE} CELLS_SIM)
+    get_target_property(YOSYS_DEVICE_CELLS_MAP ${DEVICE_TYPE} CELLS_MAP)
+
+    if (NOT "${YOSYS_DEVICE_CELLS_SIM}" MATCHES ".*NOTFOUND")
+        get_file_target(YOSYS_DEVICE_CELLS_SIM_TARGET ${YOSYS_DEVICE_CELLS_SIM})
```
litghost, line 1373: Just use "add_file_dependency" which does the same thing as these 3 lines?

environment.yml
```diff
@@ -13,6 +13,7 @@ dependencies:
   - symbiflow::icestorm
   - symbiflow::capnproto-java
   - litex-hub::gcc-riscv64-elf-newlib
+  - python=3.7
```
litghost, line 16: Why the drop to python 3.7?

litghost (changes requested):
Most of the new CMake functions need better documentation.  At a minimum, what do they do and how to call them.  A why would be good too

litghost (changes requested):
Please remove all `*` imports, e.g.:

quicklogic/common/utils/rr_utils.py:
```
from data_structs import *
```

litghost (changes requested):
This PR should probably be broken up, it is very large, and hard to review.

Suggestions for places to split the PR:
 - A PR to setup the basic directory structure and run v2x on Verilog.  I believe this PR would be relatively straight forward, and the generated output would be all the pb_types, etc.
 - A PR adding the python support libraries
 - A PR adding the new CMake scripts

litghost:
> Do you think that the solution with overwriting certain packages temporarily in the CI is a good approach?
> [9a90d28#diff-d9eb2e25be835d9549f2c62e1b9717dba5bce9f76572de38ca4e595ff9c4c60dR17](https://github.com/SymbiFlow/symbiflow-arch-defs/commit/9a90d28e68f0a38d97280a7ae11d3d59665c81d2#diff-d9eb2e25be835d9549f2c62e1b9717dba5bce9f76572de38ca4e595ff9c4c60dR17)

I think that is a terrible idea.

litghost:
> @litghost, can you provide any information on what should be added to integrate the QuickLogic support with SymbiFlow?
> Probably, your help is also required to set up proper CI.

I think the first step here is to make a small PR that adds the QuickLogic configuration and a dummy QL CI script that is an empty shell script.  We can then add that config to kokoro.  I then recommend we incrementally stage PR's, per my earlier suggestion.

I know this isn't a surprise, but it is practically impossible to review and merge a PR with 120k lines and ~500 files.  We need to break this into small PRs with more modest goals.  For example, having a PR that generates all the pb_types by invoking v2x is probably a good PR once we have the initial CI stuff setup.

litghost:
> I think the first step here is to make a small PR that adds the QuickLogic configuration and a dummy QL CI script that is an empty shell script. We can then add that config to kokoro. I then recommend we incrementally stage PR's, per my earlier suggestion.

Specifically this PR would have the following files:

```
.github/kokoro/continuous-ql.cfg
.github/kokoro/presubmit-ql.cfg
.github/kokoro/ql.sh
.github/kokoro/kokoro-cfg.py
```

I notice that in this PR you didn't modify `kokoro-cfg.py`, which seems wrong?  presubmit-ql.cfg / continuous-ql.cfg  should be generated from `kokoro-cfg.py`

rw1nkler:
> This PR should probably be broken up, it is very large, and hard to review.

Ok, I will split the PR then into smaller chunks.

> I think the first step here is to make a small PR that adds the QuickLogic configuration and a dummy QL CI script that is an empty shell script. We can then add that config to kokoro. I then recommend we incrementally stage PR's, per my earlier suggestion.

Here is the first PR with the initial CI setup: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1760

> Suggestions for places to split the PR:
> - A PR to setup the basic directory structure and run v2x on Verilog. I believe this PR would be relatively straight forward, and the generated output would be all the pb_types, etc.

This PR will require all of the `v2x` PRs to be merged:
    - https://github.com/SymbiFlow/python-symbiflow-v2x/pull/75
    - https://github.com/SymbiFlow/python-symbiflow-v2x/pull/32

From the discussions in the `v2x` PRs, I see that still some of the issues are unresolved. What's more, the solutions proposed there will require a lot of changes in the existing QuickLogic support.

litghost:
> This PR will require all of the `v2x` PRs to be merged:
> - [SymbiFlow/python-symbiflow-v2x#75](https://github.com/SymbiFlow/python-symbiflow-v2x/pull/75)
> - [SymbiFlow/python-symbiflow-v2x#32](https://github.com/SymbiFlow/python-symbiflow-v2x/pull/32)
> 
> From the discussions in the `v2x` PRs, I see that still some of the issues are unresolved. What's more, the solutions proposed there will require a lot of changes in the existing QuickLogic support.

Given your statement, that needs to be fixed first then?  So order for now should be:
 - Get relevant v2x PR's merged
 - Start creating PR's for the relevant python scripts
 - Work on updating verilog per v2x PR changes
 - etc

litghost:
@rw1nkler New QuickLogic job is up and running.  I tested the presubmit job in #1762 and hopefully the master job will be green too.

PR #1762, opened 2020-11-09
Change ql.sh to just echo a string and exit for now.

litghost:
`package_results.sh` is removed because it fails if there is nothing to package!


mithro (approved):
LGTM!

Issue #1768, opened 2020-11-11
CI is broken


HackerFoo:
It looks like the build has been broken since https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1760

HackerFoo:
Simple designs such as `counter` and `buttons` are unroutable. It could be a problem with VPR.

mithro:
@litghost / @acomodi - Can you please look into this?

acomodi:
@HackerFoo you have seen this for the vtr+Symbiflow CI, correct? In that case this PR solves the issue: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1711.
I have forced a Kokoro run as there were some infrastructure failures.

litghost:
If you are speaking about the vtr+Symbiflow, that is a result of in progress `master+wip` updates.  To be clear, there are 3 CI's on arch-defs, Travis, vtr+Symbiflow and arch-defs.  Both Travis and arch-defs should both be green right now.

HackerFoo:
Then VtR+SymbiFlow doesn't need to pass?

The top level "Architecture Definitions (Presubmit)" is also failing on https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1735, but I think it's because of the QuickLogic test.

Another strange thing is that several of the tests have not yet completed for the last 2 PRs.

litghost:
> The top level "Architecture Definitions (Presubmit)" is also failing on #1735, but I think it's because of the QuickLogic test.

If you rebase onto master, then the QuickLogic test will pass trivially.  Kokoro should be running as "presubmit as merged", but because it is part of the kokoro configuration it may be more restrictive.

> Then VtR+SymbiFlow doesn't need to pass?

[...]

PR #1769, opened 2020-11-11
Add Vivado logs to captured results for debugging purposes.

litghost:
Also fixes script when generating install collection lists.

Cleans up `third_party`, `.git` and `env` directories prior to kokoro doing artifact collection.

### DIFF
--- a/.github/kokoro/continuous-docs.cfg
+++ b/.github/kokoro/continuous-docs.cfg
@@ -19,6 +19,8 @@ action {
     regex: "**/route.log"
     regex: "**/*_sv2v.v.log"
     regex: "**/*_qor.csv"
+    regex: "**/vivado.log"
+    regex: "**/*.bit"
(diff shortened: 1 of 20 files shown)


mithro (approved):
LGTM - My only concern is that there might be too many log files?

litghost:
> LGTM - My only concern is that there might be too many log files?

I don't think that'll be an issue.  We'll see.  We should add a check that the working directory has less than 100 GB total.  This was likely causing the problems discussed in https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1768#issuecomment-725648866 and https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1747

It is annoying that kokoro doesn't generate a log when this error is occuring :(

litghost:
> We should add a check that the working directory has less than 100 GB total.

I added a check to make sure the working directory has less than 45 GB, which is hopefully a good limit.  I also printed the size of the working directory so we can trace how big it has gotten.

PR #1771, opened 2020-11-13
Revert "Adjust VPR flags to reduce runtime"

litghost:
Reverts SymbiFlow/symbiflow-arch-defs#1735


litghost:
> @litghost Why?

Interchange failures can sometimes result from misrouting, which your changes could've caused.  Given that the problem was the RapidWright open <-> closed mismatch, I closed the PR.  If #1772 shows an interchange failure caused by misrouting, I will reopen this PR.

HackerFoo:
Ah, sorry. I didn't realize that this was just for testing.

PR #1772, opened 2020-11-13
Update SymbiFlow/RapidWright fork instead of directly ingesting upstream

litghost:
A change on upstream RapidWright was added that depended on a new version of the closed portion of RapidWright, which is resulting in build failures.  For now, fork RapidWright and move the branch point back until the closed portion of RapidWright is updated.


mithro (approved):
Should we potentially be using conda packages for RapidRight?

Otherwise LGTM.

PR #1781, opened 2020-11-17
print_qor: sorted tests by name and remove prefix

acomodi:
This PR improves the `print_qor` script in the following ways:

- sorts tests, so that they are more easily readable
- takes only the last two directories in the path (to avoid common paths to be added for each test
- collects only the relevant run-times in the P&R flow, excluding the overhead.

### DIFF
--- a/utils/print_qor.py
+++ b/utils/print_qor.py
@@ -7,17 +7,16 @@ import sys
 import subprocess
 
 
-def scan_runtime(fname):
+def scan_runtime(step, fname):
     """ Find runtime of VPR log (if any), else returns empty str. """
     try:
         with open(fname, 'r') as f:
+            runtime = ""
             for line in f:
-                pass
+                if line.startswith("# {} took".format(step)):
+                    runtime = str(float(line.split()[3]))
 
-            if not line.startswith('The entire flow of VPR took'):
-                return ""
-
-            return str(float(line.split()[6]))
+            return runtime
     except FileNotFoundError:
         return ""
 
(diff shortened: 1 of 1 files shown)


utils/print_qor.py
```diff
@@ -78,23 +77,32 @@ def main():
...
     w = csv.DictWriter(sys.stdout, fields)
     w.writeheader()
+    rows = list()
     for root, dirs, files in os.walk(args.build_dir):
         if 'pack.log' in files:
             d = {}
 
-            d['path'] = root
-            d['pack time (sec)'] = scan_runtime(os.path.join(root, 'pack.log'))
+            path = root.split('/')
```
litghost, line 85: Factor this into a function like `last_n_dirs(root, 2)`, etc.

utils/print_qor.py
```diff
@@ -7,17 +7,16 @@
 import subprocess
 
 
-def scan_runtime(fname):
+def scan_runtime(step, fname):
```
litghost, line 10: Can you also output the overhead for each step.  So basically output step runtime and overhead runtime.  Make sense?

utils/print_qor.py
```diff
@@ -78,23 +77,32 @@ def main():
...
+            )
             d['place time (sec)'] = scan_runtime(
-                os.path.join(root, 'place.log')
+                "Placement", os.path.join(root, 'place.log')
             )
             d['route time (sec)'] = scan_runtime(
-                os.path.join(root, 'route.log')
+                "Routing", os.path.join(root, 'route.log')
             )
             d['t_crit (ns)'], d['Fmax (MHz)'] = scan_critical(
```
litghost, line 96: Request in meeting to add geomean CPD as well as CPD to table.

utils/print_qor.py
acomodi: Done

utils/print_qor.py
acomodi: Sure, makes sense. Done

utils/print_qor.py
acomodi: Added both CPD geomean and its corresponding frequency

litghost (commented):
Looks good.  I want to review the output from CI, so I'll revisit this once the CI's return with results.

acomodi:
@litghost All CI's went green:

for XC7, [here](https://storage.cloud.google.com/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/presubmit/xc7/3073/20201117-112047/build/xc7_qor.csv) is the output of print_qor.

PR #1782, opened 2020-11-17
Make Ibex sv2v conversion deterministic

rw1nkler:
This PR ensures that Ibex sv2v conversion is deterministic, by enforcing the original input file order (used in synthesis).

A file order during the sv2v conversion is important. When a file is provided earlier, it can force sv2v to generate additional
wrappers or resolve some advanced SystemVerilog syntax correctly (i.e., wildcards)

### DIFF
--- a/xc/xc7/tests/soc/ibex/generate.py
+++ b/xc/xc7/tests/soc/ibex/generate.py
@@ -65,7 +65,7 @@ def get_fusesoc_sources(root_dir, eda_yaml_path, f_log):
         eda_yaml_path=eda_yaml_path
     )
 
-    return set(
+    return list(
         s.decode() for s in
         subprocess.check_output(get_sources_invocation, shell=True).split()
     )


rw1nkler:
I encountered this problem working on the earlgrey example. 

I believe that the recent changes in Ibex CPD might be related to the sv2v conversion order (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1711#issuecomment-726248546)

xc/xc7/tests/soc/ibex/generate.py
```diff
@@ -65,7 +65,7 @@ def get_fusesoc_sources(root_dir, eda_yaml_path, f_log):
         eda_yaml_path=eda_yaml_path
     )
 
-    return set(
+    return list(
```
litghost, line 68: So rather than `set(` -> `list(`, change this to `sorted(set(`

rw1nkler: I believe that we want to preserve the original synthesis order, so in this case, `list()` seems to be more appropriate.

litghost:
Can you please rebase on master?

rw1nkler:
> Can you please rebase on master?

Done

mithro:
Can you do another rebase? That should make the `VtR + SymbiFlow Architecture Definitions (Presubmit)` builder go away.

PR #1787, opened 2020-11-18
Install 200t with other devices

acomodi:
This PR is to have the 200T installed alongside with all the other devices.

This is different than https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1747, as it does not split all the devices in their own packages.

I think the transition to different packages for each device should be performed once all the infrastracture is ready:
- Conda package that enables using the binary toolchain from archdefs
- SymbiFlow tool manager should be adapted to get the latest links of all devices

This PR is to have one single point of origin for the symbiflow installation package that can be used for instance in perf-tool or symbiflow-examples, without the need of changing the infrastructure to get the required devices and toolchain binaries


litghost:
I think I preferred the direction in #1747?  @mithro, thoughts?  @acomodi, why are we not finishing what was proposed in #1747?

acomodi:
> why are we not finishing what was proposed in #1747?

@litghost mainly to have a gradual change towards the goal. First I wanted to generate a single package including all the devices, including the 200T. Once that is stable we can land #1747 and propagate the changes to the various parts to have a full integration. By various parts I mean:
- Adding the binary toolchain to conda
- Changing the way we install the devices in symbiflow-examples and tool-perf
- Changing the way we get and store the latest install packages

litghost (approved):
LGTM, merge if green.

Issue #1788, opened 2020-11-18
Input sensitivity status


acomodi:
This issue is to keep track of the issue reported here: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1776.

## Problem statement

The auto-generated designs, such as litex, present some small variations in the memory initialization of the BRAMs. This causes a small perturbation in the initial conditions of a test, which lead to major changes in the output results.

This problem might not be only related to the BRAM initialization, but to any small changes that are applied to a specific design, but for the sake of this issue description, we will keep the small BRAM initialization changes as the triggering factor for this issue.

There are several levels that affect the final output results.

### Synthesis

The mem.init changes make yosys produce two completely different output eblifs. There are some tests that need to be performed in order to asses whether the two different eblifs represent the same circuit, and it is just a matter of ordering.
- run post_synthesis simulations
- apply sorting to the output blifs

The expected output of the synthesis step would be to have only a small number of lines affected, relative to the BRAM initializations.

### Packer

The perturbation in the nets ordering of the circuit might affect the way the packer acts. For instance, two different runs of the same test produced the following packing results:

**Test 1**:
```
Resource usage...
        Netlist
                1099    blocks of type: BLK-TL-SLICEL
        Architecture
                2150    blocks of type: BLK-TL-CLBLL_L
                1200    blocks of type: BLK-TL-CLBLL_R
                1800    blocks of type: BLK-TL-CLBLM_L
                3000    blocks of type: BLK-TL-CLBLM_R
```

**Test 2**:
```
Resource usage...
        Netlist
                1114    blocks of type: BLK-TL-SLICEL
        Architecture
                2150    blocks of type: BLK-TL-CLBLL_L
                1200    blocks of type: BLK-TL-CLBLL_R
                1800    blocks of type: BLK-TL-CLBLM_L
                3000    blocks of type: BLK-TL-CLBLM_R
```

The two Verilog descriptions of the design are identical, except for the memory initialization.

A possible solution to this behaviour is to make the packer algorithms more robust to changes in the ordering of the input circuit, making it less sensitive to changes in input conditions.

### Placer 

Initial placement is currently very sensitive to the seed. In fact, changing the seed can generate very different outcomes, in terms of CPD as well as router run-time.

This is proven by the following test, which, taken two initially equal packed netlists and two different SEEDs, the router behaviour drastically changed:

**default seed**:
```
[...]

PR #1790, opened 2020-11-19
litex: do not output timestamp in BIOS

acomodi:
This PR prevents LiteX designs to incur in non-deterministic P&R due to changes in the BIOS (details in https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1788).

With this PR the Build time timestamp is not being added to the mem.init files, but it requires using the latest versions of LiteX, which insert FD primitives in the design.

The FD have been retargeted with a techmap pass in the synth.tcl script as a temporary workaround, as this should rather be a fix to yosys upstream.

### DIFF
--- a/xc/xc7/tests/soc/litex/generate.py
+++ b/xc/xc7/tests/soc/litex/generate.py
@@ -33,7 +33,7 @@ class BaseSoC(SoCCore):
             platform,
             sys_clk_freq,
             ident="LiteX SoC on Arty A7",
-            ident_version=True,
+            ident_version=False,
             **kwargs
         )
 


mithro:
@acomodi Can you add the techmap change in a separate pull request so it is easy to revert when the time comes?

acomodi:
> Can you add the techmap change in a separate pull request so it is easy to revert when the time comes?

@mithro Unfortunately not, as the new litex version that I needed to bump generates the `FD` primtives in the design which will let every litex test to fail.

litghost:
> > Can you add the techmap change in a separate pull request so it is easy to revert when the time comes?
> 
> @mithro Unfortunately no, as the new litex version that I needed to bump generates the `FD` primtives in the design which will let every litex test to fail.

@mithro The `FD` change is already a separate commit, so it should be easy enough to remove it in the future.

mithro:
@acomodi it just means we have to merge the FD pull request before the LiteX pull request?

acomodi:
> @acomodi it just means we have to merge the FD pull request before the LiteX pull request?

@mithro Totally right, open a PR here https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1792 and based this one on top of that one.

litghost (approved):
LGTM, merge if green.  I'm re-run the failing CI

PR #1791, opened 2020-11-19
prjxray: use conda package instead of third_party dir

acomodi:

This PR is intended to use the conda packages for the prjxray-tools and prjxray-db:

- prjxray-db: instead of the third party dir, it gets the database from the conda-package
- prjxray-tools: locally conflicts with the riscv64-gcc-newlib package from `litex-hub`. The conflicts are being fixes, and once it's done, I will ad the prjxray-tools conda package as well

The reason behind using the conda packages is that both symbiflow-examples and fpga-perf-tool use them, so, to use the same environment.yml file as used in symbiflow-arch-defs, we require having the prjxray packages here as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,8 +62,6 @@ get_target_property_required(PYTHON3 env PYTHON3)
 include(common/cmake/image_gen.cmake)
 include(common/cmake/gen.cmake)
 
-add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/third_party/prjxray)
-
 # Target for all check tests.
 add_custom_target(all_check_tests)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,9 @@ add_env_executable(EXE flake8 REQUIRED)
 add_env_executable(EXE vvp REQUIRED)
 add_env_executable(EXE iverilog REQUIRED)
 add_env_executable(EXE xcfasm REQUIRED)
+add_env_executable(EXE xc7frames2bit REQUIRED)
+add_env_executable(EXE bitread REQUIRED)
+add_env_executable(EXE prjxray-config)
 
 get_target_property_required(PYTHON3 env PYTHON3)
 
--- a/xc/xc7/CMakeLists.txt
+++ b/xc/xc7/CMakeLists.txt
@@ -2,13 +2,25 @@ set(XC7_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(PRJXRAY_DIR ${symbiflow-arch-defs_SOURCE_DIR}/third_party/prjxray
   CACHE PATH "Path to prjxray library")
 
-execute_process(
-  COMMAND
-    prjxray-config
-  OUTPUT_VARIABLE PRJXRAY_DB_DIR
-)
+get_target_property_required(PRJXRAY_CONFIG env PRJXRAY_CONFIG)
 
-string(STRIP "${PRJXRAY_DB_DIR}" PRJXRAY_DB_DIR)
+if (NOT ${PRJXRAY_CONFIG} STREQUAL "PRJXRAY_CONFIG-NOTFOUND")
+  execute_process(
+    COMMAND
+      bash ${PRJXRAY_CONFIG}
+      OUTPUT_VARIABLE PRJXRAY_DB_DIR
+  )
+  string(STRIP "${PRJXRAY_DB_DIR}" PRJXRAY_DB_DIR)
+else ()
+  set(PRJXRAY_DB_DIR "PRJXRAY_DB_DIR-NOTFOUND")
+endif ()
+
+set(PRJXRAY_DB_DIR "${PRJXRAY_DB_DIR}"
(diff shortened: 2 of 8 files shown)


xc/xc7/CMakeLists.txt
```diff
@@ -1,9 +1,14 @@
 set(XC7_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(PRJXRAY_DIR ${symbiflow-arch-defs_SOURCE_DIR}/third_party/prjxray
   CACHE PATH "Path to prjxray library")
-set(PRJXRAY_DB_DIR
-  ${symbiflow-arch-defs_SOURCE_DIR}/third_party/prjxray-db
-  CACHE PATH "Path to prjxray database files")
+
+execute_process(
+  COMMAND
+    prjxray-config
```
litghost, line 7: You should use `find_program` to locate `prjxray-config`.   This causes CMake to report if `prjxray-config`  is missing from the path

litghost: Make sure to specific the "DOC <description>" argument, and REQUIRED argument.  See https://cmake.org/cmake/help/latest/command/find_program.html for details.

tcal-x:
Will there be a way that a developer can use their own prjxray-db (to test changes before the official prjxray-db is rebuilt)?

acomodi:
> Will there be a way that a developer can use their own prjxray-db (to test changes before the official prjxray-db is rebuilt)?

Good point, with this change the only way would be to manually modify the content of the database directory within the conda environment. I may add a CMAKE variable to control which prjxray-db to use, and set it to the conda-package db by default

litghost:
> Will there be a way that a developer can use their own prjxray-db (to test changes before the official prjxray-db is rebuilt)?

So `prjxray-config` is just a shell script the returns that path to the directory.  If you pointed `prjxray-config` at something akin to:
```
echo ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}
```

Another option would be to keep the `PRJXRAY_DB_DIR` cache variable, and have the default either the result from `prjxray-config` (if on path) or `PRJXRAY_DB_DIR-NOTFOUND` and generate a config error if the user doesn't supply it.  I think this is more flexible, and is easy enough to support both use cases?

litghost:
> > Will there be a way that a developer can use their own prjxray-db (to test changes before the official prjxray-db is rebuilt)?
> 
> Good point, with this change the only way would be to manually modify the content of the database directory within the conda environment. I may add a CMAKE variable to control which prjxray-db to use, and set it to the conda-package db by default

I don't think we want to have the CMake script explicitly know about conda.  The idea of `prjxray-config` was to simply have something on the path that provides the needed information.  Because we are in CMake, I think rather than having a CMake variable to control whether it uses `prjxray-config`, just keep the current CMake variable, but change how the default is determined (per my earlier comment).

acomodi:
> can you implement this second form?

@litghost Done, now it is as follows:

- if `prjxray-config` is found, set PRJXRAY_DB_DIR default to its output, otherwise, the default would be `PRJXRAY_DB_DIR-NOTFOUND`
- if the user does not set PRJXRAY_DB_DIR in the cmd line, the error is produced.

I have also added `prjxray-tools` to the conda packages so to use `bitread` and `xc7frames2bit` from there instead of the one in the third_party dir.

xc/xc7/CMakeLists.txt
acomodi: I have actually added prjxray-config with the `add_env_executable` function

CMakeLists.txt
```diff
@@ -53,6 +53,9 @@ add_env_executable(EXE flake8 REQUIRED)
 add_env_executable(EXE vvp REQUIRED)
 add_env_executable(EXE iverilog REQUIRED)
 add_env_executable(EXE xcfasm REQUIRED)
+add_env_executable(EXE xc7frames2bit REQUIRED)
+add_env_executable(EXE bitread REQUIRED)
+add_env_executable(EXE prjxray-config)
 
 get_target_property_required(PYTHON3 env PYTHON3)
```
litghost, line 60: Should remove `third_party/prjxray` on line 65 below?

litghost (changes requested):
Comments in

CMakeLists.txt
acomodi: prjxray is still needed as it contains all the various python scripts to access the db. The prjxray-tools conda-package instead contains only the bit tools

litghost: I didn't say remove the submodule, I said from `CMakeLists.txt`.  Do you believe there are any remaining CMake targets used from `third_party/prjxray`?

acomodi: Aha, my bad, I misunderstood. Anyway no, looking at the CMake targets in prjxray, it seems that those are only needed to build the tools, therefore yes, that line can be removed.

acomodi: Fixed

xc/xc7/CMakeLists.txt
acomodi: Done

PR #1792, opened 2020-11-19
yosys: add retargeting step for FD primitives

acomodi:

### DIFF
--- a/xc/xc7/techmap/retarget.v
+++ b/xc/xc7/techmap/retarget.v
@@ -1,0 +1,7 @@
+module FD (output reg Q, input C, D);
+
+parameter [0:0] INIT = 1'b0;
+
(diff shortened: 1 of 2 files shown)


mithro (approved):
@acomodi - Just an FYI, having a good library for retargeting Xilinx primitives is what the old code which was generating stuff from the Xilinx documentation was for.

mithro:
@acomodi This stuff --> https://github.com/SymbiFlow/symbiflow-arch-defs/tree/master/xc/common/libraries

mithro:
@acomodi -- Go ahead and land this if CI looks green.

I think there is a larger discussion about how retargetting like this should work. I **think** we actually want something which converts the Xilinx specific `FD` cell into the Yosys primitive, not another Xilinx primitive. This should happen just after the netlist creation before any of the later stages.

PR #1795, opened 2020-11-20
Switch to packages from 'litex-hub' channel

tjurtsch:
Most of the Conda dependencies are now provided via 'litex-hub' channel.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,18 +1,17 @@
 name: symbiflow_arch_def_base
 channels:
   - defaults
-  - symbiflow
   - litex-hub
 dependencies:
-  - symbiflow::symbiflow-yosys=0.8_6021_gd8b2d1a2=20200708_083630
-  - symbiflow::symbiflow-yosys-plugins=1.0.0.7_0174_g5e6370a=20201012_171341
-  - symbiflow::symbiflow-vtr=8.0.0.rc2_5415_gd6d69ff92=20201120_180018
-  - symbiflow::zachjs-sv2v=0.0.5_0018_ga170536
-  - symbiflow::openocd
-  - symbiflow::iverilog
-  - symbiflow::icestorm
-  - symbiflow::capnproto-java
-  - litex-hub::gcc-riscv64-elf-newlib
+  - litex-hub::capnproto-java=0.1.5_0012_g44a8c1e=20201104_165332
+  - litex-hub::gcc-riscv64-elf-newlib=9.2.0=20201119_154229
+  - litex-hub::icestorm=0.0_0719_g792cef0=20201120_145821
+  - litex-hub::iverilog=s20150603_0957_gad862020=20201120_145821
+  - litex-hub::openocd=0.10.0_1514_ga8edbd020=20201119_154304
+  - litex-hub::symbiflow-yosys=0.8_6021_gd8b2d1a2=20201120_145821_libffi33
+  - litex-hub::symbiflow-yosys-plugins=1.0.0_7_g59ff1e6_23_g3a95697_17_g00b887b_0194_g40efa51=20201120_145821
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -4,14 +4,14 @@ channels:
...
 dependencies:
-  - symbiflow::symbiflow-yosys=0.8_6021_gd8b2d1a2=20200708_083630
-  - symbiflow::symbiflow-yosys-plugins=1.0.0.7_0174_g5e6370a=20201012_171341
-  - symbiflow::symbiflow-vtr=8.0.0.rc2_5097_gf1a3bcc2a=20200916_072439
-  - symbiflow::zachjs-sv2v=0.0.5_0018_ga170536
-  - symbiflow::openocd
-  - symbiflow::iverilog
-  - symbiflow::icestorm
-  - symbiflow::capnproto-java
+  - litex-hub::symbiflow-yosys
```
litghost, line 7: Regardless of whether the packages come from symbiflow or litex-hub channel, removing version pinning right now will result in unstable builds.  What is the plan to deal with incompatible packages?  I'm specifically concerned with VTR and yosys packages, as those are the ones mostly likely to break.

mithro:
@PiotrZierhoffer - I think we need the version pinning process here (and auto rolling bot).

ajelinski:
@litghost @mithro All `litex-conda-*`-based packages have been pinned to their latest versions available in the `litex-hub` channel. Being the latest ones most of those versions must've been tested in the previous CI of this PR so I hope this would be successful too.

`symbiflow` channel has been removed because currently all packages `defaults` channel lacks are taken from the `litex-hub` channel. I also rebased commits on top of the current master.

ajelinski:
@litghost are you able to restart `Documentation` and `Architecture Definitions` tests alone? AD failed only because the `Documentation` failed which had some error which seems to be some random Conda or connection issue:
```
CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://repo.anaconda.com/pkgs/main/linux-64/repodata.json>
Elapsed: -

An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.
```

ajelinski:
The branch was rebased again because there was a conflict on `symbiflow-vtr` version which needed to be resolved.

The `symbiflow-vtr` basically stays the same after the change because it was built using the latest `d6d69ff` commit. It will be just taken from the `litex-hub` channel.

litghost (changes requested):
With the move to `litex-hub`, how are changes to packages introduced?  What is the new process?  For example, what does https://github.com/SymbiFlow/conda-packages/pull/143 look like?

Also the `litex-hub` packages are only built with `python 3.7`, and the packages from `SymbiFlow/conda-packages` are only built with `python 3.8`.  Should the `litex-hub` packages be built with `python 3.7` and `python 3.8`?  Should `litex-hub` simply move from `python 3.7` to `python 3.8`?

Until these questions are answered, this PR cannot be merged.  If these questions have already been asked and answered, where are the answers?  Thanks.

litghost:
So as a concrete question, how does the labelling on litex-hub work?  When do packages get the `main` label?

PiotrZierhoffer:
@litghost I added a description of the transition we want to have: https://github.com/SymbiFlow/conda-packages/issues/154

The process of merging in SymbiFlow/conda-packages#143 will look, I assume, quite similar. 

I'll verify the Python thingy. Is conda-packages actually relying on 3.8?

Labels are attached to packages when they are built as part of a PR. When they get merged and the CI is green, they are moved to main.

litghost:
Thanks @piotr-binkowski and @ajelinski, I've take this PR and iterated on it in #1816

litghost:
Superceed #1816

ajelinski:
Thanks @litghost, it's great to have it merged.

PR #1796, opened 2020-11-20
Package also circuits

acomodi:
This PR is to add also the selected tests' circuits to the install directory.

It is possible to add a new test to the installed ones through the CMake option `INSTALL_CIRCUIT` when calling the `add_fpga_target` function.


common/cmake/devices.cmake
```diff
@@ -1447,16 +1456,40 @@ function(ADD_FPGA_TARGET)
...
 
   if(NOT "${ADD_FPGA_TARGET_INPUT_SDC_FILE}" STREQUAL "")
     append_file_dependency(VPR_DEPS ${ADD_FPGA_TARGET_INPUT_SDC_FILE})
     get_file_location(SDC_LOCATION ${ADD_FPGA_TARGET_INPUT_SDC_FILE})
     set(SDC_ARG --sdc_file ${SDC_LOCATION})
+    set(SDC_FILE ${SDC_LOCATION})
+  endif()
+
+  if (${ADD_FPGA_TARGET_INSTALL_CIRCUIT})
+    add_custom_target(
```
litghost, line 1475: Not sure what this custom target is for?

.github/kokoro/continuous-install-200t.cfg
```diff
@@ -11,7 +11,7 @@ timeout_mins: 4320
 action {
   define_artifacts {
     # File types
-    regex: "**/symbiflow-arch-defs-install-200t*.tar.xz"
+    regex: "**/symbiflow-arch-defs-*.tar.xz"
```
litghost, line 14: Don't see a change to the template python script?

HackerFoo (approved):

mithro (approved):
LGTM if you fix @litghost's stuff.

.github/kokoro/continuous-install-200t.cfg
acomodi: Done.

I am thinking whether we need the install-200t CI as the 200T is now built and packaged in the install one.

common/cmake/devices.cmake
acomodi: It was to build the dependency to install the eblif, but I think it is not needed actually. Removed

acomodi: Actually this target is needed when calling the install target, otherwise the eblifs do not get built.

acomodi:
Install targets fail with no log output: [Link to failed build](https://source.cloud.google.com/results/invocations/e378b652-7596-4d03-b638-01bc66627d24/log).

This seems similar to the problems had when packaging the 200T alongside with the other devices: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1747#issuecomment-721100912

litghost:
> Install targets fail with no log output: [Link to failed build](https://source.cloud.google.com/results/invocations/e378b652-7596-4d03-b638-01bc66627d24/log).

```
CMake Error at xc/xc7/tests/soc/ibex/cmake_install.cmake:50 (file):
  file INSTALL cannot find
  "/tmpfs/src/github/symbiflow-arch-defs-presubmit-install/build/xc/xc7/tests/soc/ibex/pins_artya7.sdc":
  No such file or directory.
  xc/xc7/tests/soc/cmake_install.cmake:52 (include)
  xc/xc7/tests/cmake_install.cmake:177 (include)
  xc/xc7/cmake_install.cmake:88 (include)
  xc/cmake_install.cmake:52 (include)
  cmake_install.cmake:72 (include)
```

acomodi:
@litghost All right, thanks. Something wrong in the PR then, I'll fix that

PR #1797, opened 2020-11-20
Remove connection box from rr-graph patching.

litghost:


acomodi (approved):
LGTM. 

It's worth opening an issue to keep track of removing the connection box annotations from the channels/edges generation step.

acomodi:
Here is a spreadsheet with the results for 50/100T and 200T: [results](https://docs.google.com/spreadsheets/d/1-_w2GmlLVRQWALYvq44sd8-moWJhl13UNporD2QjC8c/edit?usp=sharing)

They look good in general, with the biggest tests having better QoR in average.

(Note that the `t_crit_geomean` metric is a more robust indicator of CPD, as it might happen that a different clock CPD gets displayed in `t_crit`

PR #1806, opened 2020-11-25
ci: move from travis to GH actions

acomodi:

Travis has greatly slowed down, therefore this PR is to move the Travis CI to use GH actions instead.


acomodi:
@mithro @kgugala I have moved Travis to GH actions:

removed scripts:
- git-fixup: seemed to not be useful as the `git describe` command never gets used, and, without this step, CI passes correctly
- after_success/failure: these steps were only printing a `TODO!`
- compare-files: never called by any step
- before-cache: never called by any step

maintained scripts:
- script.sh: runs the whole test bench

mithro:
FYI - @HackerFoo

mithro (approved):
LGTM?

kgugala (approved):
LGTM

umarcor:
@acomodi, see https://github.com/antmicro/symbiflow-arch-defs/compare/move-to-gh-actions...umarcor:move-to-gh-actions. Apply/ignore them freely.

- I added the recursive option to the checkout step because this repository has many submodules, and that's what `git-fixup.sh` did (although, as you said, it was not executed). Then, I saw that it's done in the `Makefile`. I am unsure whether cloning ahead of time is useful, or it will be cloned elsewhere.
- There is no `action_fold` feature exactly comparable to Travis' `travis_fold`. Travis allows nested folds/groups, GitHub Actions does not. Nevertheless, it seems that nested groups/folds are not used in this repo. Hence the replacement should work.
  - At the same time, Travis shows the execution time at the right of the folded group, but GitHub does not have such feature.
  - I implemented the closest I could using `::group::`, `::endgroup::` and `SECONDS`.
- *Supressing* notes are directly changed from "cannot be built on travis" to "cannot be built on GH actions". Is that correct? I mean, should it be because execution time, it might be possible to run them on GH. Otherwise, if it's a disk size constraint, it might not. Anyway, I think it's desirable to explain the reason in the comments.
- One *supressing* note is followed by a commented out make call, which makes sense. However, two other notes are printed right before make calls without any specific modification/tweak. Do these comments still apply? Where are "some xml linting" and "some demo bitstreams" supressed?
  - I enabled `make all_rrgraph_xmls` and removed the note.

See example execution: https://github.com/umarcor/symbiflow-arch-defs/runs/1457116366?check_suite_focus=true

acomodi:
@umarcor Thanks for the detailed explanations and the time spent on this. Given that all CIs are currently all green and we need to move ASAP to GH Actions I will merge this PR and open a follow-up with your changes which are valuable.

PR #1807, opened 2020-11-26
Move to gh actions followup

acomodi:
This PR is a follow-up to https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1806, with some additional refinements to the GH actions scripts and config file by @umarcor, as explained in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1806#issuecomment-734073120.


.github/ci/test.sh
```diff
@@ -22,48 +18,27 @@ start_section "info.conda.config" "Info on ${YELLOW}conda config${NC}"
...
+make all_rrgraph_xmls
 end_section "symbiflow.build_all_rrgraph_xmls"
 
-$SPACER
-
 make_target all_route_tests "Complete all routing tests"
 
-$SPACER
-
 echo "Suppressing some xml linting, as the 5k/8k parts cannot be built on GH actions."
```
umarcor, line 35: This row and L43 are the ones which I don't know whether to preserve.

acomodi: I think it is better to preserve them to keep the same functionality as we had with Travis CI

umarcor: My point is that I don't know if those comments are actually meaningful. That is, what is the "supressed linting" the comment refers to? Which would be the alternative for not supressing it? The command right below is `make_target all_xml_lint`, which suggests that the comment might be outdated.

umarcor: The same question applies to L43-L44: `make_target all "Building all demo bitstreams"`.

acomodi: So, the fact is that we are not disabling those targets, but we are not including some of the devices in the GH/Travis CI. For instance, the whole `xc` directory does not get built: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/0e9281577f10ba2e2688c1de51e45eaa81302dea/CMakeLists.txt#L86-L89.

These comments do refer to this, IIRC.

umarcor: Thanks a lot for clarifying! Might be worth appending `See usage of CI envvar in CMakeLists.txt`.

PR #1808, opened 2020-11-26
Partial bitstream support

kowalewskijan:
This PR adds partial bitstream generation support to the `switch_processing` example.

Submodule dependencies:
prjxray: https://github.com/SymbiFlow/prjxray/pull/1492
xc-fasm: https://github.com/SymbiFlow/xc-fasm/pull/6


litghost:
The changes in this PR look okay, but I think a couple things are missing.

First, there needs to a be a usage document on how to use this flag safely.  Just reading the comment in the cmake file doesn't seem nearly enough.  For example, there is no discussion on what happens if the partial reconfiguration region overlaps a frame address, etc.

Second, there doesn't appear to be a test that verifies the partial reconfiguration files work as intended.  For example, I would expect that you'd want to do something like a simulation on fasm2bits before and after the partial reconfiguration was applied that shows the different behavior.

Third, has this been tested in hardware?  If so, how was that test done, and how could others redo the test?

litghost (commented):
Comments in

PR #1809, opened 2020-11-27
Bump prjxray 

syed-ahmed:
This would let us use NodeModel from prjxray.


syed-ahmed:
Oops. Closing this in favor of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1801.

PR #1813, opened 2020-11-27
ioddr: fix techmap and fasm features

acomodi:

This PR fixes some problems with the IDDR/ODDR techmapping:

- ODDR:
  - The D2 input never got the correct inverted parameter
  - Incorrect assignment of constant values of the input ports. When a zero-constant was used, the signal should be bound to VCC and the cell inverter enabled

- IDDR:
  - The `SAME_EDGE_PIPELINED` feature is enabled by default and it requires no fasm feature, but the techmapping incorrectly assigned either `SAME_EDGE` or `OPPOSITE_EDGE` with the result of never getting the `SAME_EDGE_PIPELINED` mode.

This change requires fixes in fasm2bels: https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/36. Without the fasm2bel fix for the `EDGE` feature, fasm2bels fails.


litghost (changes requested):
This PR needs to add a test with IDDR `DDR_CLK_EDGE` = `SAME_EDGE_PIPELINED`?

litghost:
Also some tests of the `D` inversion?

acomodi:
> This PR needs to add a test with IDDR DDR_CLK_EDGE = SAME_EDGE_PIPELINED?

This should alredy be tested. We actually never properly test it as it ended up being a SAME_EDGE parameter due to a double bug in the techmap as well as in fasm2bels.

https://github.com/SymbiFlow/symbiflow-arch-defs/blob/0e9281577f10ba2e2688c1de51e45eaa81302dea/xc/xc7/tests/iddr_oddr/iddr_basys3.v#L36-L39

> Also some tests of the D inversion?

Yes, this is actually not tested, I'll fix it.

PR #1817, opened 2020-11-30
Add memory and disk space paging.

litghost:


.github/kokoro/steps/start_monitor.sh
```diff
@@ -0,0 +1,10 @@
+(
+    while :
+    do
+        date
+        uptime
+        free -h
+        sleep 300
+    done
+) &
+MONITOR=$!
```
mithro, line 10: BTW You can actually set up a trap so the kill is called when the script ends...

https://stackoverflow.com/questions/360201/how-do-i-kill-background-processes-jobs-when-my-shell-script-exits

litghost: Yep, done!

mithro (approved):
LGTM!

PR #1827, opened 2020-12-01
Fix pip requirements

### LINKED ISSUE #1826: New pip resolver breaks installation
The new pip resolver breaks the current `requirements.txt`.  In particular, installing a package in edit mode is not compatible with git URL eggs.

litghost:
Fixes #1826

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,18 +7,12 @@
 [submodule "third_party/icestorm"]
 	path = third_party/icestorm
 	url = https://github.com/cliffordwolf/icestorm.git
-[submodule "third_party/fasm"]
-	path = third_party/fasm
-	url = https://github.com/SymbiFlow/fasm.git
 [submodule "third_party/symbiyosys"]
 	path = third_party/symbiyosys
 	url = https://github.com/YosysHQ/SymbiYosys.git
 [submodule "third_party/python-sdf-timing"]
 	path = third_party/python-sdf-timing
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -7,18 +7,12 @@
...
-[submodule "third_party/fasm"]
-	path = third_party/fasm
-	url = https://github.com/SymbiFlow/fasm.git
 [submodule "third_party/symbiyosys"]
 	path = third_party/symbiyosys
 	url = https://github.com/YosysHQ/SymbiYosys.git
 [submodule "third_party/python-sdf-timing"]
 	path = third_party/python-sdf-timing
 	url = https://github.com/SymbiFlow/python-sdf-timing.git
-[submodule "third_party/vtr-xml-utils"]
```
mithro, line 19: Isn't this used to generate the XML output compatible for VPR?

litghost: So as alluded to in the issue, you can no longer specify a package via `-e <local package>` and as `<package>@git+...`.  This causes the pip resolver to go insane and literally download every version of every package to try to resolve the conflict.  It is bizzare.

litghost: To answer your question more precisely.  `vtr-xml-utils` is required for `python-symbiflow-v2x`.  `python-symbiflow-v2x` uses the `<package>@<location>` syntax: https://github.com/SymbiFlow/python-symbiflow-v2x/blob/3cc21667a616f2cf5d69d436a8704022bc87bb1c/setup.py#L34

PR #1830, opened 2020-12-02
install: circuits: add also place constraints

acomodi:

This PR adds the placement constraints to the benchmarks tarball.

There are some tests that strictly require these constraints (e.g. Litex with DDR) to have all the IOs and clock tiles placed correctly, and to pass the symbiflow tests in upstream VTR.


litghost (approved):
LGTM, merge if green

PR #1836, opened 2020-12-03
Use parallel compression

litghost:
`xz` takes a while, and `pixz` might drop the runtime by taking advantage of multiple cores.


mithro (approved):
Is the tar output still deterministic?

mithro (approved):
LGTM

PR #1838, opened 2020-12-03
Fix missing width on RAMB18E1 techmap.

litghost:


xc/xc7/techmap/cells_map.v
```diff
@@ -1034,43 +1035,47 @@ module RAMB18E1 (
...
         && WRITE_WIDTH_A != 1
+        && WRITE_WIDTH_A != 2
         && WRITE_WIDTH_A != 4
         && WRITE_WIDTH_A != 9
         && WRITE_WIDTH_A != 18)
-        $error("Invalid WRITE_WIDTH_A: ", WRITE_WIDTH_A);
+        _TECHMAP_FAIL_ <= GENERATE_ERROR;
     if(WRITE_WIDTH_B != 0
         && WRITE_WIDTH_B != 1
+        && WRITE_WIDTH_B != 2
```
mithro, line 1060: This line is the actual change right?

mithro (commented):
The `GENERATE_ERROR` update is separate to the real width addition right?

litghost:
> The `GENERATE_ERROR` update is separate to the real width addition right?

Yosys still doesn't support `$error`, so trying to use it doesn't work and doesn't generate a useful error diagnostic.

mithro:
> > The `GENERATE_ERROR` update is separate to the real width addition right?
> 
> Yosys still doesn't support `$error`, so trying to use it doesn't work and doesn't generate a useful error diagnostic.

Yeah, but it sounds like these should be separate commits?

litghost:
> Yeah, but it sounds like these should be separate commits?

Sure, done.

mithro (approved):
LGTM!

PR #1843, opened 2020-12-04
install: fix typo in package device package generation

acomodi:


mithro (approved):
LGTM

CI failed? If it goes green merge ahead.

acomodi:
Yeah, infrastructure failure:

```
Exhausted retries waiting for bash to return an integer exit code, returning exit code 137 instead. This is not a Kokoro error! This usually means: 1) Your build was killed by something in the VM 2) Exited normally, but that the parent shell of the nohup'd build process is IO blocked and could not write out the exit_code_file in 60 seconds. In this case be sure that all processes launched by your build are properly cleaned up before exiting to free up resources and reduce IO contention, or 3) There was a BSOD, the VM rebooted, the builder has reconnected but the process is of course gone.
```

I'll force a rerun

litghost:
Merged this fix in #1848 , after rebasing.

PR #1847, opened 2020-12-04
Move HOME to /tmpfs.

### LINKED ISSUE #1670: Kokoro disk is full

litghost:
Fixes #1670

### DIFF
--- a/.github/kokoro/steps/movehome.sh
+++ b/.github/kokoro/steps/movehome.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+echo
+echo "========================================"
+echo "Moving home directory."
+echo "----------------------------------------"
+echo
+echo "Old home directory: ${HOME}"
+echo
+mkdir home
+cp -r $HOME/. home/
+export HOME=$(pwd)/home
+echo "New home directory: ${HOME}"
+echo "                    $(echo ~/)"
--- a/.github/kokoro/common.sh
+++ b/.github/kokoro/common.sh
@@ -27,6 +27,7 @@ exec 2>&1
 cd github/$KOKORO_DIR/
 
 # Run the common setup steps
+source ./.github/kokoro/steps/movehome.sh
 source ./.github/kokoro/steps/hostsetup.sh
 source ./.github/kokoro/steps/hostinfo.sh
 source ./.github/kokoro/steps/git.sh


.github/kokoro/steps/movehome.sh
```diff
@@ -0,0 +1,14 @@
...
+
+echo
+echo "========================================"
+echo "Moving home directory."
+echo "----------------------------------------"
+echo
+echo "Old home directory: ${HOME}"
+echo
+mkdir home
+cp -r $HOME/. home/
```
mithro, line 11: Why not `mv` and symlink?
```shell
mv $HOME /xxxx
ln -s /xxx/$(basename $HOME) $HOME
```

mithro (commented):
Should we just make this part of the Kokoro image itself?

.github/kokoro/steps/movehome.sh
litghost: Because that would affect the state of the VM for the next job.  That is a bad idea.

mithro: If I understand correctly the machines are throw away after they are used.

litghost: What do you see as the advantage of `mv` over `cp`.  Given that these locations are on different disks, there is no speed advantage to `mv`, and `mv` may glitch if something reloads `~`/`HOME`.  I think the `cp` is the better option.

mithro: With the cp you have to do a rm and then ln afterwards?

litghost: The `rm` and `ln` is not required at all?

mithro (approved):

.github/kokoro/steps/movehome.sh
mithro: Was worried about things that might not be using `$HOME` for home resolution.

PR #1849, opened 2020-12-05
Remove install directory once tarballs are created.

litghost:


mithro (approved):
LGTM I guess? Why do we need this?

litghost:
> LGTM I guess? Why do we need this?

Because kokoro is dumb, and copies everything in the working directory (no matter the collection filter).  The uncompressed install directory is 28G, which takes a non-trival amount of time to download.  Also kokoro has a 50/100GB max working directory size before it just fails with no warning.

mithro:
> > LGTM I guess? Why do we need this?
> 
> Because kokoro is dumb, and copies everything in the working directory (no matter the collection filter). The uncompressed install directory is 28G, which takes a non-trival amount of time to copy to the result collection host. Also kokoro has a 50/100GB max working directory size before it just fails with no warning.

Yay. Go ahead and merge.

PR #1860, opened 2020-12-08
Fix PS7 tests

acomodi:
PS7 tests for zynq were disabled on the XC7 target, for an issue which was closed a while ago: #548.

In addition, the PS7 instantiation has been cleaned, removing all the input unconnected ports, as they are automatically connected to `$false` during techmap.


mithro (approved):
LGTM if someone else also approves.

kgugala (approved):
LGTM

PR #1870, opened 2020-12-10
oddr: fix inv and zinv techmaps

acomodi:

This is a follow-up to https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1813 and fixes one more issue related to the inverted pins of the ODDR.

In fact, depending on which type of the ODDR (`TDDR` or `ODDR`), the inverted pins should have different features:

- TDDR: `ZINV_T`
- ODDR: `IS_D_INVERTED`


mkurc-ant (approved):
LGTM

PR #1251, opened 2019-12-25
Zybo: Add minitest with Litex/Vex booting Linux

kgugala:
The system uses PS7's DDR and is able to boot Linux.

PR #1874, opened 2020-12-11
Zybo: Add minitest with Litex/Vex booting Linux (Rebased)

acomodi:
This PR supersedes https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1251. It is rebased on top of the current master.

Routing seems to take a reasonable amount of time (~8 minutes) w.r.t. the previous tests which took a couple of hours.

I still need to verify whether litex it is correctly working on HW, but the design can be placed and routed correctly.


xc/xc7/tests/ps7/linux_litex/AxiPeriph.v
```diff
@@ -0,0 +1,842 @@
+module AXI4LiteCSR( // @[:@3.2]
```
litghost, line 1: Is this generated code?  If so, can we generate this using the existing dependencies?

litghost (changes requested):
Does this test something in arch-defs that we have a coverage gap on?  If there is no coverage gap, can this example be moved to symbiflow-examples (and fpga-tool-perf) rather than added to arch-defs?

xc/xc7/tests/ps7/linux_litex/VexRiscv_LinuxNoDspFmax.v
```diff
@@ -0,0 +1,7469 @@
+// Generator : SpinalHDL v1.3.6    git head : 9bf01e7f360e003fac1dd5ca8b8f4bffec0e52b8
```
litghost, line 1: This also looks like generated code?  Same question as the AXI example.  If using the baked verilog is typical, should this go in third_party, per @mithro 's recommendation?

mithro: Correct.

xc/xc7/tests/ps7/linux_litex/AxiPeriph.v
acomodi: This is generated, but the problem is that it uses some additional dependencies (scala) which would only be needed to build this module.

acomodi:
> Does this test something in arch-defs that we have a coverage gap on?

Actually no, I'll move this into examples (and tool-perf) as you suggested, as this design does not really test anything that is already being tested (for the PS7 connection there is already the AXI tester and the simple EMIO-based counter)

PR #1880, opened 2020-12-14
Add ANTLR4 dependencies for fasm.

HackerFoo:


litghost:
```
========================================
Host install packages
----------------------------------------
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package libantlr4-runtime-dev
```

litghost:
I don't believe build the antlr4 runtime within the arch-defs CI is the right approach here, especially given the fact that 4 other repos will need to do the same thing.  I think using a PPA to get a package or adding the dep to the litex hub conda misc are better approaches. @mithro, thoughts?

HackerFoo:
This is a temporary fix until we have a proper package. It doesn't take long to build.

litghost:
> This is a temporary fix until we have a proper package. It doesn't take long to build.

You would have to add this temporary fix to 5 build systems, so even if it doesn't take a long time, it like a bad approach.

HackerFoo:
I think this is a better solution: https://github.com/SymbiFlow/fasm/pull/41

HackerFoo:
Superseded by https://github.com/SymbiFlow/fasm/pull/41

Issue #1882, opened 2020-12-15
Bad links in benchmarks tarball from CI


HackerFoo:
```
$ curl "https://www.googleapis.com/download/storage/v1/b/symbiflow-arch-defs/o/artifacts%2Fprod%2Ffoss-fpga-tools%2Fsymbiflow-arch-defs%2Fcontinuous%2Finstall%2F113%2F20201210-021339%2Fsymbiflow-arch-defs-benchmarks-27c47842.tar.xz?generation=1607610517465322&alt=media" | tar xJ

$ ls -l benchmarks/sdc/
total 10
lrwxrwxrwx 1 dusty users  94 Dec 10 06:07 ibex_arty.sdc -> /tmpfs/src/github/symbiflow-arch-defs-continuous-install/xc/xc7/tests/soc/ibex/pins_artya7.sdc
-rw-r--r-- 1 dusty users 832 Dec 10 03:30 linux_arty_100t.sdc
-rw-r--r-- 1 dusty users 832 Dec 10 02:59 linux_arty.sdc
-rw-r--r-- 1 dusty users 187 Dec 10 02:22 minilitex_arty_a7.sdc
-rw-r--r-- 1 dusty users 187 Dec 10 02:22 minilitex_arty.sdc
-rw-r--r-- 1 dusty users 728 Dec 10 03:29 minilitex_ddr_arty_a7.sdc
-rw-r--r-- 1 dusty users 728 Dec 10 02:57 minilitex_ddr_arty.sdc
-rw-r--r-- 1 dusty users 832 Dec 10 03:29 minilitex_ddr_eth_arty_100t.sdc
-rw-r--r-- 1 dusty users 832 Dec 10 02:58 minilitex_ddr_eth_arty.sdc
-rw-r--r-- 1 dusty users 117 Dec 10 02:57 picosoc_basys3_full_100.sdc
-rw-r--r-- 1 dusty users 116 Dec 10 02:57 picosoc_basys3_full_50.sdc

```

Related PR: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1830

PR #1883, opened 2020-12-15
install: resolve symlinks before packaging

acomodi:

Fixes https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1882

### DIFF
--- a/.github/kokoro/install.sh
+++ b/.github/kokoro/install.sh
@@ -46,6 +46,12 @@ echo "Compressing and uploading install dir"
 echo "----------------------------------------"
 (
 	rm -rf build
+
+	# Remove symbolic links and copy content of the linked files
+	for file in $(find install -type l)
+		do cp --remove-destination $(readlink $file) $file
+	done
+
 	du -ah install
 	export GIT_HASH=$(git rev-parse --short HEAD)
 	tar -I "pixz" -cvf symbiflow-arch-defs-install-${GIT_HASH}.tar.xz -C install bin share/symbiflow/techmaps share/symbiflow/scripts environment.yml


HackerFoo (approved):

common/cmake/devices.cmake
```diff
@@ -1470,7 +1470,7 @@ function(ADD_FPGA_TARGET)
     append_file_dependency(VPR_DEPS ${ADD_FPGA_TARGET_INPUT_SDC_FILE})
     get_file_location(SDC_LOCATION ${ADD_FPGA_TARGET_INPUT_SDC_FILE})
     set(SDC_ARG --sdc_file ${SDC_LOCATION})
-    set(SDC_FILE ${SDC_LOCATION})
```
litghost, line 1473: This will not work generally speaking.  If the SDC is generated, this will break.

acomodi: This is in the if branch where the SDC is manually written. The case in which the SDC is auto-generated, the actual generated SDC is taken instead: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/7183cb5e59d50955686c12e4b800a55fafd1fae5/common/cmake/devices.cmake#L1461-L1467

litghost: Even in that cases, the user can provide a generated SDC, e.g.:
```
add_file_target(FILE my_generated.sdc GENERATED)
add_fpga_target(
 ...
 SDC_FILE my_generated.sdc
 )
```

The new code will break in this case.

acomodi: I have changed approach and now the symlinks are resolved before packaging results into the tarball

PR #1885, opened 2020-12-15
Add bot locking and updating Conda and Pip packages

ajelinski:
This PR adds a bot for creating and updating locks for Conda and Pip packages. It consists of a GitHub Actions workflow which handles `git` and creates a Pull Request and `update_conda_lock.py` script, which creates or updates locks.

Basically, it creates Conda environment based on `environment.yml` (which shouldn't be locked, but locked will work too) and creates locks based on Conda and Pip packages installed.

Currently the workflow added can only be run manually and that's how I tested it. It can be easily modified to work regularly on schedule of course. For now there are some simple constants used for PR title, user creating commit and commit title but these are easily configurable. For testing purposes I set them to be set as an input of the manual workflow.

To to use locks during building the scripts will need to be slightly modified. Instead of using `conda env create --file environment.yml` it needs to be `conda create --file conda.lock -n env` followed by installing pip packages by the pip from such environment, eg. `conda run -n env python -m pip install -r pip.lock`.

The script `update_conda_lock.py` can be tested locally after setting required environment variables:
* `BOT_ENV_NAME` -- custom name for the temporarily created environment (such environment cannot exist),
* `BOT_ENV_YML` -- path of `environment.yml` file, relative to the current directory (in CI it's repository root),
* `BOT_CONDA_LOCK` -- path of the conda lock being updated or created, eg. `conda.lock`,
* `BOT_PIP_LOCK` -- path of the pip lock being updated or created, eg. `pip.lock`.

It doesn't call git or create a PR so it's safe to use locally. Of course its requirement is that `conda` command is available.

### DIFF
--- a/.github/workflows/update_conda_lock.yml
+++ b/.github/workflows/update_conda_lock.yml
@@ -1,0 +1,29 @@
+name: update_conda_lock
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+
+jobs:
+  update-locks:
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules:  'recursive'
+        ref:         'master'    # It will be the PR's base
+
+    - name: Update Lock and Issue a Pull Request
+      uses: SymbiFlow/actions/update_conda_lock@7a34d4c1a27dc369cfc618ae0902fcfaf1a36c59
+      with:
+        branch_name_core: 'update_conda_lock'
+        conda_lock_file:  'conda_lock.yml'
+        environment_file: 'environment.yml'
+        gh_access_token:  ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/update_conda_lock.yml
+++ b/.github/workflows/update_conda_lock.yml
@@ -1,7 +1,6 @@
 name: update_conda_lock
 
 on:
-  pull_request:
   push:
   schedule:
     - cron: '0 3 * * *'
@@ -21,6 +20,7 @@ jobs:
         ref:         'master'    # It will be the PR's base
 
     - name: Update Lock and Issue a Pull Request
+      if: ${{ github.ref == 'refs/heads/master' }}
       uses: SymbiFlow/actions/update_conda_lock@7a34d4c1a27dc369cfc618ae0902fcfaf1a36c59
       with:
         branch_name_core: 'update_conda_lock'
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Makefile
 TOP_DIR := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
 REQUIREMENTS_FILE := requirements.txt
-ENVIRONMENT_FILE := environment.yml
+ENVIRONMENT_FILE := conda_lock.yml
 
 third_party/make-env/conda.mk:
 	git submodule init
@@ -23,6 +23,9 @@ all: env
 clean::
 	rm -rf build
 
+# Conda Lock contains all dependencies. PIP often fails if a package is both
+# specified to be installed directly and as a dependency of another package.
+env:: export PIP_NO_DEPS = true
 env:: | $(CONDA_ENV_PYTHON)
 	git submodule init
 	git submodule update --init --recursive
--- a/conda_lock.yml
+++ b/conda_lock.yml
@@ -1,0 +1,180 @@
+name: symbiflow_arch_def_base
+channels:
+  - litex-hub
+  - defaults
+dependencies:
+  - _libgcc_mutex=0.1=main
+  - _openmp_mutex=4.5=1_gnu
+  - binutils-riscv64-elf=2.34=20210513_184432
+  - bison=3.7.5=h2531618_1
+  - bzip2=1.0.8=h7b6447c_0
+  - c-ares=1.17.1=h27cfd23_0
+  - ca-certificates=2021.7.5=h06a4308_1
+  - capnproto=0.8.0=20210316_201220
+  - capnproto-java=0.1.5_0012_g44a8c1e=20201104_165332
+  - certifi=2021.5.30=py37h06a4308_0
+  - cloog=0.18.0=0
+  - cmake=3.19.6=h973ab73_0
+  - cython=0.29.24=py37h295c915_0
+  - expat=2.4.1=h2531618_2
+  - flake8=3.9.2=pyhd3eb1b0_0
+  - flex=2.6.4=ha10e3a4_1
+  - gcc-riscv64-elf-newlib=9.2.0=20201119_154229
+  - gcc-riscv64-elf-nostdc=9.2.0=20210513_184432
+  - gmp=6.2.1=h2531618_2
+  - icestorm=0.0_0719_g792cef0=20201120_145821
+  - icu=58.2=he6710b0_3
+  - importlib-metadata=3.10.0=py37h06a4308_0
+  - isl=0.21=20210513_184432
+  - iverilog=s20150603_0957_gad862020=20201120_145821
+  - krb5=1.19.2=hac12032_0
+  - ld_impl_linux-64=2.35.1=h7274673_9
+  - libcurl=7.78.0=h0b77cf5_0
+  - libedit=3.1.20210216=h27cfd23_1
+  - libev=4.33=h7b6447c_0
+  - libffi=3.3=he6710b0_2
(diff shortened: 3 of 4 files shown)


.github/update_conda_lock.py
```diff
@@ -0,0 +1,123 @@
...
+                return False
+    except FileNotFoundError:
+        print(path + ' doesn\'t exist; it will be created.')
+    with open(path, 'w') as f:
+        f.write(new_lock)
+    print()
+    return True
+
+updated_files = []
+if try_updating_lock_file(conda_lock_path, conda_lock):
```
litghost, line 110: Please put top level logic into a `main` or otherwise named function.

ajelinski: Done.

.github/update_conda_lock.py
```diff
@@ -0,0 +1,130 @@
...
+    return env_var
+
+def _remove_conda_env(conda_env_name):
+    print('Removing `' + conda_env_name + '` Conda environment... ', end='')
+    _run('conda env remove -n ' + conda_env_name, stdout=DEVNULL,
+            stderr=DEVNULL)
+    print('done!')
+    print()
+
+# Returns True if lock file has been updated, False otherwise
```
litghost, line 33: Use docstrings over comments before the method.

.github/update_conda_lock.py
```diff
@@ -0,0 +1,130 @@
...
+    if return_stdout:
+        return run(cmd, check=True, encoding='utf-8', stdout=PIPE,
+                **kwargs).stdout
+    else:
+        return run(cmd, check=True, **kwargs)
+
+def _get_env(env_name, required=True):
+    if env_name not in environ:
+        if required:
+            print('ERROR: Required environment variable not found: ' + env_name + '!')
```
litghost, line 20: Why does this print, versus throw an exception?  Seems like an error is an error?

.github/update_conda_lock.py
```diff
@@ -0,0 +1,130 @@
...
+        return None
+    env_var = environ[env_name]
+    print('* ' + env_name + ': ' + env_var)
```
litghost, line 23: Should this print be behind a verbose flag?

.github/update_conda_lock.py
```diff
@@ -0,0 +1,130 @@
...
+            if conda_pkg_match is None:
+                pip_lock += pip_spec + '\n'
+            else:
+                print('Conda package removed from pip.lock: ' + conda_pkg_match.group(1))
+        print()
+        print('Pip packages captured.')
+        print()
+
+    # Conda environment isn't needed anymore
+    _remove_conda_env(conda_env)
```
litghost, line 112: If we care about cleaning up the conda env, we should likely use a context, rather than flow (see line 89 for a flow example).  So something like:

```
with CondaEnv('path to setup conda env') as env:
  <blah>
```

.github/update_conda_lock.py
```diff
@@ -0,0 +1,130 @@
...
+    # Capture frozen list of pip packages
+
+    # Variable will stay None in case there's no pip in environment.yml
+    pip_lock = None
+
+    # Check pip version
+    pip_version_match = search(r'pip\-([0-9]+)\.([0-9]+)[^\-]+-[^\-]+\.',
+            conda_lock)
+    if pip_version_match is not None:
+        if pip_lock_path is None:
```
litghost, line 85: Can you explain (and add the revelant comments) around how `BOT_PIP_LOCK` is intended to be used?

.github/update_conda_lock.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
```
litghost, line 1: This file needs a nice block comment.  I believe your PR comment has some of the required information, but it should be in the python file, not just the github PR.

.github/update_conda_lock.py
litghost: One way to do this easily is use: [temp directory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), which supports context.

PiotrZierhoffer:
@mithro @litghost @HackerFoo there's a question we need to address and would appreciate your opinion.

The bot is a github action that creates a PR. Unfortunately, such PRs, when created with the standard `GITHUB_TOKEN`, don't trigger further GitHub actions - so we don't know if our update works or not.

This is a pain, but there are workarounds described here: https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#triggering-further-workflow-runs

TL;DR:
* manually close and reopen a PR to trigger CI. Very easy, but manual
* create a personal token and push using it instead of `GITHUB_TOKEN`. There is a symbiflow-bot account that could be used for that, but it would require someone with proper accesses to set things up (who would that be? @mithro ?). **I'd recommend this option**.
* Create a dedicated GitHub App - an overkill imho
* Use deploy SSH keys - would work only on push events, not pull requests, so it's not really good

mithro:
@PiotrZierhoffer - I think a dedicated GitHub App is the long term solution (the goal is that this bot should be eventually usable by anyone who wants to roll conda-lock files).

In the short term I can set up the personal token. What do you want the environment variable to be called (so it is different from the repository provided `GITHUB_TOKEN`)?

ajelinski:
To be honest the name doesn't really matter as it will be passed to the bot as a GitHub Action `with: ` variable so every repository using that Action in a workflow can pass it's own variable.

If this is going to be the `symbiflow-bot`'s token then maybe `SYMBIFLOW_TOKEN` is a reasonable name?

.github/update_conda_lock.py
ajelinski: I didn't want to exit in this precise moment to print all required variables that are missing.

.github/update_conda_lock.py
ajelinski: I'd rather have those seen by default in the workflow's log so I don't think this should be behind such flag.

.github/update_conda_lock.py
ajelinski: This is done: https://github.com/SymbiFlow/actions/blob/4c154b89bb203d0d67620ce72d3c4f2553cba517/update_conda_lock/update_lock.py#L187

.github/update_conda_lock.py
ajelinski: It was intended to be used after creating Conda environment. I found out I can merge `pip freeze` output into the locked `environment.yml` file so this file doesn't exist anymore.

ajelinski:
There were significant changes after thorough testing with `symbiflow-arch-defs` repository. Previously it's been tested during development on `symbiflow-examples` which eg. lacks local pip packages which caused many problems.

The Action has been separated from the workflow and made much easier to use on any repository. The Action itself is now the subject of this PR: https://github.com/SymbiFlow/actions/pull/5

The workflow `update_conda_lock` has failed because of permission problems: https://github.com/SymbiFlow/symbiflow-arch-defs/runs/1628574436?check_suite_focus=true#step:3:858 and it seems to be a cross-repository PR problem. Running this workflow after merge the same token should work. Definitely after changing token to the PAT it should work too but also should trigger the CI on the created PR.

Pushing and creating a PR from this PR's latest commit has been successful on the Antmicro's fork: https://github.com/antmicro/symbiflow-arch-defs/runs/1628574276?check_suite_focus=true#step:3:856 and the PR adding a lock to the master branch from that workflow is: https://github.com/antmicro/symbiflow-arch-defs/pull/23

ajelinski:
It seems `update_conda_lock` workflow failed because of the cross-repository restrictions of the GH Actions tokens.
The `git push` result was:
>   remote: Permission to SymbiFlow/symbiflow-arch-defs.git denied to github-actions[bot].

It works well in a fork, though: https://github.com/antmicro/symbiflow-arch-defs/runs/3374983149?check_suite_focus=true#step:3:991

mithro:
BTW Are we using https://github.com/conda-incubator/conda-lock ?

ajelinski:
> BTW Are we using https://github.com/conda-incubator/conda-lock ?

No, not at all. I've tested it and I don't even remember all of the issues but certainly it didn't include PIP packages at all.
Conda's built-in `conda env export` is used to capture Conda packages instead.

mithro:
Can we figure out how to collaborate with the conda-lock project?

PiotrZierhoffer:
@mithro sure, it's probably quite specific, but we can think about that

acomodi:
All of the kokoro CI went green in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1885/commits/79ff9e9c0fa75aba889f9a919dafeaca4e33b8f7 and https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1885/commits/13c4e835bc1874733d214ca5c67d6456744895b3 only changes the conda lock CI to not be triggered on PRs and to be executed only on the default branch. Merging

PR #1886, opened 2020-12-15
Lower timeout to workaround Vivado 2019.2 hang.

litghost:
This is ~3x current maximum runtime.


litghost:
Looks like a RTD network failure is the reason for the RTD build failure?

mithro (approved):

PR #1892, opened 2020-12-16
Update environment.yml to use vtr-optimized (vtr compiled with PGO)

HackerFoo:


environment.yml
```diff
@@ -11,7 +11,7 @@ dependencies:
   - litex-hub::symbiflow-yosys-plugins=1.0.0_7_g59ff1e6_23_g3a95697_17_g00b887b_0194_g40efa51=20201120_145821
   - litex-hub::prjxray-tools=0.1_2697_g0f939808=20201120_145821
   - litex-hub::prjxray-db=0.0_0239_gd87c844=20201120_145821
-  - litex-hub::vtr=v8.0.0_3015_gd912bdb88=20201210_100534
+  - litex-hub::vtr-optimized-v8.0.0_3049_g64d15e2dd=20201216_065815
```
acomodi, line 14:
```suggestion
  - litex-hub::vtr-optimized=v8.0.0_3049_g64d15e2dd=20201216_065815
```

acomodi (changes requested):

environment.yml
HackerFoo: Thanks.

HackerFoo:
@mithro Could you help me figure out why Conda fails? `make env` works on my workstation.

HackerFoo:
My theory is that the condarc file in the package causes this, because `vtr-gui` also fails: https://github.com/HackerFoo/symbiflow-arch-defs/tree/test-ci

HackerFoo:
I think I know what's happening. Some dependencies are coming from `conda-forge`, so that needs to be included in `environment.yml`.

acomodi:
The fact that there are some conda-forge dependencies is a hint that there is something wrong with the vtr-optmized package. Looking at the package description in litex-hub-conda, I see that there is a condarc file which should be deleted from there: https://github.com/litex-hub/litex-conda-eda/blob/master/pnr/vtr-optimized/condarc.

The problem is that conda-forge generates this kinds of issues and leads to instability, and we should not use it at all.

HackerFoo:
Unfortunately, there is no other good way to get `python-constraints`. I can't use pip: https://github.com/conda/conda-build/issues/548

HackerFoo:
I have created a package for `python-constraint` to avoid depending on `conda-forge`.

environment.yml
```diff
@@ -11,7 +11,7 @@ dependencies:
   - litex-hub::symbiflow-yosys-plugins=1.0.0_7_g59ff1e6_23_g3a95697_17_g00b887b_0194_g40efa51=20201120_145821
   - litex-hub::prjxray-tools=0.1_2697_g0f939808=20201120_145821
   - litex-hub::prjxray-db=0.0_0239_gd87c844=20201120_145821
-  - litex-hub::vtr=v8.0.0_3015_gd912bdb88=20201210_100534
+  - litex-hub::vtr-optimized=v8.0.0_3049_g64d15e2dd=20201216_212525
```
acomodi, line 14:
```suggestion
  - litex-hub::vtr-optimized=v8.0.0_3049_g64d15e2dd=20201217_025303
```

AFAIK, only the packages labeled as `main` are available. Using a non-main labeled package results in conda not being able to find it.

acomodi (changes requested):

environment.yml
HackerFoo: Fixed. I was still waiting on the `main` package, so I thought I'd try it.

acomodi (approved):
LGTM, let's wait for CI

litghost (approved):
LGTM, merge if green

litghost:
I'm not seeing an improvement in overall runtime.  This is likely because most of the time is spent during the lookahead computation.  @HackerFoo have you explored/tested PGO improvements on the lookahead computation stuff?

HackerFoo:
I have not. It might be worth trying a separate PGO build optimized for lookup computation, but it's probably not worth the ~10% performance gain, and I suspect that lookup is mostly memory bound, so it will probably have less effect.

I think the most effective way to improve lookup time is by balancing the parallel workload better, and shortening the tail.

PR #1895, opened 2020-12-17
Update third_party/fasm

HackerFoo:
Use latest fasm parser to benefit from increased performance.


litghost (approved):
LGTM, merge if green and if the FASM warning about the fallback doesn't print.

PR #1906, opened 2020-12-21
xc7: yosys: allow custom top module instead of using only auto-top

acomodi:


tmichalak (approved):
LGTM

PR #1907, opened 2020-12-21
toolchain: yosys: add possibility to assign frontend

acomodi:

This PR adds the possibility to instruct yosys on which kind of frontend to use to parse input files.

https://github.com/SymbiFlow/symbiflow-examples/issues/113 shows that there are some riscv CPU implementation that have non standard verilog files for which the frontend type cannot be guessed by yosys (SERV core).

This PR assigns a default frontend (i.e. verilog) that can be overridden with the `-f|-frontend` flag in the toolchain wrappers.

### DIFF
--- a/xc/xc7/toolchain_wrappers/symbiflow_synth
+++ b/xc/xc7/toolchain_wrappers/symbiflow_synth
@@ -18,12 +18,14 @@ XDC_FILES=()
 TOP=top
 DEVICE="*"
 PART=""
+FRONTEND=verilog
 
 VERILOGLIST=0
 XDCLIST=0
 TOPNAME=0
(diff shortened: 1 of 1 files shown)


mithro (commented):
I'm unsure if this is the correct fix here? Shouldn't the type be per file, not a global?

acomodi:
I am unsure as well here actually. I don't think there is a way to specify the frontend for each single file in the list of input files.

Another solution would be to add `.vh` (verilog header) files to the list in [yosys directly](https://github.com/YosysHQ/yosys/blob/d15c63effc49e6227e99412afa8a78afb48de0e1/kernel/yosys.cc#L960-L986).

PR #1933, opened 2021-01-06
WIP: Add support for new Vivado 2020.2 image.

litghost:


litghost:
I expected this to resolve the hanging Vivado issue we've been seeing, but Vivado 2020.2 still appears to have the issue.  This likely points to something else is going on.

PR #1947, opened 2021-01-14
Handle different location of fabric data

dnltz:
Handles the new location of the fabric data inside the database. A new tool is needed to get the fabric for a part number.

This PR needs the latest version of prjxray and prjxray-db. I'm not sure how the process of upgrading these packages is.

### DIFF
--- a/xc/common/cmake/project_ray.cmake
+++ b/xc/common/cmake/project_ray.cmake
@@ -36,7 +36,17 @@ function(PROJECT_RAY_ARCH)
   set(ARCH_IMPORT ${symbiflow-arch-defs_SOURCE_DIR}/xc/common/utils/prjxray_arch_import.py)
   set(CREATE_SYNTH_TILES ${symbiflow-arch-defs_SOURCE_DIR}/xc/common/utils/prjxray_create_synth_tiles.py)
   set(CREATE_EDGES ${symbiflow-arch-defs_SOURCE_DIR}/xc/common/utils/prjxray_create_edges.py)
-  set(DEPS ${PRJRAY_DB_DIR}/${PRJRAY_ARCH}/${PART}/tilegrid.json)
+  set(GET_FABRIC ${symbiflow-arch-defs_SOURCE_DIR}/xc/common/utils/prjxray_get_fabric.py)
+  execute_process(
+    COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${PRJRAY_DIR}:${symbiflow-arch-defs_SOURCE_DIR}/utils
+    ${PYTHON3} ${GET_FABRIC}
+      --db_root ${PRJRAY_DB_DIR}/${PRJRAY_ARCH}/
+      --part ${PART}
+      -cmake
+    OUTPUT_VARIABLE FABRIC
+    )
+  string(REPLACE "\n" "" FABRIC "${FABRIC}")
+  set(DEPS ${PRJRAY_DB_DIR}/${PRJRAY_ARCH}/${FABRIC}/tilegrid.json)
 
   if("${PROJECT_RAY_ARCH_PB_TYPES}" STREQUAL "")
     set(PROJECT_RAY_ARCH_PB_TYPES ${PROJECT_RAY_ARCH_TILE_TYPES})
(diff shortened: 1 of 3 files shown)


xc/common/cmake/project_ray.cmake
```diff
@@ -36,7 +36,17 @@ function(PROJECT_RAY_ARCH)
...
+  set(GET_FABRIC ${symbiflow-arch-defs_SOURCE_DIR}/xc/common/utils/prjxray_get_fabric.py)
+  execute_process(
+    COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${PRJRAY_DIR}:${symbiflow-arch-defs_SOURCE_DIR}/utils
+    ${PYTHON3} ${GET_FABRIC}
+      --db_root ${PRJRAY_DB_DIR}/${PRJRAY_ARCH}/
+      --part ${PART}
+      -cmake
+    OUTPUT_VARIABLE FABRIC
+    )
+  string(REPLACE "\n" "" FABRIC "${FABRIC}")
```
litghost, line 48: I think you don't need the `string(REPLACE` anymore, because you added the `end=""`?

litghost:
```
SymbiFlow Arch Defs: - Check code formatting
  -------------------------------------------------------------------
  Scanning dependencies of target check_python
  yapf needs to reformat /home/runner/work/symbiflow-arch-defs/symbiflow-arch-defs/xc/common/utils/prjxray_get_fabric.py
  make[3]: *** [CMakeFiles/check_python.dir/build.make:76: CMakeFiles/check_python] Error 1
  make[3]: Target 'CMakeFiles/check_python.dir/build' not remade because of errors.
  make[2]: *** [CMakeFiles/Makefile2:4655: CMakeFiles/check_python.dir/all] Error 2
  make[1]: *** [CMakeFiles/Makefile2:4662: CMakeFiles/check_python.dir/rule] Error 2
  make[1]: Target 'check_python' not remade because of errors.
  make: *** [Makefile:395: check_python] Error 2
```

xc/common/cmake/project_ray.cmake
dnltz: Well, the bash prints a newline at the beginning of the output. That's why I need this replace. I could remove the end="" in the print because that isn't necessary now.

litghost: Ya, remove the `end=""` in that case.

litghost:
@dnltz You need to update to a newer prjxray-db package here: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/bff52005b478381125b235c66f394f0f993db287/environment.yml#L13

See newer versions here (hopefully): https://anaconda.org/LiteX-Hub/prjxray-db/files

dnltz:
@acomodi - Do you know who maintains the Anaconda prjxray-db package? Would be great to have a release with v0.0_243_g057e179 because of the most recent changes in the database.

mithro:
FYI - @PiotrZierhoffer

mithro:
See also https://github.com/litex-hub/litex-conda-eda/issues/69

litghost (changes requested):
With the manual labelling of prjxray-db, @dnltz you should be able to update this PR, yes?

dnltz:
Need to find out how to update to the other labels. See: https://github.com/litex-hub/litex-conda-eda/issues/69
But will send new patches with the new package versions than, yes.

litghost:
This appears to be running now, we'll see what happens when the CI finishes.  The 7-series CI has already failed for unrelated reasons, we'll see what the other CIs do.

dnltz:
I found this:
```
Dumping database to '/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/xc/xc7/archs/artix7/devices/xc7a50t-basys3-roi-virt/channels.db'
No progressbar disabled because non-interactive terminal.
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc/common/utils/prjxray_create_edges.py", line 103, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc/common/utils/prjxray_create_edges.py", line 72, in main
    ccio_sites = create_edges(args)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc/common/utils/prjxray_edge_library.py", line 2312, in create_edges
    mark_track_liveness(conn, input_only_nodes, output_only_nodes)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/xc/common/utils/prjxray_db_cache.py", line 61, in __exit__
    self.file_connection, pages=100, progress=self._progress
sqlite3.OperationalError: database or disk is full
```
Maybe not enough memory?

litghost:
> Maybe not enough memory?

There is a known bug with kokoro where it sometimes doesn't mount/attach the temp disk that builds happen on.  I'll re-run, but I think your changes will pass.

dnltz:
@litghost - as you said

PR #1951, opened 2021-01-15
WIP: Remove gcc-riscv5, causing lots of conflicts?

litghost:


acomodi:
Originally, the gcc-riscv was required to build the LiteX BIOS. Removing it causes the LiteX tests to fail. 

We could find some other gcc-riscv5 package and install it in the host directly instead of relying on the conda-package.

The question is, what has changed to cause these conda-conficts? All the litex-hub packages are pinned, so is it possible that a new version of miniconda is causing this?

litghost:
@acomodi I didn't expect this to work, but wanted to confirm that it cleared the conda error.  I agree that a newer version of miniconda could be the problem, I suspect the upstream c++ `.so` is generating the conflicts.  It is worth noting that litex-hub package generation is also broken, see https://github.com/issues

PR #1972, opened 2021-01-25
Fix build error with picosoc example

### LINKED ISSUE #1967: Building picosoc test firmware fails
I just wanted to try to modify the firmware for the picosoc demo in symbiflow-examples. I found the firmware for it in this symbiflow-arch-defs repository.

I installed the toolchain with:
`conda install -c litex-hub gcc-riscv32-elf-newlib` 

But when building I get this error:
```
riscv32-unknown-elf-gcc -DHX8KDEMO -DCLK_FREQ=25000000 -DNOSPIFLASH -march=rv32im -Wl,-Bstatic,-T,sections.lds,--strip-debug -ffreestanding -nostdlib -mstrict-align -O0 -o firmware_noflash_25.elf start.s firmware.c
cc1: error: requested ABI requires '-march' to subsume the 'D' extension
make: *** [Makefile:22: firmware_noflash_25.elf] Error 1
```

There is some details about the compiler args on https://www.sifive.com/blog/all-aboard-part-1-compiler-args

I found that I could add `-mabi=ilp32` and then it would build. I am not sure if this is the correct fix, but it appears to be working when I synth the verilog and put it on the basys3. At least it includes my new "printf" and it runs as before.

```
./hex2progmem.py --rom-style initial firmware/firmware_noflash_100.hex  > ~/symbiflow-examples/xc7/picosoc_demo/progmem.v
.....
make
openocd ....
```

Version info about compiler for completeness:
```
riscv32-unknown-elf-gcc --version
riscv32-unknown-elf-gcc (9.2.0) 9.2.0
```

nickoe:
Add mabi option to make the picosoc example build

Fixes #1967

### DIFF
--- a/tests/9-soc/picosoc/firmware/Makefile
+++ b/tests/9-soc/picosoc/firmware/Makefile
@@ -2,7 +2,7 @@ CC = ${RISCV}/bin/riscv32-unknown-elf-gcc
 OC = ${RISCV}/bin/riscv32-unknown-elf-objcopy
 OD = ${RISCV}/bin/riscv32-unknown-elf-objdump
 
-CFLAGS = -march=rv32im -Wl,-Bstatic,-T,sections.lds,--strip-debug -ffreestanding -nostdlib -mstrict-align -O0  
+CFLAGS = -march=rv32im -mabi=ilp32 -Wl,-Bstatic,-T,sections.lds,--strip-debug -ffreestanding -nostdlib -mstrict-align -O0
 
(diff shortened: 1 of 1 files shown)


kgugala:
The change looks good. @nickoe can you sign-off your commit, so the DCO is happy

mithro:
@kgugala -- We should check if something like this is needed in LiteX in the future too?

nickoe:
@kgugala updated with signoff.

litghost:
> @kgugala updated with signoff.

Thanks!

PR #1984, opened 2021-01-28
Add GTP primitives to artix7 architectures

kowalewskijan:
This is a PR which adds GTP primitives to the SymbiFlow toolchain:

- [x] GTPE2_COMMON
- [x] GTPE2_CHANNEL

For artix7 achitectures:

- [x] xc7a50t
- [ ] xc7a100t
- [x] xc7a200t


xc/common/primitives/ipad/ipad.model.xml
```diff
@@ -0,0 +1,7 @@
+<models>
+  <model name="IPAD_VPR">
```
mkurc-ant, line 2: This will require `never_prune="true"` so that VPR does not remove this. From VPR perspective its a leaf cell with inputs only.

xc/common/primitives/opad/opad.model.xml
```diff
@@ -0,0 +1,7 @@
+<models>
+  <model name="OPAD_VPR">
```
mkurc-ant, line 2: Probably `never_prune` would be applicable here too.

acomodi:
Closing this PR and opening a new one as Kokoro is not triggered with every push.

PR #1987, opened 2021-01-28
Add RapidYaml git URL to requirements.

litghost:


mithro:
Should this be pulled in through a dependencies in some way?

litghost:
> Should this be pulled in through a dependencies in some way?

It's not on PyPi, so no.

mithro:
BTW What depends on yaml?

litghost:
> BTW What depends on yaml?

FPGA interchange

PR #2003, opened 2021-02-03
LiteX submodules update

kowalewskijan:
This PR bumps LiteX submodules and includes `generate.py` fixups to work with the latest LiteX.
PR can be merged after HW tests using generated bitstreams from litex examples are proven to work

### DIFF
--- a/xc/xc7/tests/soc/litex/generate.py
+++ b/xc/xc7/tests/soc/litex/generate.py
@@ -3,10 +3,10 @@
 import os
 import argparse
 
-from litex.boards.targets.arty import _CRG
+from litex_boards.targets.arty import _CRG
 from liteeth.phy.mii import LiteEthPHYMII
 from litex.soc.cores.led import LedChaser
-from litex.boards.platforms import arty
+from litex_boards.platforms import arty
 from litex.soc.integration.soc_core import SoCCore
 from litex.soc.integration.builder import Builder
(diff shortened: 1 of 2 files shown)


xc/xc7/tests/soc/litex/generate.py
```diff
@@ -20,7 +20,7 @@ class BaseSoC(SoCCore):
     def __init__(
             self,
             toolchain="vivado",
-            sys_clk_freq=int(60e6),
+            sys_clk_freq=int(100e6),
```
litghost, line 23: It probably isn't a good idea to increase the sys clock freq.  I don't believe that VPR can achieve 100 MHz.  @acomodi ?

acomodi: With latest LiteX we have a minimum of 80MHz, below which DDR does not work even in a full Vivado flow. I think that we can set it to 80MHz.

80MHz worked on HW with symbiflow, but might not work on every single arty.

litghost (commented):
LGTM, @acomodi please review.  If green merge.

acomodi (approved):
LGTM. I'd wait for CI and check the bitstream produced on HW, to verify that the frequency increase does work.

umarcor:
@kowalewskijan can you please rebase to solve the conflicts?

PR #2004, opened 2021-02-03
Add litesata example for nexys_video board

kowalewskijan:
This PR adds:
* support for nexys_video in litex tests
* litesata example for nexys video board

Depends on:
* https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2003
* https://github.com/litex-hub/litex-boards/pull/171


xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -786,7 +786,7 @@ def main():
 
     idelayctrl_instances = place_constraints.get_used_instances("IDELAYCTRL")
 
-    assert len(idelayctrl_cmts) == len(
+    assert len(idelayctrl_cmts) <= len(
         idelayctrl_instances
```
litghost, line 790: This will not operate properly, the previous assertion is correct.  In order to relax this contraint, the synthesis tool needs to emit duplicate IDELAYCTRL and tie their outputs together with a `AND<n>` LUT.  See TODO/FIXME above or below this code.

Issue #1991, opened 2021-01-29
PLLE2_BASE techmap doesn't match yosys' cells


Xiretza:
yosys [`cells_xtra.v`](https://raw.githubusercontent.com/YosysHQ/yosys/master/techlibs/xilinx/cells_xtra.v) contains:

```verilog
    output CLKFBOUT;
    output CLKOUT0;
    output CLKOUT1;
    output CLKOUT2;
    output CLKOUT3;
    output CLKOUT4;
    output CLKOUT5;
    output LOCKED;
    input CLKFBIN;
    input CLKIN1;
    input PWRDWN;
    input RST;
```

while [`cells_map.v`](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc/xc7/techmap/cells_map.v#L4295) contains:

```verilog
input         CLKFBIN,
input         CLKIN,

output        CLKFBOUT,
output        CLKOUT0,
output        CLKOUT1,
output        CLKOUT2,
output        CLKOUT3,
output        CLKOUT4,
output        CLKOUT5,

input         RST,
output        LOCKED
```

Compared to yosys, `CLKIN` should be `CLKIN1`, and the `PWRDWN` pin is missing - fixing these two makes techmap work. This is the only module I had problems with, I don't know if something similar applies to other modules.

litghost:
Per https://github.com/Xilinx/XilinxUnisimLibrary/blob/1c8e05fd1e9a79ceb8b996a0996674122eed086f/verilog/src/unisims/PLLE2_BASE.v#L37-L48 , I agree that the techmap in https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc/xc7/techmap/cells_map.v#L4295 is wrong.

litghost:
@mkurc-ant Can you fix this when you bring in the MMCM support?

mkurc-ant:
@litghost Sure, I'll take a look into it.

PR #2007, opened 2021-02-04
[WIP] Fixed support for PLLE2_BASE

mkurc-ant:
Attempt to solve https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1991. Expecting CI to fail for now due to `BUFHCE` handling in fasm2bels.


mkurc-ant:
Vendor tool tests (Vivado) fails on the PLL test with:
```
ERROR: [DRC PLCR-1] Placement Constraints Check for Clock Region(s): Design Check found an error in Clock Region X0Y0. This clock region has 14 clocks locked whereas only 12 clocks can be routed per clock region. The cells were either constrained through Area Groups or locked via user constraints. This situation can be resolved by unlocking some cells from the user constraints so that there are only 12 clocks in this region.
```
This is due to missing correct `BUFHCE` route-through handling in fasm2bels.

I'll try to limit the number of output clocks from the PLL in the test design. This should mitigate the problem.

mkurc-ant:
:/ Looks like there are some conflicting bits in the `prjxray-db` related to routing around `PLL`:
```
prjxray.fasm_assembler.FasmInconsistentBits: FASM line "CMT_TOP_R_UPPER_T_X8Y44.CMT_TOP_R_UPPER_T_PLLE2_CLKIN1.CMT_TOP_R_UPPER_T_CLKIN1" wanted to clear bit (4194461, 74, 0) but was set by FASM line "CMT_TOP_R_UPPER_T_X8Y44.PLL_CLK_FREQ_BB0_NS_ACTIVE"
```

acomodi:
@mkurc-ant This should have been fixed with https://github.com/SymbiFlow/prjxray/pull/1692, I guess that the prjxray-db conda package needs a bump.

mkurc-ant:
@acomodi Thanks, I've just done that. I was under impression that we used the most recent one on master.

mkurc-ant:
Hmm, after rebasing which removed the fixed version of prjxray-db package the bit conflict came back again.

acomodi:
Right, the fact is that now, `conda_lock.yml` is used to build the environment, and this is generated by the conda bot updater. The `environment.yml` file is used as a guide for the conda bot updater to know which packages to check. I believe that for the time being you can grab the correct package from here https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2251/files# and update the conda-lock file manually.

mkurc-ant:
For some reason Vivado fails the `RTRES-2` DRC check in kokoro even though the check is explicitly disabled in the runme script. This happens for the `picosoc` test on `xc7a200t`.

dnltz:
Hey @mkurc-ant, any update on this PR? Do you need help?

PR #2012, opened 2021-02-05
Updated dependencies to include swig

Arvind-Srinivasan:
Due to the fact rapidyaml now has to be build locally, the build fails without the swig package installed. I added this as a dependency for now to make it easier to have the install process work automatically without having to manually install this package.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,6 +19,9 @@ dependencies:
   - flake8
   - nodejs
   - libiconv
+  # swig is required to build the rapidyaml package
+  # It can be removed once that is available in PyPi
+  - swig
   # libxml2 2.9.10 contains an integer overflow fix required for arch-defs.
   - libxml2>=2.9.10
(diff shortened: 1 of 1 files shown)


litghost:
> Due to the fact rapidyaml now has to be build locally, the build fails without the swig package installed. I added this as a dependency for now to make it easier to have the install process work automatically without having to manually install this package.

Please DCO your commit. Let's see if the CI is happy with the change.

Arvind-Srinivasan:
What exactly is DCO?

litghost:
> What exactly is DCO?

https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin

When you commit, use `git commit -s`.  You can also add signout on an existing branch with `git rebase --signoff ...`.

Arvind-Srinivasan:
Fixed!

litghost:
Looks like CI didn't explode: https://github.com/SymbiFlow/symbiflow-arch-defs/runs/1840294925?check_suite_focus=true

If this goes green I'll merge it!

PR #2036, opened 2021-02-17
Support for QuickLogic k4_N8 OpenFPGA-based device

mkurc-ant:
This PR adds support for `k4_N8` device based on OpenFPGA architecture.

Changes to SymbiFlow include:
- Changes/improvements to a few common CMake scripts,
- New QuickLogic specific CMake scripts for arch and device definition,
- VPR arch.xml file for the new device (temporarly in SymbiFlow, will be moved to the SOFA submodule repo),
- Python scripts required for adapting the OpenFPGA VPR arch XML so that it can be used by upstream VPR,
- Yosys synthesis scripts for the new device,
- A test design suite for the new device

The majority of this PR is contained within the test design suite.

Currently `k4_N8` device requires Yosys from the QuickLogic fork: https://github.com/QuickLogic-Corp/yosys/tree/quicklogic-rebased

Upstream VPR can be used but it requires the patch from: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1656


mkurc-ant:
@lnsharma @tpagarani FYI

litghost (changes requested):
Please move the tests out of the repo.  I don't believe we want to add 30k LOC of test data.

litghost:
I don't believe you have wired in the QL arch to QL CI target? https://source.cloud.google.com/results/invocations/e896f916-e831-493b-bee3-e6bdd9373bcb/log still looks like the empty script?

quicklogic/README.md
```diff
@@ -0,0 +1,199 @@
+# SymbiFlow for Quicklogic FPGAs
```
mithro, line 1:
```suggestion
# SymbiFlow for QuickLogic FPGAs
```

quicklogic/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Quickstart guide
+
+### 1. Install Yosys with QuickLogic support
```
mithro, line 5: This should probably need to rework this. 

The Yosys should be installed from conda like we do with the other tools.

quicklogic/common/utils/flatten_layout.py
```diff
@@ -0,0 +1,515 @@
+#!/usr/bin/env python3
```
mithro, line 1: License header?

quicklogic/dummy.eblif
```diff
@@ -0,0 +1,4 @@
+.model top
```
mithro, line 1: This is probably in the wrong location.

quicklogic/openfpga/devices/k4_N8/k4_N8_tileable_reset_softadder_register_scan_chain_nonLR_caravel_io_skywater130nm.xml
```diff
@@ -0,0 +1,786 @@
+<!--
```
mithro, line 1: It is unclear to me if this XML file should be directly landed here....

quicklogic/openfpga/tests/quicklogic_testsuite/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+add_custom_target(all_openfpga_k4_N8_tests_no_adder)
```
mithro, line 1: These should all be under `third_party`.

quicklogic/openfpga/tests/quicklogic_testsuite/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+add_custom_target(all_openfpga_k4_N8_tests_adder)
+
```
mithro, line 3: A bunch of these tests should be **common** tests which are run on all architectures and under https://github.com/SymbiFlow/symbiflow-arch-defs/tree/master/tests directory.

mithro:
@mkurc-ant - Any idea why the license checking scripts are not firing here?

mithro (changes requested):
There needs to be a bit of restructure before we can move forward here....

mkurc-ant:
@mithro @litghost I've presented QuickLogic people with the issue with the test suite and asked them how they would like to proceed with that.

@mithro I don't believe that we have added license checking to the symbiflow-arch-defs CI yet. Nor we did add license headers to any files.

quicklogic/dummy.eblif
mkurc-ant: It wasn't actually needed. I removed it.

quicklogic/openfpga/devices/k4_N8/k4_N8_tileable_reset_softadder_register_scan_chain_nonLR_caravel_io_skywater130nm.xml
mkurc-ant: At least for now QuickLogic wants the VPR arch file to be part of the SymbiFlow repository.

quicklogic/openfpga/tests/quicklogic_testsuite/CMakeLists.txt
mkurc-ant: After presenting this to the QuickLogic people I removed the test suite. It is to be added as a submodule here or to the symbiflow-examples repo.

quicklogic/openfpga/tests/quicklogic_testsuite/CMakeLists.txt
mkurc-ant: I added the board with the QuickLogic device to the relevant common tests.

mkurc-ant:
I've integrated the CI for QuickLogic. It currently fails due to lack of QuickLogic support in the currently used Yosys version.

mkurc-ant:
Also for "common" tests to pass for QuickLogic a VPR version is required with the fix: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1656

quicklogic/README.md
mkurc-ant: I agree. But currently there is a Yosys fork with QuickLogic support that would conflict with the version used in SymbiFlow. How do you propose to solve that until the QuickLogic support is in upstream Yosys?

Maybe create a separate conda env to be used with QuickLogic devices?

quicklogic/common/utils/flatten_layout.py
mkurc-ant: It seems that we do not have any of those in SymbiFlow... Which one should I use? The same as in other SymbiFlow related repos? (like prjxray?)

litghost (changes requested):
@mkurc-ant Line count looks much better now, thanks!

CI's are still red?  I know you know why, but you need to make the CI's green before this is ready for review?

mkurc-ant:
For the CI to pass we need a VPR version with the fix mentioned in the PR description. Unfortunately simply updating VPR package breaks CI for Xilinx 7-series. There is an ongoing effort to integrate recent packages (including VPR) with SymbiFlow: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2072

litghost:
> For the CI to pass we need a VPR version with the fix mentioned in the PR description. Unfortunately simply updating VPR package breaks CI for Xilinx 7-series. There is an ongoing effort to integrate recent packages (including VPR) with SymbiFlow: #2072

Agreed.  Looks like @acomodi almost has #2072 working, hopefully it is green now.

quicklogic/README.md
mkurc-ant: I've updated the README.md so that it reflects the current QuickLogic support state to be merged.

quicklogic/qlf_k4n8/devices/umc22/qlf_k4n8_umc22-arch.xml
```diff
@@ -0,0 +1,7998 @@
+<!--
```
litghost, line 1: Was this file generated by hand?  Seems like portions of this should be refactored into some generator scripts.  Thoughts?

quicklogic/qlf_k4n8/techmap/cells_sim.v
```diff
@@ -0,0 +1 @@
+// Simulation library with cells supported by this VPR architecture
```
litghost, line 1: Because these are empty, I assume we have a 1-to-1 relationship?  Maybe add a comment?

quicklogic/common/cmake/quicklogic_qlf_device.cmake
```diff
@@ -0,0 +1,99 @@
...
+
+    CACHE_PLACE_DELAY
+    CACHE_LOOKAHEAD
+    CACHE_ARGS
+      --constant_net_method route
+      --clock_modeling ideal # Do not route clocks
+      --place_delay_model delta_override
+      --place_delta_delay_matrix_calculation_method dijkstra
+      --router_lookahead extended_map
+      --disable_errors check_unbuffered_edges:check_route:check_place
```
litghost, line 90: Are all of these `disable_errors` required, or just copied from the xc7 init?

quicklogic/common/cmake/quicklogic_qlf_device.cmake
```diff
@@ -0,0 +1,99 @@
...
+  set(DEVICE_TYPE ${DEVICE}-virt)
+
+  set(PACKAGE ${DEVICE})
+
+  # .......................................................
+
+  # Arch XML base file
+  get_file_target(ARCH_XML_TARGET ${ARCH_XML})
+  if (NOT TARGET ${ARCH_XML_TARGET})
+      add_file_target(FILE ${ARCH_XML} SCANNER_TYPE)
```
litghost, line 48: Should be `SCANNER_TYPE xml`??

quicklogic/common/cmake/quicklogic_qlf_arch.cmake
```diff
@@ -0,0 +1,94 @@
...
+#      ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/common/utils/create_ioplace.py
+#    PLACE_TOOL_CMD "${CMAKE_COMMAND} -E env \
+#      PYTHONPATH=${symbiflow-arch-defs_SOURCE_DIR}/utils:$PYTHONPATH \
+#      \${PYTHON3} \${PLACE_TOOL} \
+#          --map \${PINMAP} \
+#          --blif \${OUT_EBLIF} \
+#          --pcf \${INPUT_IO_FILE} \
+#          --net \${OUT_NET}"
+
+    NO_PLACE_CONSTR
```
litghost, line 61: Without place constraints, IO pads cannot be specified?  Am I missing something?

common/cmake/file_targets.cmake
```diff
@@ -264,8 +264,12 @@ function(ADD_FILE_TARGET)
   endforeach()
 
   if(NOT ${ADD_FILE_TARGET_GENERATED} AND NOT ${ADD_FILE_TARGET_ABSOLUTE})
+    get_filename_component(DEST_PATH ${CMAKE_CURRENT_BINARY_DIR}/${ADD_FILE_TARGET_FILE} DIRECTORY)
     add_custom_command(
       OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${ADD_FILE_TARGET_FILE}
+      COMMAND
```
litghost, line 270: This probably means you have some weird `add_file_target` that contain paths in the name.  Probably a bad idea (consider something like `../../example.v`).  Probably worth scrubbing the `add_file_target` and removing the `weird` ones.  Also consider generating a warning and/or error on "weird" paths.

common/cmake/devices.cmake
```diff
@@ -1491,7 +1545,12 @@ function(ADD_FPGA_TARGET)
 
   get_target_property_required(VPR env VPR)
 
-  get_target_property_required(ROUTE_CHAN_WIDTH ${ARCH} ROUTE_CHAN_WIDTH)
+  # Use route channel width from the device. If not provided then use the
+  # one from the arch.
+  get_target_property(ROUTE_CHAN_WIDTH ${DEVICE} ROUTE_CHAN_WIDTH)
```
litghost, line 1550: Is this change still required?

common/cmake/devices.cmake
```diff
@@ -586,6 +599,15 @@ function(DEFINE_DEVICE)
     # Generate a rr_graph for a device.
     #
 
+    # Use the device specific channel with for the virtual graph if provided.
+    # If not use a dummy value (assuming that the graph will get patched
+    # anyways).
+    if("${DEFINE_DEVICE_ROUTE_CHAN_WIDTH}" STREQUAL "")
+      set(RRXML_VIRT_ROUTE_CHAN_WIDTH 6) # FIXME: Where did the number come from?
```
litghost, line 606: This value was the smallest value that could consistently route a wire.  Largest numbers result in large virtual graphs, which take more time to generate and write/read from disk.  An even number was chosen because it is required for BIDIR channel definitions.

common/cmake/devices.cmake
```diff
@@ -547,8 +556,12 @@ function(DEFINE_DEVICE)
     RR_PATCH_TOOL ${DEFINE_DEVICE_ARCH} RR_PATCH_TOOL
   )
   get_target_property_required(RR_PATCH_CMD ${DEFINE_DEVICE_ARCH} RR_PATCH_CMD)
-  get_target_property_required(ROUTE_CHAN_WIDTH ${DEFINE_DEVICE_ARCH}
-    ROUTE_CHAN_WIDTH)
+
+  if("${DEFINE_DEVICE_ROUTE_CHAN_WIDTH}" STREQUAL "")
+    get_target_property_required(ROUTE_CHAN_WIDTH ${DEFINE_DEVICE_ARCH} ROUTE_CHAN_WIDTH)
```
litghost, line 561: Is this `if` statement here twice?  See line 602?

common/cmake/devices.cmake
```diff
@@ -490,6 +490,7 @@ function(DEFINE_DEVICE)
   #   [CACHE_PLACE_DELAY]
   #   [CACHE_LOOKAHEAD]
   #   [CACHE_ARGS <args>]
+  #   [ROUTE_CHAN_WIDTH]
```
litghost, line 493: Should be `[ROUTE_CHAN_WIDTH <width>]`, yes?

litghost: Need `[DONT_INSTALL]` in arg summary?

quicklogic/qlf_k4n8/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    --router_lookahead extended_map "
+)
+
+# Define the architecture
+quicklogic_define_qlf_arch(
+  FAMILY   ${FAMILY}
+  ARCH     qlf_k4n8
+  VPR_ARGS ${VPR_ARCH_ARGS}
+
+  ROUTE_CHAN_WIDTH 100 # Will be overriden by each device
```
litghost, line 23: I would make this value smaller?

quicklogic/qlf_k4n8/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+include(../common/cmake/quicklogic_qlf_device.cmake)
+include(../common/cmake/quicklogic_board.cmake)
+
+get_filename_component(FAMILY ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+
+add_subdirectory(techmap)
+
+# VPR args
+set(VPR_ARCH_ARGS "\
+    --clock_modeling ideal \
```
litghost, line 11: This seems weird?  Does this arch have no clock routing?  Please add a clear comment what is going on here.

quicklogic/qlf_k4n8/utils/pinmap_parse.py
```diff
@@ -0,0 +1,348 @@
+#!/usr/bin/env python3
+import argparse
```
litghost, line 2: Please add root level docstring explaining what is in here.

quicklogic/common/utils/flatten_layout.py
```diff
@@ -0,0 +1,515 @@
...
+"""
+This script loads VPR architecture definition and flattens all layouts defined
```
litghost, line 3: This script probably belongs in VPR upstream, probably as a executable in [libarchfpga](https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/libs/libarchfpga).  You've effectively re-written the behavior of libarchfpga here, and if you have a disagreement with VPR's implementation, it could cause significant confusion.

@vaughnbetz Thoughts?

quicklogic/qlf_k4n8/yosys/synth.tcl
```diff
@@ -0,0 +1,32 @@
...
+# Load the QuickLogic qlf_k4n8 support plugin. Note that this is only temporary
+# until support for the device is merged into the upstream Yosys
+plugin -i ql-qlf-k4n8
+yosys -import
+
+# Read VPR cells library
+read_verilog -lib $::env(TECHMAP_PATH)/cells_sim.v
+
+# Synthesize
+if {[info exists ::env(SYNTH_OPTS)]} {
```
litghost, line 12: I think flattening this probably makes sense.  E.g.

```
set extra_opts
if {[info exists ::env(SYNTH_OPTS)]} {
  lextend extra_opts $::env(SYNTH_OPTS)
}
synth_quicklogic -family qlf_k4n8 $extra_opts
```

tests/1-wire/CMakeLists.txt
```diff
@@ -1,6 +1,17 @@
+add_file_target(FILE wire.v)
+
 add_simple_test(
```
litghost, line 3: Rather than copy paste every test under tests for the QL OpenFPGA stuff, please adjust `add_simple_test` to correctly do the right thing for the QL OpenFPGA arch.  There should be enough data in the arch and device definitions to do so.  The entire reason we use `add_simple_test` instead of `add_fpga_test` is to handle special cases like this.

quicklogic/common/utils/flatten_layout.py
mkurc-ant: @litghost I'm going to remove this script as it is not needed here anymore and it looks like it ended up here by mistake.

Answering your comment: Yes this is a re-implementation of the behavior of a VPR part. It is needed when you have an architecture with layout containing other constructs than `<single>` and you want to add FASM prefixes to tiles. Currently VPR provides only support for prefixing `<single>` tiles hence such a layout needs to be re-written so that it uses only those tags.

quicklogic/qlf_k4n8/devices/umc22/qlf_k4n8_umc22-arch.xml
mkurc-ant: Yes, this architecture has been written manually and then processed by a script. The flow of  processing the architecture is still under development. I think that subsequent versions may have all comments stripped.

common/cmake/devices.cmake
mkurc-ant: Thanks for the explanation. I'll include it in a comment

quicklogic/qlf_k4n8/CMakeLists.txt
mkurc-ant: This architecture is based on the OpenFPGA project where for a chip the clock network is global and non-configurable hence VPR need not to worry about the routing.

quicklogic/common/utils/flatten_layout.py
litghost: I understand the reason such a script might exist, I'm saying re-implementing it in python, rather than just calling into the relevant VPR code seems like a bad idea.

quicklogic/qlf_k4n8/CMakeLists.txt
litghost: Like I said above, add a comment.

common/cmake/devices.cmake
mkurc-ant: Currently yes. There is an ongoing effort to use externally supplied rr graphs. With that done specifying channel width per device will no longer be necessary.

quicklogic/common/cmake/quicklogic_qlf_device.cmake
mkurc-ant: Yes, they got copied and are not needed for this architecture. I removed them.

quicklogic/common/cmake/quicklogic_qlf_device.cmake
mkurc-ant: Right, fixed.

quicklogic/common/cmake/quicklogic_qlf_arch.cmake
mkurc-ant: That is correct. I added a comment explaining it.

common/cmake/devices.cmake
mkurc-ant: I've integrated those two conditions into a single one.

common/cmake/devices.cmake
mkurc-ant: I updated the function documentation.

quicklogic/qlf_k4n8/CMakeLists.txt
mkurc-ant: Ok. Changed to a smaller value.

quicklogic/qlf_k4n8/CMakeLists.txt
mkurc-ant: Comment added.

quicklogic/qlf_k4n8/utils/pinmap_parse.py
mkurc-ant: Comment added.

quicklogic/qlf_k4n8/yosys/synth.tcl
mkurc-ant: I agree. Thanks for the hint. I updated the script.

mkurc-ant: Unfortunately the approach you suggested doesn't work :/ Yosys treats `$extra_opts` as an additional argument even if the list is empty. I'd stick with the original code for now.

tests/1-wire/CMakeLists.txt
mkurc-ant: Ok. I've reworked that.

common/cmake/file_targets.cmake
mkurc-ant: So you are suggesting that `add_file_target` should only be used for targets that write to the "current binary directory"? I think that we already violate that for all file targets added in `add_fpga_target`.

quicklogic/qlf_k4n8/techmap/cells_sim.v
mkurc-ant: Comment added.

common/cmake/devices.cmake
litghost: The lines are 602 are still there?  Am I missing something?

common/cmake/devices.cmake
litghost: I don't see the comment.  Did you forget to push?

litghost (changes requested):
@mkurc-ant I believe you've forgetting to push, as many responses say you've fixed things, but the diff shows the comments still stand?  Please check what is going on?

mkurc-ant:
@litghost Should be fine now. Sorry for the confusion.

PR #2039, opened 2021-02-18
xc7: techmap: Remove localparam from generate blocks

### LINKED ISSUE #2020: xc7 cells_map.v contains illegal verilog constructs now rejected by latest yosys
See https://github.com/YosysHQ/yosys/issues/2579.

An example of this construct can be seen here: https://github.com/SymbiFlow/symbiflow-arch-defs/blob/45c83a2592923b9a785aa05c9464b5f453ebd151/xc/xc7/techmap/cells_map.v#L1140-L1146

I don't know how many more instances of this there are in the codebase, and I don't know what a proper fix for this would look like.

dnltz:
Yosys throws following error:
	Parameter x with non-constant value!

This only happens on localparams which are included in a generate if block.
Move the localparam statement out of the generate fixes the issue.

Closes #2020 

Does not feel right to add nested "? :" statement but do not know a better solution for else if blocks.

### DIFF
--- a/xc/xc7/techmap/cells_map.v
+++ b/xc/xc7/techmap/cells_map.v
@@ -1079,21 +1079,11 @@ module RAMB18E1 (
 
   end
 
-if(RAM_MODE == "SDP" && READ_WIDTH_A == 36) begin
-    localparam EFF_READ_WIDTH_A = 18;
-    localparam EFF_READ_WIDTH_B = 18;
-end else begin
-    localparam EFF_READ_WIDTH_A = READ_WIDTH_A;
-    localparam EFF_READ_WIDTH_B = READ_WIDTH_B;
-end
+  localparam EFF_READ_WIDTH_A = (RAM_MODE == "SDP" && READ_WIDTH_A == 36) ? 18 : READ_WIDTH_A;
+  localparam EFF_READ_WIDTH_B = (RAM_MODE == "SDP" && READ_WIDTH_A == 36) ? 18 : READ_WIDTH_B;
 
-if(RAM_MODE == "SDP" && WRITE_WIDTH_B == 36) begin
-    localparam EFF_WRITE_WIDTH_A = 18;
-    localparam EFF_WRITE_WIDTH_B = 18;
-end else begin
-    localparam EFF_WRITE_WIDTH_A = WRITE_WIDTH_A;
-    localparam EFF_WRITE_WIDTH_B = WRITE_WIDTH_B;
-end
+  localparam EFF_WRITE_WIDTH_A = (RAM_MODE == "SDP" && WRITE_WIDTH_B == 36) ? 18 : WRITE_WIDTH_A;
+  localparam EFF_WRITE_WIDTH_B = (RAM_MODE == "SDP" && WRITE_WIDTH_B == 36) ? 18 : WRITE_WIDTH_B;
 
   wire REGCLKA;
   wire REGCLKB;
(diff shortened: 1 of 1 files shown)


litghost:
This looks reasonable, did you test with the latest Yosys and see if it accepts the Verilog?  The CI should verify that the flow doesn't break.

dnltz:
Yes, I used the " linux-64/yosys-v0.9_5191_geff18a2b-20210209_143455_py37.tar.bz2" version which is 9 days old. I run into the same issue from #2020 and with these changes it works again.

dnltz:
@litghost - The CI output looks a little bit strange. Can we restart it?

xc/xc7/techmap/cells_map.v
```diff
@@ -1497,40 +1477,20 @@ module RAMB36E1 (
...
 
-if(RAM_MODE == "SDP" && WRITE_WIDTH_B > 36) begin
-    localparam EFF_WRITE_WIDTH_A = 36;
-    localparam EFF_WRITE_WIDTH_B = 36;
-end else begin
-    localparam EFF_WRITE_WIDTH_A = WRITE_WIDTH_A;
-    localparam EFF_WRITE_WIDTH_B = WRITE_WIDTH_B;
-end
+  localparam EFF_WRITE_WIDTH_A = (RAM_MODE == "SDP" && WRITE_WIDTH_B == 36) ? 18 : WRITE_WIDTH_A;
+  localparam EFF_WRITE_WIDTH_B = (RAM_MODE == "SDP" && WRITE_WIDTH_B == 36) ? 18 : WRITE_WIDTH_B;
```
acomodi, line 1484: Hi @dnltz, These lines are not equivalent to the original version, causing CI error.

Should be like:
```
localparam EFF_READ_WIDTH_A = (RAM_MODE == "SDP" && READ_WIDTH_A > 36) ? 36 : READ_WIDTH_A;
```

For reference, the CI error is related to BRAM_READ_WIDTH:
```
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/env/conda/envs/symbiflow_arch_def_base/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/env/conda/envs/symbiflow_arch_def_base/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/__main__.py", line 4, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/fasm2bels.py", line 443, in main
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/fasm2bels.py", line 107, in process_tile
    PROCESS_TILE[tile_type](top.conn, top, tile, tile_features)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/models/bram_models.py", line 1382, in process_bram
    sites.append(process_bram_site(top, brams[bram], bram_features[bram]))
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/third_party/symbiflow-xc-fasm2bels/fasm2bels/models/bram_models.py", line 518, in process_bram_site
    assert 'READ_WIDTH_B_18' in set_features
```

acomodi (changes requested):
Comment in

xc/xc7/techmap/cells_map.v
dnltz: Ah! Thanks for pointing me out the issue.

acomodi (approved):
LGTM. I believe it might be worth to also update yosys to the latest version and check this indeed solves the original problem of the `localparam` within  a generate/if statement. @litghost thoughts on that?

litghost:
> LGTM. I believe it might be worth to also update yosys to the latest version and check this indeed solves the original problem of the `localparam` within a generate/if statement. @litghost thoughts on that?

I asked @dnltz if they tested against the new yosys, and they said they had.

PR #2053, opened 2021-02-25
Add PCIe support

kowalewskijan:
This PR adds PCIe support for symbiflow-arch-defs for xc7 architectures.

- [x] PCIe primitive definition
- [x] PCIe techmapping
- [x] PCIe basic test (for now there is a simple instantiation of the block to check how far we can get in the flow)

Target architectures:
- [x] xc7a50t
- [x] xc7a100t
- [x] xc7a200t

Additional work:
- [ ] add autogeneration of XMLs and cells_sim/map modules to the build-time flow


xc/common/primitives/pcie_2_1/pcie_2_1.pb_type.xml
```diff
@@ -0,0 +1,730 @@
+<pb_type name="PCIE_2_1" num_pb="1" blif_model=".subckt PCIE_2_1_VPR">
```
litghost, line 1: If this is generated, don't check it in?  Generate it?

acomodi (approved):
LGTM

As soon as a new prjxray-db is generated (which will contain info on the clock ports), we may proceed with the addition of the autogeneration tool to get the XMLs of the PCIE_2_1 and GTP primitives.

PR #2061, opened 2021-03-01
Mark CARRY4_VPR as lib_whitebox

Ravenslofty:
`CARRY4_VPR` is parametric, but also marked as a `blackbox`; this confuses ABC9 because you can't monomorphise blackboxes, and so it has no way of knowing if the timings it computes are accurate. I believe it is correct to instead mark it as `lib_whitebox` so that ABC9 can monomorphise it properly.

### DIFF
--- a/xc/xc7/techmap/cells_sim.v
+++ b/xc/xc7/techmap/cells_sim.v
@@ -85,7 +85,6 @@ assign COUT = CIN;
 
 endmodule
 
-(* abc9_box, blackbox *)
 module CARRY4_VPR(O0, O1, O2, O3, CO0, CO1, CO2, CO3, CYINIT, CIN, DI0, DI1, DI2, DI3, S0, S1, S2, S3);
   parameter CYINIT_AX = 1'b0;
   parameter CYINIT_C0 = 1'b0;


litghost:
Reposting my question from https://github.com/SymbiFlow/symbiflow-examples/issues/120, will `lib_whitebox` allow ABC9 to optimize through the CARRY4? Because we specifically need ABC9 to leave the CARRY4 alone at this part of the flow. I want ABC9 to be able to do timing analysis of the CARRY4, but only want it handle the LUT's on the outputs, not try to optimize the carry logic itself.

litghost:
Per conversation in https://github.com/SymbiFlow/symbiflow-examples/issues/120, this change needs to be paired with adding the `keep` attribute to the `CARRY4_VPR` cells to prevent ABC from optimization across the `CARRY4_VPR` cell.

litghost:
I'm going to kick off the full CI.  I expect we'll get some packing failures as ABC optimizes away the output de-congestion logic.

Ravenslofty:
Yeah, @kgugala asked me to file this PR so that it could get a CI stress test. I think it's best to wait for CI to finish first and then decide what to do from there.

litghost:
So with this change the CI fails with:
```
/home/travis/build/litex-hub/litex-conda-eda/workdir/conda-env/conda-bld/vtr-optimized_1608176351341/work/vpr/src/place/place_macro.cpp:135 find_all_the_macro: Assertion 'cluster_ctx.clb_nlist.net_sinks(curr_net_id).size() == 1' failed.
```

Which is basically a long winded way of saying "the carry chain no longer makes sense".  In the past this caused by ABC9 optimizing the carry chain in a way that is not valid for the VPR flow.  The specific problem varies, and the general solution has been to tell ABC to not mess with the CARRY4 ports.   This was the reason that `blackbox` was used.

Ravenslofty:
Force-pushed with the `(* abc9_box, blackbox *)` attributes removed entirely; you'll need to do the honours again.

litghost:
Result does appear to be a little worse with `abc9_box` removed.  I'm going to run it again to see if the result is consistent, this might be noise.

litghost:
Results from second run are closer to master, merging.  I'll create an issue about adding these attributes back once yosys supports "look, don't touch" in ABC9.

PR #2072, opened 2021-03-04
Update packages

acomodi:
This PR updates the following packages that are generated from the HDL conda-build project:

- yosys
- yosys-plugins
- VTR
- prjxray-tools

There seems to be an issue with VTR around the RR graph, as some of the ROI tests do now fail.

This PR also moves to use the fasm package from PyPI


acomodi:
> There seems to be an issue with VTR around the RR graph, as some of the ROI tests do now fail.

It seems that the problem corresponds to the one described on IRC. Basically, the placer does not preserve some of the constraints, resulting in an unroutable situation (as well as most probably broken bitstreams).

acomodi:
@Xiretza Sure, that is a good idea. I'll update the PR soon.

acomodi:
@Xiretza actually, we need to wait for a new prjxray-db update as the fixes in https://github.com/SymbiFlow/prjxray/pull/1609 are not checked in the current db.

(cc @mithro)

acomodi:
The updated VTR and yosys packages went green. The next step was the prjxray-db update with the new LVDS and TMDS iostandards, but it requires https://github.com/SymbiFlow/xc-fasm/pull/11 to go green.

Until new install packages are generated, we need to fix all the xc-fasm versions in all places that use it (e.g. symbiflow-examples) to the [current master](https://github.com/SymbiFlow/xc-fasm/commit/e12f31334e96fedf3af86d13cf51f70ad2270f5f) as the change to fasm2frames will break the bitstream generation step.

requirements.txt
```diff
@@ -19,12 +19,12 @@ textx
 tinyfpgab
 tinyprog
 yapf==0.26.0
+fasm
 git+https://github.com/SymbiFlow/symbiflow-rr-graph.git#egg=rr-graph
 # TODO: https://github.com/SymbiFlow/python-fpga-interchange/issues/11
 git+https://github.com/litghost/rapidyaml.git@fixup_python_packaging#egg=rapidyaml&subdirectory=api/python
```
litghost, line 25: This `rapidyaml` line can be dropped.

requirements.txt
```diff
@@ -19,12 +19,12 @@ textx
 tinyfpgab
 tinyprog
 yapf==0.26.0
+fasm
 git+https://github.com/SymbiFlow/symbiflow-rr-graph.git#egg=rr-graph
```
litghost, line 23: @mithro I believe this package is on PyPI soon or now?

xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -361,10 +361,9 @@ def __init__(
                 clock = self.clock_blocks[cname]
 
                 for sink in sinks:
-                    assert sink in ports, (
-                        cname,
-                        sink,
-                    )
+                    if sink not in ports:
```
litghost, line 364: I don't believe this change is right?

acomodi: There are multiple reasons for this:

- apparently there was a change that caused all the `$false` or `$true` nets of bus ports to be named with the full name bus instead of the single bit:
  - e.g. PS7 example: `FCLKCLKTRIGN=$false` where the `FCLKCLKTRIGN` port is a multi-bit bus. This causes issues with https://github.com/SymbiFlow/symbiflow-arch-defs/blob/098fe2e8e724bca3902b8121fb254caaed7cb5d3/xc/common/utils/prjxray_create_place_constraints.py#L57-L60.

The above can be actually treated in a better way, and I'll address it.

- with the GTP_COMMON, there is a situation in which either the `GTREFCLK0` or `GTREFCLK1` will be present in the blif.
Given that the GTP_COMMON is a clock-related tile which needs to be tightly coupled with a corresponding IBUFDS_GTE2, we need to add it to the CLOCKS dict, but the fact is that, to avoid impossible routes to route GND, to [the unused GTREFCLK are dropped](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/3b3a14a8864ea9c9c20688cf76b1b3aeaf03080b/xc/xc7/techmap/cells_map.v#L8428-L8633).

litghost: > apparently there was a change that caused all the $false or $true nets of bus ports to be named with the full name bus instead of the single bit:

This sounds like a bug in the yosys blif writer?

acomodi: Rather than the blif writer, this is caused by something else, I could not identify what yet though.

The top_synth.json has a difference regarding ports connected to GND:

Old yosys version:
```
...
"FCLKCLK": [ 476, 477, 478, 479 ],
"FCLKCLKTRIGN": [ "0", "0", "0", "0" ],
"FCLKRESETN": [ 480, 481, 482, 483 ], 
...
```

New yosys version:
```
...
"FCLKCLK": [ 476, 477, 478, 479 ],
"FCLKCLKTRIGN": [ "0" ],
"FCLKRESETN": [ 480, 481, 482, 483 ], 
...
```

litghost: I would take a look at the cell definition (e.g. `cells_sim.v`).

acomodi: In yosys, the PS7 definition did not change between the two versions (in the [cells_xtra.v](https://github.com/YosysHQ/yosys/blob/master/techlibs/xilinx/cells_xtra.v)) and neither nothing changed in arch-defs as far. (The PS7 is just an example here, I should verify for other primitives as well).

I shall also run bisect probably to see what change is causing this behaviour.

acomodi: After some debugging, I discovered the reason behind this. Basically, it corresponds to a techmap issue with the PS7 block around the handling of unconnected ports, so this is fixable in arch-defs.

litghost (approved):
LGTM, merge if green

litghost (changes requested):
Failure in the SerDes diff fasm tests.  Looks like an issue with the PLL features:
```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/serdes && diff -u /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/serdes/serdes_basys3_idelay_8_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/top.bit.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/serdes/serdes_basys3_idelay_8_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/interchange/serdes_basys3_idelay_8_sdr_vivado_interchange.bit.fasm
--- /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/serdes/serdes_basys3_idelay_8_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/top.bit.fasm	2021-03-16 06:02:52.051872259 -0700
+++ /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7-vendor/build/xc/xc7/tests/serdes/serdes_basys3_idelay_8_sdr/artix7-xc7a50t-bottom-virt-xc7a50t-bottom-test/interchange/serdes_basys3_idelay_8_sdr_vivado_interchange.bit.fasm	2021-03-16 06:06:04.272769378 -0700
@@ -308,7 +308,6 @@
 CMT_TOP_L_UPPER_T_X106Y44.PLLE2_ADV.CLKOUT2_CLKOUT1_HIGH_TIME[0]
 CMT_TOP_L_UPPER_T_X106Y44.PLLE2_ADV.CLKOUT2_CLKOUT1_LOW_TIME[0]
 CMT_TOP_L_UPPER_T_X106Y44.PLLE2_ADV.CLKOUT2_CLKOUT1_OUTPUT_ENABLE[0]
-CMT_TOP_L_UPPER_T_X106Y44.PLLE2_ADV.CLKOUT2_CLKOUT2_NO_COUNT[0]
 CMT_TOP_L_UPPER_T_X106Y44.PLLE2_ADV.CLKOUT3_CLKOUT1_HIGH_TIME[0]
 CMT_TOP_L_UPPER_T_X106Y44.PLLE2_ADV.CLKOUT3_CLKOUT1_LOW_TIME[0]
 CMT_TOP_L_UPPER_T_X106Y44.PLLE2_ADV.CLKOUT3_CLKOUT2_NO_COUNT[0]
```

acomodi:
> Failure in the SerDes diff fasm tests. Looks like an issue with the PLL features:

Hmm, where can I find the CI log? At the moment I see that the XC7 Vendor is still pending. I'll try to reproduce locally

mkurc-ant:
I think I know the cause of the diff FASM failure. There is a bug in PLL decoding in fasm2bels. It was to be merged along with https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/44. It may be worth to separate all the fixes I've made in that PR from BUFHCE handling change. I wanted to merge them together but it has been stale for too long.

mkurc-ant:
I've updated the https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/44 PR so that it includes only PLL/MMCM decoding fixes. With fasm2bels from there this PR should get green.

PR #2073, opened 2021-03-04
Add utility tool for generating prototype files for new primitives

tmichalak:
This PR adds the **build_primitive_proto** script that can be used for generating prototypes of the pb_type / model / cell_[sim|map] module from segbits and site definitions


kowalewskijan:
This tool can be reworked to support attribute and ports files introduced in the prjxray PR: https://github.com/SymbiFlow/prjxray/pull/1614
Using these JSONs would be much more clear.

PR #2082, opened 2021-03-08
Retry marking CARRY4_VPR as whitebox

Ravenslofty:
After talking to Eddie, I noticed that CARRY4_VPR was not marked as `(* abc9_carry *)`, so I want to see if this stops ABC9 from sweeping the cells away.

cc @litghost @kgugala


litghost:
CI failure on xc7 install looks like a kokoro infrastructure error, I'll re-run it in a bit.  xc7 CI was green, so that is good.

PR #2089, opened 2021-03-10
Added loading rr graph nodes directions

mkurc-ant:
This PR adds reading of VPR node directions to the rr graph XML handling library. It also exposes the `load_edges` option.

### DIFF
--- a/utils/lib/rr_graph/graph2.py
+++ b/utils/lib/rr_graph/graph2.py
@@ -278,14 +278,14 @@ class Graph(object):
             assert key not in self.loc_map
             self.loc_map[key] = loc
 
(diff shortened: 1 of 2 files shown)


litghost:
I might be missing something, but do we have double copies right now between here and https://github.com/SymbiFlow/symbiflow-rr-graph ?

mkurc-ant:
@litghost Hmm, I wasn't aware that we've separated the code to a different repo already... If that's the case then I can make the change there.

mkurc-ant:
Closing this one. I've opened a similar PR to the repository with the library isolated: https://github.com/SymbiFlow/symbiflow-rr-graph/pull/4

PR #2109, opened 2021-03-17
gh-actions: ci: use ubuntu-18.04

acomodi:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
 
   Run-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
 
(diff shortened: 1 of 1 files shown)


litghost:
Don't have a problem with this.  Are you planning to do a follow up and update this to `ubuntu-20.04`?

acomodi:
> Are you planning to do a follow up and update this to ubuntu-20.04?

Yes. The main reason for this PR was to fall back to 18.04 as the `ubuntu-latest` is now giving [problems with GH CI](https://github.com/SymbiFlow/symbiflow-arch-defs/runs/2133527914) on master (the `latest` version must have changed recently):

```
Package g++-6 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'g++-6' has no installation candidate
```

PR #2113, opened 2021-03-19
Retry marking CARRY4_VPR as whitebox

Arvind-Srinivasan:
Rebased a PR (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2082) previously made by Dan Ravensloft

Signed-off-by: Dan Ravensloft <dan.ravensloft@gmail.com>

### DIFF
--- a/xc/xc7/techmap/cells_sim.v
+++ b/xc/xc7/techmap/cells_sim.v
@@ -75,7 +75,7 @@ assign OUT = O ^ S;
 endmodule
 
 (* abc9_box, blackbox *)
-module CARRY_COUT_PLUG(input CIN, output COUT);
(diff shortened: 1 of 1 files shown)


litghost:
I don't have a problem with trying, but what changed such that we expect this to return a different result?

Arvind-Srinivasan:
My understanding was there would be an install issue which would cause CI to fail otherwise.

litghost:
> My understanding was there would be an install issue which would cause CI to fail otherwise.

I believe last time we tried this it caused an "as expected" packing issue.  The solution was then to remove the black/whitebox completely so that the cell was opaque to abc.  This resulted in a smaller negative (or eligible) CPD impact, and was merged.

Arvind-Srinivasan:
My bad, then I will close.

PR #2119, opened 2021-04-02
Upstream quicklogic install

acomodi:
This PR enables QuickLogic devices installation.

A CMake cache variable is added to control which FPGA family will get installed.


kkumar23 (commented):
Hi @acomodi : The tool wrappers have mix of eos-s3 device and qlf_k4n8.
I think if we are planning for only qlf_k4n8, we need to remove the eos-s3 related support/information

mkurc-ant:
@kkumar23 Hi Kishor. I've taken over the work on supporting installation of QuickLogic devices in the upstream. I'm currently working on making `make install` to write only files relevant to the specific device (eg. qlf_k4n8).

Currently the wrapper scripts have hard coded device names within. So do you want to remove all references to EOS-S3 from them for now?

kkumar23:
@mkurc-ant : Since master branch will have only the qlf_k4n8, we should remove eos-s3. We have make it generic so that in future any new device added it will be easy for anyone to add.

mkurc-ant:
With this PR `make install` works both for Xilinx 7-series and QuickLogic qlf_k4n8 but not for both at the same time. I've extended the control over which devices / architectures / families are installed and which are not installed (see `common/cmake/install.cmake` for comments and details).

The toolchain wrappers for QuickLogic qlf_k4n8 support only this particular device.

.github/kokoro/ql.sh
```diff
@@ -2,8 +2,9 @@
 
 SCRIPT_SRC="$(realpath ${BASH_SOURCE[0]})"
 SCRIPT_DIR="$(dirname "${SCRIPT_SRC}")"
+INSTALL_DIR="$(pwd)/github/${KOKORO_DIR}/install"
 
-export CMAKE_FLAGS="-GNinja"
+export CMAKE_FLAGS="-GNinja -DINSTALL_FAMILIES=qlf_k4n8 -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR}"
```
kgugala, line 7: INSTALL_FAMILIES flag usage seems to be inconsistent. In 7 series FAMILY is the whole xc7, while devices are selected with INSTALL_DEVICES. Here `INSTALL_FAMILIES` is used to select a single device

quicklogic/common/cmake/quicklogic_board.cmake
```diff
@@ -132,7 +132,7 @@ function(ADD_QUICKLOGIC_BOARD)
     add_custom_command(
       OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${CLKMAP_CSV}
       COMMAND
-        ${CMAKE_COMMAND} -E create_symlink 
+        ${CMAKE_COMMAND} -E copy
```
kgugala, line 135: why do we need to copy this?

quicklogic/common/cmake/quicklogic_install.cmake
```diff
@@ -47,55 +54,14 @@ function(DEFINE_QL_TOOLCHAIN_TARGET)
...
-  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/common/utils/fasm2bels.py
-          DESTINATION bin/python
-          PERMISSIONS WORLD_EXECUTE WORLD_READ OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE)
-
-  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/common/utils/quicklogic-fasm/quicklogic_fasm/bitstream_to_jlink.py
-          DESTINATION bin/python
-          PERMISSIONS WORLD_EXECUTE WORLD_READ OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE)
-
-  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/common/utils/quicklogic-fasm/quicklogic_fasm/bitstream_to_header.py
+   install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/utils/split_inouts.py
```
kgugala, line 60: indentation

quicklogic/common/cmake/quicklogic_install.cmake
```diff
@@ -116,45 +82,28 @@ function(DEFINE_QL_TOOLCHAIN_TARGET)
...
-  # Unsupported family
-  else()
-    message(FATA_ERROR "Family '${FAMILY}' is not suitable for installed toolchain")
-  endif()
+  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/${FAMILY}/utils/create_ioplace.py
+          DESTINATION bin/python
+          PERMISSIONS WORLD_EXECUTE WORLD_READ OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE)
 
+  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/${FAMILY}/devices/umc22/interface-mapping_24x24.xml
+          DESTINATION share/symbiflow/arch/qlf_k4n8-qlf_k4n8_umc22_qlf_k4n8-qlf_k4n8_umc22
```
kgugala, line 101: can't this filename and install be somehow calculated? hardcoding it see doesn't feel right (this file is common to the whole QL flow)

quicklogic/common/cmake/quicklogic_install.cmake
```diff
@@ -116,45 +82,28 @@ function(DEFINE_QL_TOOLCHAIN_TARGET)
...
+          PERMISSIONS WORLD_EXECUTE WORLD_READ OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE)
 
+  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/${FAMILY}/devices/umc22/qlf_k4n8-qlf_k4n8_umc22_24x24.csv
+          DESTINATION share/symbiflow/arch/qlf_k4n8-qlf_k4n8_umc22_qlf_k4n8-qlf_k4n8_umc22
```
kgugala, line 105: ditto

quicklogic/common/toolchain_wrappers/ql_symbiflow
```diff
@@ -0,0 +1,459 @@
...
+      exit 1
+    elif ! [ -f "$SOURCE/$PCF" ]; then
+      echo "The pcf file: $PCF is missing at: $SOURCE"
+      exit 1
+    fi
+  fi
+  if [ -z "$DEVICE" ]; then
+    echo "DEVICE name is missing. Refer -h/--help"
+    exit 1
+  elif ! [[ "$DEVICE" =~ ^(ql-eos-s3_wlcsp|ql-ap3-ql745a_ql745a|qlf_k4n8_qlf_k4n8)$ ]]; then
```
kgugala, line 289: why do we check for eos-s3 and ap3? those are not supported by this toolchain

quicklogic/common/toolchain_wrappers/symbiflow_generate_constraints
```diff
@@ -0,0 +1,33 @@
...
+NET=$3
+PART=$4
+DEVICE=$5
+ARCH_DEF=$6
+
+if [[ "$DEVICE" =~ ^(qlf_k4n8_qlf_k4n8)$ ]];then
+	PINMAPCSV="qlf_k4n8-qlf_k4n8_umc22_24x24.csv"
+	DEVICE_1="qlf_k4n8-qlf_k4n8_umc22"
+	PINMAPXML="interface-mapping_24x24.xml"
+else
```
kgugala, line 18: can there be `else`? I suppose we should fail here

kgugala (changes requested):

acomodi (commented):
Comments in.

I believe that in this PR or in a follow-up one we need to add an install test similar to the xc7 one, to verify that the installed toolchain is indeed working correctly.

See https://github.com/SymbiFlow/symbiflow-arch-defs/tree/master/xc/xc7/tests/install_test

.github/kokoro/ql.sh
mkurc-ant: It looks that way but in this case `qlf_k4n8` refers both to the device and the family (same name). See: https://github.com/antmicro/symbiflow-arch-defs/blob/4b4854ef256a4de4a3672e6b1f076161a11b56ef/quicklogic/qlf_k4n8/CMakeLists.txt#L6

The family may be renamed to eg. `qlf` but I suppose that it will come later as more QuickLogic architectures are added.

quicklogic/common/cmake/quicklogic_board.cmake
mkurc-ant: Symlinking works when running everything from the repo using CMake but creates issues with installation. Basically CMake installs the symlink itself rather than the real file. There is no option in the CMake `install` command like eg. "resolve symlinks". I couldn't solve that in any way in a CMake script as at the  generation stage the link isn't present yet so it cannot be resolved and I cannot identify the real path of a file to pass it to the `install` command.

And more generally, the file is symlinked/copied to maintain consistency of SymbiFlow across all architectures. I could make the flow use the original file directly but that would require more changes to other CMake scripts.

quicklogic/common/cmake/quicklogic_install.cmake
mkurc-ant: Right. I see that this file belongs actually to a device not the arch and is now being installed with the arch. Need to fix that to be more generic.

quicklogic/common/cmake/quicklogic_install.cmake
mkurc-ant: same answer as above

quicklogic/common/cmake/quicklogic_install.cmake
mkurc-ant: Fixed.

quicklogic/common/cmake/quicklogic_install.cmake
mkurc-ant: Fixed.

quicklogic/common/cmake/quicklogic_install.cmake
mkurc-ant: Fixed.

quicklogic/common/toolchain_wrappers/ql_symbiflow
mkurc-ant: Right, removed all the reference to eos-s3 and ap3.

quicklogic/common/toolchain_wrappers/symbiflow_generate_constraints
mkurc-ant: I've rewritten the condition so that it falls back to generic file names for other devices than `qlf_k4n8`. The actual check if the device provided by the user is supported is performed in `ql_symbiflow` so no need for failing here.

quicklogic/common/toolchain_wrappers/symbiflow_generate_constraints
```diff
@@ -0,0 +1,33 @@
...
+PCF=$1
+EBLIF=$2
+NET=$3
+PART=$4
+DEVICE=$5
+ARCH_DEF=$6
+
+if [[ "$DEVICE" =~ ^(qlf_k4n8_qlf_k4n8)$ ]];then
+	PINMAPCSV="qlf_k4n8-qlf_k4n8_umc22_24x24.csv"
+	DEVICE_1="qlf_k4n8-qlf_k4n8_umc22"
```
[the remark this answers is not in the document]
mkurc-ant: This is to handle some irregularities of device naming. `DEVICE` is what user specifies and `DEVICE_1` is what the toolchain expects.

common/cmake/install.cmake
```diff
@@ -1,3 +1,290 @@
+function(PARSE_INSTALL_LISTS)
```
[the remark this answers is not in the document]
mkurc-ant: I've updated the comments. Do you have in mind any other way than setting the global variables in CMake?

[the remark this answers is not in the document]
acomodi: Not sure, but I think using cached vars is ok. In general though, the way things are handled in arch-defs CMake infrastructure is to either pass information either through target properties or through function arguments. Usage of global variables is reserved for special cases only.

kkumar23:
@mkurc-ant This PR is still not merged.

mkurc-ant:
@kkumar23 It looks like the CI has failed  for an unknown reason. Probably there is an issue with the infrastructure. I'm investigating it.

kgugala (approved):
LGTM

PR #2120, opened 2021-04-02
QuickLogic: Upstream basic tests

acomodi:
This PR upstreams some of the basic tests for the quicklogic qlf_k4n8 device.


mkurc-ant:
The `shift_reg_1024` test case fails because it is too big to fit in the device:
```
Failed to find device which satisfies resource requirements required: io: 4, clb: 1024 (available io: 1536, clb: 576)
```
Since the device has been resized from 32x32 to 24x24 it now contains less `clb` tiles so the design needs to be updated.

mkurc-ant:
@lnsharma The `shift_reg_1024` test doesn't fit into the 24x24 device. Can I change the test and make it smaller eg `shift_reg_576` ?

lnsharma:
> @lnsharma The `shift_reg_1024` test doesn't fit into the 24x24 device. Can I change the test and make it smaller eg `shift_reg_576` ?

Sure, please go ahead and make the changes.

PR #2121, opened 2021-04-06
QuickLogic qlf_k4n8 repacker

mkurc-ant:
The repacking is a step during which so called "operating cells" used by the VPR packer are converted into so called "phsyical cells" used by the physical device architecture. Both cell sets are defined within different models of relevant pb_types in the VPR architecture. The step involves revoking packing and routing of each cluster and then rebuilding it using physical cells only. The conversion is compliant with the `repack` step as deifined in the OpenFPGA project (https://openfpga.readthedocs.io/en/master/manual/openfpga_shell/openfpga_commands/fpga_bitstream_commands/#repack).

Repacking is added to the CMake flow through a new general mechanism of packed netlist patching. One may define an external utility run between VPR placement and VPR routing stages. See `common/cmake/device.cmake` for details. A similar step is added to the flow of the installed toolchain.

The set of rules for the repacker that define operating and physical cell correspondences are stored in a JSON file under the `third_party/qlf_symbiflow_plugins` submodule.

There is currently no format check whether the repacker operates correctly - such a test leveraging Yosys' equivalence checking is being worked on. Besides that there are more low-level tests for parts of the repacker code available under `utils/qlf_k4n8/repacker/tests`.

### DIFF
--- a/quicklogic/common/toolchain_wrappers/vpr_common
+++ b/quicklogic/common/toolchain_wrappers/vpr_common
@@ -79,7 +79,8 @@ function parse_args {
 
      export ARCH_DIR=`realpath ${MYPATH}/../share/symbiflow/arch/${DEVICE_1}_${DEVICE_1}`
      export ARCH_DEF=${ARCH_DIR}/arch_${DEVICE_1}_${DEVICE_1}.xml
-     export RR_GRAPH=${ARCH_DIR}/rr_graph_${DEVICE_1}_${DEVICE_1}.rr_graph.real.bin
+#     export RR_GRAPH=${ARCH_DIR}/rr_graph_${DEVICE_1}_${DEVICE_1}.rr_graph.real.bin
+     export RR_GRAPH=${ARCH_DIR}/${DEVICE_1}.rr_graph.bin
      export PLACE_DELAY=${ARCH_DIR}/rr_graph_${DEVICE_1}_${DEVICE_1}.place_delay.bin
      export ROUTE_DELAY=${ARCH_DIR}/rr_graph_${DEVICE_1}_${DEVICE_1}.lookahead.bin
 
--- a/quicklogic/qlf_k4n8/utils/repacker/repack.py
+++ b/quicklogic/qlf_k4n8/utils/repacker/repack.py
@@ -1096,6 +1096,12 @@ def main():
     net_xml = ET.parse(args.net_in, ET.XMLParser(remove_blank_text=True))
     packed_netlist = PackedNetlist.from_etree(net_xml.getroot())
 
+    # Count blocks
+    total_blocks = 0
+    for clb_block in packed_netlist.blocks.values():
+        total_blocks += clb_block.count_leafs()
+    logging.debug(" {} leaf blocks".format(total_blocks))
+
     init_time = time.perf_counter() - init_time
     repack_time = time.perf_counter()
 
@@ -1415,6 +1421,12 @@ def main():
         with open(fname, "w") as fp:
             fp.writelines(placement)
 
+    # Count blocks
+    total_blocks = 0
+    for clb_block in packed_netlist.blocks.values():
+        total_blocks += clb_block.count_leafs()
+
+    # Print statistics
     logging.info("Finished.")
 
     logging.info("")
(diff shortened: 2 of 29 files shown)


quicklogic/common/toolchain_wrappers/vpr_common
```diff
@@ -79,7 +79,8 @@ function parse_args {
 
      export ARCH_DIR=`realpath ${MYPATH}/../share/symbiflow/arch/${DEVICE_1}_${DEVICE_1}`
      export ARCH_DEF=${ARCH_DIR}/arch_${DEVICE_1}_${DEVICE_1}.xml
-     export RR_GRAPH=${ARCH_DIR}/rr_graph_${DEVICE_1}_${DEVICE_1}.rr_graph.real.bin
+#     export RR_GRAPH=${ARCH_DIR}/rr_graph_${DEVICE_1}_${DEVICE_1}.rr_graph.real.bin
```
acomodi, line 82: Why is this commented out?

quicklogic/qlf_k4n8/utils/repacker/repack.py
```diff
@@ -0,0 +1,1394 @@
...
+            # Update port map
+            for src_pin, dst_pin in zip(src_pins, dst_pins):
+                port_map[src_pin] = dst_pin
+
+        rule.port_map = port_map
+
+        # DEBUG
+#        logging.debug(" ", rule.src, "->", rule.dst)
+#        for k, v in rule.port_map.items():
+#            logging.debug("  ", k, "->", v)
```
acomodi, line 821: Maybe remove these or get them under an `if DEBUG` block with a global DEBUG variable to set it on or off.

quicklogic/qlf_k4n8/utils/repacker/repack.py
```diff
@@ -0,0 +1,1394 @@
...
+            logging.debug("   " + str(src_block))
+
+            # Find the original pb_type
+            src_path = src_block.get_path(with_indices=False)
+            src_pbtype = clb_pbtype.find(src_path)
+            assert src_pbtype is not None, src_path
+
+            #            # Get the source BLIF model
+            #            assert src_pbtype.blif_model is not None
+            #            src_blif_model = src_pbtype.blif_model
```
acomodi, line 1182: Is it correct to comment out this assertion?

acomodi (commented):
What I do not fully understand about the repacker is why it is required in the flow and why we require to have the "physical cells" instead of the "operating" ones?
As far as I understand, by looking also at the qlf_4k8n arch, is that there are these `physical` modes that are used in the repacker to be used in place of other modes, so, a broader question is, why not expressing the arch already using these `physical` modes? The main concern here is that adding yet another step outside of the control of VPR further increases complexity and maintainability.

I think that a thoroughly in code comment on why this step is intended and a slightly more detailed description of what the repacker actually does under the hood might be good to have on top of the repacker code.

mkurc-ant:
@acomodi I am aware of the possibility of expressing the arch using only the physical cells plus introducing Yosys techmaps that would map all cells of a design to them.

The approach with the repacker in the SymbiFlow flow strictly follows the flow in the OpenFPGA project which the `qlf_k4n8` device is based on. There VPR operates on a set of logical cells (called "operating") which are then translated to their physical counterparts according to a given set of rules. The bitstream then is generated using the design expressed using physical cells only. Apart from all that the OpenFPGA project is responsible for generating design files for actual FPGA chips.

Replicating the OpenFPGA flow in SymbiFlow has the benefit of keeping common architecture and cell models between these two projects. This eases development of the device as the source architecture is modified in a single place. There is no need to continuously manually re-work VPR architecture for SymbiFlow each time a change is made to the OpenFPGA  which would be required if they were significantly different.

Besides that the repacking approach allows for more low-level control of how the cells are mapped. For example for a `clb` you can have separate modes for a carry chain and a shift register chain. During VPR packing they have to use different cell types because otherwise VPR cannot pack correctly. After repacking both type of chains are represented by physical cells of the same type, just connected differently.

Ultimately the VPR arch used in SymbiFlow is a processed version of the OpenFPGA one but the processing is limited to syntax adjustments between VPR versions used and FASM annotation injection.

mkurc-ant:
@acomodi I've added documentation for the repacker.

acomodi:
> I've added documentation for the repacker.

@mkurc-ant Great, thanks, that helps to better understand the reasons and workings of the repacker.

I think that, as mentioned in the PR's description, some kind of equivalence checking tests is definitely required to verify the correctness of the output repacked netlist.

mkurc-ant:
@acomodi I agree. I have a WIP equivalence checking mechanism. Though it requires some more work. It is not included here not to stall this PR.

PR #2126, opened 2021-04-19
Updating quicklogic yosys plugin .so name to ql-qlf

lnsharma:


acomodi:
@lnsharma I think the yosys plugins need to be bumped to the latest version in the environment.yml file:

E.g.:
```
litex-hub::symbiflow-yosys-plugins=1.0.0_7_305_g9e969e2=20210415_092356
```

mkurc-ant:
xc7 install CI hanged, the rest is green. Since this PR does not touch xc7 install related code I'm merging it to not stall it further

PR #2129, opened 2021-04-20
cmake: fix rr graph targets to generate the real rr graph

acomodi:

This PR solves the issue for which, if no external RR graph was provided and the rr_graph patching was required (this is true for xc7 devices), the real RR graph was never generated, and the patched one was directly used. This resulted in the installation package to lack the real RR graph, with a consequent failure of the install tests.

The fix in this PR enables the generation of the real RR graph in case the patching is required, otherwise, the externally provided rr graph (used only by the QL devices) is used as final RR graph target.

### DIFF
--- a/common/cmake/devices.cmake
+++ b/common/cmake/devices.cmake
@@ -721,76 +721,25 @@ function(DEFINE_DEVICE)
 
     endif()
 
-    # Convert patched rr_graph.xml to rr_graph.bin if necessary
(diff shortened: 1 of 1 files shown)


mkurc-ant (approved):
Can you add a brief description to the PR what was the problem and how you fixed it? If the CI for xc7 install passes then LGTM.

acomodi:
> Can you add a brief description to the PR what was the problem and how you fixed it? If the CI for xc7 install passes then LGTM.

Sure, done

PR #2130, opened 2021-04-21
Adding lib file generation support

lnsharma:
This generates liberty file for QuickLogic eFPGA device - qlf_k4n8.


kgugala:
there seems to be Python code formatting issue

quicklogic/qlf_k4n8/utils/create_lib.py
```diff
@@ -0,0 +1,416 @@
...
+
+    header_templ_file = os.path.join(
+        curr_dir, '../devices/umc22/lib_header_template.txt'
+    )
+    in_str = open(header_templ_file, 'r').read()
+    curr_str = in_str.replace("@lib_name@", lib_name)
+    today = date.today()
+    curr_date = today.strftime("%B %d, %Y")
+    str_rep_date = curr_str.replace("@Date@", curr_date)
+    str1 = str_rep_date.replace("@cell_name@", cell_name)
```
mkurc-ant, line 112: Instead of doing `str.replace` with custom placeholder strings you may process the whole text with `str.format`. It already has a mechanism for what you want and is easily extendable. This is just a suggestion.

quicklogic/qlf_k4n8/utils/create_lib.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+def form_out_reset_timing_group(reset_name, timing_sense, timing_type):
+    '''
+    Form timing group for output pin when related pin is reset
+    '''
+    cell_fall_val = "4.6000"
+    fall_tran_val = "0.2000"
+    curly_braces_open = "{"
+    curly_braces_close = "}"
```
mkurc-ant, line 231: You don't have to pass that as a variable to `format`. To insert a literal brace you just need to double it. `{{` will end up as `{`. Same for the closing one.

quicklogic/qlf_k4n8/utils/create_lib.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+def form_out_timing_group(clk_name, timing_type):
+    '''
+    Form timing group for output pin in a Pin group in a library file
+    '''
+    cell_rise_val = "2.59"
+    cell_fall_val = "2.59"
+    rise_tran_val = "0.85"
+    fall_tran_val = "0.85"
```
mkurc-ant, line 259: There are lot of hard-coded values throughout the script. If they are hard constants common to all devices then this is fine. But if they are device-specific I'd add a mechanism to read them eg. from a JSON file (like the lib header template is). We can leave that as is if it is not critical for now.

quicklogic/qlf_k4n8/utils/create_lib.py
```diff
@@ -0,0 +1,416 @@
...
+    )
+    parser.add_argument(
+        "--xml",
+        "-x",
+        "-X",
+        type=str,
+        required=True,
+        help='Input interface-mapping xml file'
+    )
+
```
mkurc-ant, line 75: The files `dedicated_pin_lib_data.txt` and `lib_header_template.txt` have hard-coded names and paths relative to this script. I presume that they are specific to a device. It may be better to add eg. an option to specify a path to the device folder where the files are. This would make the script more generic and usable for future devices to come.

mkurc-ant (approved):
I have a couple of comments how to improve the Python script(s). But I'm ok with how it is done right now. Those improvements may be added along with a support for more qlf devices.

kgugala:
I pinged the CI to run the tests on this PR

quicklogic/qlf_k4n8/utils/create_lib.py
lnsharma: I cannot use str.format in this case because curly braces are even used in the text and that conflicts with keywords specification for format.

quicklogic/qlf_k4n8/utils/create_lib.py
lnsharma: Incorporated this in the last checkin

quicklogic/qlf_k4n8/utils/create_lib.py
lnsharma: Incorporated this in the last checkin. Created a JSON file to pick these values.

quicklogic/qlf_k4n8/utils/create_lib.py
lnsharma: Incorporated this in the last checkin. Added a command line option to get template data folder path.

kgugala:
7 series install test hanged, since this PR does not touch this part I'm merging it to not stall it

PR #2136, opened 2021-05-14
bump yosys and yosys-plugins

acomodi:


umarcor:
As discussed in a meeting, this can be closed. In this repository, those tools are pinned through the conda lock file, so versions don't need to be specified in the environment file.

PR #2137, opened 2021-05-17
QuickLogic: upstreaming toolchain changes

acomodi:
This PR adds the following changes:

- toolchain wrappers able to generate bitstreams and 4byte data
- usage of the external qlf-fasm library
- bump of qlfpga-symbiflow-plugins with correct architecture data
- changes and addition of qlf library data
- fixes to the repacker

Note: most of the changes happen to be library files additions


quicklogic/common/cmake/quicklogic_install.cmake
```diff
@@ -63,10 +63,19 @@ function(DEFINE_QL_TOOLCHAIN_TARGET)
...
+	  DESTINATION share/symbiflow/tests/counter_16bit
+	  PERMISSIONS WORLD_READ OWNER_WRITE OWNER_READ GROUP_READ)
   install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/${FAMILY}/tests/counter_16bit/counter_16bit.sdc
           DESTINATION share/symbiflow/tests/counter_16bit
           PERMISSIONS WORLD_READ OWNER_WRITE OWNER_READ GROUP_READ)
-  
+  install(FILES ${symbiflow-arch-defs_SOURCE_DIR}/quicklogic/${FAMILY}/tests/counter_16bit/clk_constraints.json
+          DESTINATION share/symbiflow/tests/counter_16bit
+          PERMISSIONS WORLD_READ OWNER_WRITE OWNER_READ GROUP_READ)
+
```
mkurc-ant, line 78: I'm sure that installation of the `counter_16bit` example for the installed toolchain should not be done here.

quicklogic/common/toolchain_wrappers/symbiflow_place
```diff
@@ -45,6 +45,12 @@ else
 
 fi
 
+# Generate lib file
+if [ -s $PART ]; then
```
mkurc-ant, line 49: Hmmm, Why this is generated during placement? Shouldn't there be a separate option in `ql_symbiflow` to do it?

mkurc-ant (commented):
I've added some comments about the changes in context of them being in upstream.

We could think of moving those long .lib files to qlffpga_symbiflow_plugins. We could move there even pinmap files - there are kind of long too.

quicklogic/common/toolchain_wrappers/symbiflow_place
acomodi: I didn't see an option to generate these lib file in `ql_symbiflow`.

@kkumar23, do you know whether this should be here or is it better to move it in `ql_symbiflow`?

quicklogic/common/cmake/quicklogic_install.cmake
acomodi: Hmm, yeah... I think we might need to add a function to install examples tests and use that one.

kkumar23:
Its here :
/quicklogic/common/toolchain_wrappers/symbiflow_generate_libfile<https://github.com/QuickLogic-Corp/symbiflow-arch-defs/blob/master/quicklogic/common/toolchain_wrappers/symbiflow_generate_libfile>

acomodi:
@kkumar23, Right, but the call to `generate_symbiflow_lib` is correct to be included in `symbiflow_place`? I might be missing something, but I think it might be better to have it explicitly generated in `ql_symbiflow`

kkumar23:
Yes it can be done. We added this for easy reference.

quicklogic/common/cmake/quicklogic_install.cmake
kkumar23: true. This we added during the release, so that customer can have alteast one test design to start with.

acomodi: Ok, I think that, for the time being, we can leave the installation of the example out of this PR and add it properly once everything lands upstream.

quicklogic/common/toolchain_wrappers/symbiflow_place
acomodi: Given that the lib data files are being removed and will be added to the external data repository, this change is not part of this PR anymore. As soon as the plugin will contain `lib` and `csv` files the lib data generation will be implemented once again.

quicklogic/CMakeLists.txt
```diff
@@ -1 +1,3 @@
+set(QLF_FPGA_DATABASE_DIR ${symbiflow-arch-defs_SOURCE_DIR}/third_party/qlfpga-symbiflow-plugins)
```
mkurc-ant, line 1: It looks like this is never used. Can we remove?

mkurc-ant (approved):
One comment about a CMake file. Otherwise LGTM

quicklogic/CMakeLists.txt
acomodi: This is actually used when defining the arch and device, also to access the fasm database

PR #2138, opened 2021-05-17
QuickLogic: upstream post synth analysis fixes

acomodi:
This is based on top of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2137, and it will need a rebase after #2137 is merged.

The PR adds fixes to allow a correct post-P&R analysis


mkurc-ant (approved):
Ok, looks good.

PR #2140, opened 2021-05-17
QuickLogic: upstream repacker fixes

acomodi:
This is based on top of #2137, and it will need a rebase after #2137 is merged.

The PR fixes an issue with the repacker around the LUT fasm outputs. It also adds a test to verify that the fix is working as expected.


mkurc-ant (approved):
LGTM

PR #2139, opened 2021-05-17
QuickLogic: upstream new testcases

acomodi:
This PR is based on top of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2137 and it requires a rebase once #2137 is merged.

This PR adds an external submodule that holds all the quicklogic benchmarks. In addition, it increases the test coverage.

PR #2141, opened 2021-05-17
QuickLogic: upstream new qlf k6n10 device

acomodi:
This PR is based on top of #2137 and #2139 it requires a rebase once they are merged.

This PR adds the new qlf k6n10 device


mithro:
We should not be using `git lfs` in the `symbiflow-arch-defs` repository.

acomodi:
@mithro `lfs` is only used in the QL flow to get the RR graph, which is externally generated and stored with `git lfs`. It will be used only during the `ql.sh` script run to pull it.

A different solution might be to upload those as artifacts and pull them instead of using `git lfs`.

PR #2142, opened 2021-05-18
QuickLogic: upstream multiple corners

acomodi:
This PR builds on top of https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2139.

It adds multiple corner devices based on the same architecture.


mkurc-ant (approved):
LGTM

PR #2149, opened 2021-06-15
[EOS-S3] Changes to common CMake and Python scripts

mkurc-ant:
This PR incorporates all changes needed for common CMake scripts as well as for common utility Python scripts required for QuickLogic EOS-S3 eFPGA support.

### DIFF
--- a/common/cmake/v2x.cmake
+++ b/common/cmake/v2x.cmake
@@ -39,8 +39,6 @@ function(V2X)
   get_target_property_required(YOSYS env YOSYS)
   list(APPEND DEPENDS_LIST ${YOSYS})
 
-  set(V2X_DIR ${symbiflow-arch-defs_SOURCE_DIR}/third_party/python-symbiflow-v2x)
-
   set(REAL_SOURCE_LIST "")
   foreach(SRC ${V2X_SRCS})
     if(NOT "${SRC}" MATCHES "\\.sim\\.v$")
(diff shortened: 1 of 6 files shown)


acomodi:
Might be worth squashing all commits

mkurc-ant:
@acomodi I did some squashing / cleanup but wanted to preserve as much of the history as possible

acomodi:
@mkurc-ant just a nitpick actually, but I think that it would be best to just squash everything in one/two commits.

Take this for example: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2149/commits/97f547f83ad84659a46b16455b268e7b71d2b65c. It just removes one line and does specify that something else is done in the commit description.

I think that it would make more sense to squash everything in a couple of commits, and maintaining the co-authors, as well as the various commit messages.

mkurc-ant:
@acomodi Ok, Now there is only one comment

utils/vpr_io_place.py
```diff
@@ -22,6 +22,7 @@ def __init__(self):
         self.inputs = set()
         self.outputs = set()
         self.net_to_block = None
+        self.block_to_inst = {}
```
acomodi, line 25: Just saw this, probably it is redundant as it is already defined below.

mkurc-ant: Right... fixed.

acomodi (approved):
LGTM

PR #2151, opened 2021-06-16
QuickLogic: Re-enable CLB input pin equivalence

mkurc-ant:
This PR re-enabled equivalent input pins of `qlf_k4n8` CLBs and re-enables/adds previously failing tests.
It also bumps VTR version so that the one with a `genfasm` bug fixed is used.


acomodi (approved):
LGTM

PR #2152, opened 2021-06-16
QuickLogic: Update qlf-fasm and yosys-symbiflow-plugins

mkurc-ant:
This PR updates revisions of both mentioned packages which is required for up-to-date `qlf_k4n8` eFPGA support. The updated `qlf_fasm` provides support for default bitstream which is then mutated by a design FASM file.


acomodi (approved):
LGTM

PR #2169, opened 2021-06-22
quicklogic: pp3: add utils scripts

acomodi:
This PR adds all the utils script necessary to build the arch and RR graph devices, as well as generating IO and placement constraints when running VPR.


quicklogic/common/utils/testarch_graph.py
```diff
@@ -0,0 +1,404 @@
+#!/usr/bin/env python3
```
mkurc-ant, line 1: Hmmm, I don't recall this script being needed for QuickLogic PP3 architecture. If it is not used anywhere and is not an important utility we might remove it.

mkurc-ant (changes requested):
Most of this scripts are PP3 specific - they haven't been re-used for another architecture so far. If so then we may consider moving them to `pp3/utils` instead of keeping them in `common`.

quicklogic/common/utils/testarch_graph.py
lpawelcz: I don't see it being used anywhere in the codebase so I removed it

lpawelcz:
This PR should be merged only after https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2170

PR #2167, opened 2021-06-22
quicklogic: add ql common cmake for EOS and PP3

acomodi:
This PR adds the base CMake files to add the QL PP3 devices (eos and pp3e).

It also sorts in alphabetical order some binary dependencies and requirements to make them more readable and robust to merge conflicts

PR #2170, opened 2021-06-22
quicklogic: pp3: added primitives

acomodi:
This PR adds all the primitives XMLs required for the PP3 devices.


quicklogic/pp3/primitives/fpga_interconnect/fpga_interconnect.sim.v
```diff
@@ -0,0 +1,16 @@
+`timescale 1ns/10ps
+(* whitebox *)
+module fpga_interconnect(
```
mkurc-ant, line 3: This one actually does not belong to the architecture. This is a cell present in post-pnr verilog netlist from VPR. I'm not sure whether we want it here or eg. in `yosys/techmap/cells_sim.v`. I guess the only reason why it ended up here is to be included in the auto-generated `cells_sim.v` for the arch.

quicklogic/pp3/primitives/ramfifo/ram.sim.v
```diff
@@ -0,0 +1,248 @@
+`timescale 10 ps /1 ps
```
mkurc-ant, line 1: This file could use indentation fixing and/or reformatting with Verible

quicklogic/pp3/primitives/ramfifo/ramfifo.sim.v
```diff
@@ -0,0 +1,1128 @@
+`timescale 1 ns /10 ps
```
mkurc-ant, line 1: This one could make use of reformatting too. It may be worth to compare models included here with those in `symbiflow-yosys-plugins`. I think that they may be the same and if so then there is no reason to have them here.

mkurc-ant (changes requested):
A few files require attention but otherwise LGTM

quicklogic/pp3/primitives/fpga_interconnect/fpga_interconnect.sim.v
acomodi: Yeah, the `cells_sim.v` file is generated by grouping and merging the various `sim.v` files from the various primitives. There is actually no static cells_sim in the techmap, so I think to conform to the way the `cells_sim.v` is generated, this might just stay here.

mkurc-ant: Ok. If you can then maybe add a comment that this is only a cell emitted by VPR and not a part of the PP3 arch.

quicklogic/pp3/primitives/ramfifo/ramfifo.sim.v
acomodi: Yeah, I have checked and the three sim files added in ramfifo are merged all together in the `pp3_brams_sim.v` file in the yosys-plugins. I have removed ramfifo.

quicklogic/pp3/primitives/fpga_interconnect/fpga_interconnect.sim.v
acomodi: Done

lpawelcz:
This PR should me merged after https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2167

acomodi:
This PR only adds primitives to the quicklogic subdirectory and does not modify at all any common infrastructure that may affect xc7, merging.

PR #2175, opened 2021-06-24
kokoro: update to jdk 11

acomodi:
Recent changes to rapidwright seem to require a newer version of jdk

PR #2177, opened 2021-06-25
[WIP] Use OpenJDK 11

mkurc-ant:
Switched to OpenJDK 11 to support latest RapidWright


acomodi:
FYI, there is a PR that fixes the openJDK version issue here: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2175

mkurc-ant:
@acomodi Great, so this one is not needed. Closing

PR #2183, opened 2021-06-29
QuickLogic: Unify k4n8 and k6n10 utilities

mkurc-ant:
This PR unifies common utilities for the `qlf_k4n8` (already supported) and `qlf_k6n10` (to be supported in upstream).
It also fixes a bug with interpretation of device family install / no-install lists.


acomodi (approved):
LGTM, merging

PR #2211, opened 2021-07-19
Use gcpbucket instead of gitlfs

lpawelcz:
This PR deletes `qlfpga-symbiflow-plugins` submodule and adds new cmake targets that will fetch required files for `qlf_k4n8` from GPC.


acomodi (approved):
LGTM

PR #2228, opened 2021-07-29
docs: clarify "getting started" section

lpawelcz:
This PR renames the `Getting started` section to clarify that it is a part of toolchain description. It also emphasis this in the note and points user directly to user documentation instead of `symbiflow-examples` repository.


docs/source/getting-started.rst
```diff
@@ -1,10 +1,10 @@
-===============
-Getting Started
-===============
+====================================================
+Getting Started with SymbiFlow Toolchain development
+====================================================
 
 .. note::
```
mithro, line 5: Might be worth changing this to a warning?

mithro (approved):
One minor comment, otherwise LGTM!

docs/source/getting-started.rst
lpawelcz: That's a good idea - updated

mithro (approved):
LGTM!

PR #2238, opened 2021-08-02
[PP3] Update Yosys plugins

mkurc-ant:
This PR updates the `yosys-symbiflow-plugins` conda package and renames ports of some QuickLogic PP3 cells to align with the change.


acomodi (approved):
LGTM

PR #2242, opened 2021-08-11
Update yosys and VTR

acomodi:
This PR updates yosys and VTR to a more recent version, and it also adds a few fixes required to have update the packages.


xc/xc7/yosys/synth.tcl
```diff
@@ -14,9 +14,14 @@ source [file join [file normalize [info script]] .. utils.tcl]
 # -flatten is used to ensure that the output eblif has only one module.
 # Some of symbiflow expects eblifs with only one module.
 #
+# To solve the carry chain congestion at the output, the synthesis step
```
mithro, line 17: Can you log a bug about this issue?

mithro (commented):
Any info on the auto-updater for this?

xc/xc7/yosys/synth.tcl
acomodi: Sure, as soon as I can get a small and reproducible test I'll open an issue

syed-ahmed:
@acomodi Could you please share a little more about what kind of trouble abc9 is causing when it gets called multiple times in the flow?

acomodi:
@syed-ahmed Sure, I'll also open an issue upstream, but for context, I have encountered this error when the design included or inferred usage of FDPE cells.

In particular, the error reported is the following:

```
3.44. Executing ABC9 pass.

3.44.1. Executing ABC9_OPS pass (helper functions for ABC9).

3.44.2. Executing ABC9_OPS pass (helper functions for ABC9).

3.44.3. Executing SCC pass (detecting logic loops).
Found 0 SCCs in module top.
Found 0 SCCs.

3.44.4. Executing ABC9_OPS pass (helper functions for ABC9).
ERROR: Attempted to add new module named '\FDPE_$abc9_byp', but a module by that name already exists
Makefile:2: recipe for target 'all' failed
make: *** [all] Error 1
```

[Here](https://github.com/SymbiFlow/symbiflow-arch-defs/files/7019727/abc-fail.tar.gz) is a packaged tar containing a minimal case to reproduce the error with yosys @ 8bf9cb407.

PR #2244, opened 2021-08-17
quicklogic: cmake: fix jlink and openocd cmake deps

acomodi:


mkurc-ant:
Surprisingly Kokoro fails for the `picosoc` design on Xilinx xc7a200t on a DRC check:
```
ERROR: [DRC RTRES-2] Global clock nets are using local routing resources: Global clock nets are using local routing resources. 1 net(s) have at least one node with COST_CODE_NAME equal to BOUNCEACROSS or INTENT_CODE_NAME equal to NODE_DOUBLE, NODE_HLONG, NODE_HQUAD, NODE_SINGLE, NODE_VLONG, or NODE_VQUAD. This situation occurs when a global clock net must enter a clock region where all global clock routing resources are occupied. Check the report from report_clock_utilization to determine which clock regions are traversed by the problem net and use floorplanning or other physical constraints to ensure a maximum of 12 global clock nets occupy each clock region. The problem net(s) are soc.simpleuart.clk.
```

PR #2245, opened 2021-08-18
quicklogic: add initial PP3 support

lpawelcz:
This PR introduces basic support for QuickLogic PP3 devices.
It adds a third party repository with architecture data in a similar way as it was done for PP3E.
Since there is no timing data available for PP3 at this moment, we are using PP3E data
which will be changed in the nearest future for proper PP3 files.
`data_import.py` utility for QL had to be modified in order to ignore empty switchboxes
in architecture techfile that were causing failures while importing the data.
Additionally, basic test designs were enabled for the `jimbob4` boards and the test cases
for routing were added.


lpawelcz:
This PR is moved to https://github.com/QuickLogic-Corp/symbiflow-arch-defs/pull/467
I'm closing this one.

PR #2265, opened 2021-09-10
quicklogic: pin quicklogic_fasm python library

acomodi:


mithro (commented):
We should probably be using something like `pip-lock` like we are doing for `conda-lock`?

acomodi:
This actually will be pinned in the conda-lock file, which contains all the pip packages locked already (if the package is pinned in the requirements file)

PR #2289, opened 2021-10-01
gh: conda lock: reduce frequency to once a week

acomodi:

### DIFF
--- a/.github/workflows/update_conda_lock.yml
+++ b/.github/workflows/update_conda_lock.yml
@@ -3,7 +3,7 @@ name: update_conda_lock
 on:
   push:
   schedule:
-    - cron: '0 3 * * *'
+    - cron: '0 3 * * 0'
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


mithro:
@acomodi - I think we want to figure out how we make sure that these conda-lock PRs are getting automatically merged if CI passes, or escalated to a human if the CI fails?

acomodi:
@mithro I agree, I believe it is more convenient to have these PRs issued less frequently, and once automatic CI and merging is enabled, we can just re-enable once-per-day PRs

acomodi:
XC7 Vendor Tools failed for a similar reason as for https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2244#issuecomment-933230461 (opened issue [here](https://github.com/SymbiFlow/symbiflow-arch-defs/issues/2297)). This PR just changes an unrelated part, which does not affect regression testing CI. Merging

PR #2290, opened 2021-10-01
Fix SSL certificate issues with kitware.com causing Kokoro CI failures

mithro:
Fixes the issue appearing as;
```
========================================
Host adding PPAs
----------------------------------------
gpg: no valid OpenPGP data found.


[ID: 7946271] Build finished after 82 secs, exit value: 2
```

Which is actually an issue caused by the kitware.com SSL certificate not being understood by the ancient Ubuntu running on the kokoro host.

Info was discovered by @sdamghan at https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1864

### DIFF
--- a/.github/kokoro/steps/hostsetup.sh
+++ b/.github/kokoro/steps/hostsetup.sh
@@ -11,9 +11,29 @@ echo "----------------------------------------"
 
 echo
 echo "========================================"
+echo "Update the CA certificates"
+echo "----------------------------------------"
+sudo apt-get install -y ca-certificates
+sudo update-ca-certificates
+echo
+
+echo
+echo "========================================"
+echo "Host not verifying certificate name matches server name"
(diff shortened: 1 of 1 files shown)


mithro:
FYI - The "best solution" seems to be to remove the old `/usr/share/ca-certificates/mozilla/DST_Root_CA_X3.crt` letsencrypt certificate that has expired. This is fine because the new non-expired `/usr/share/ca-certificates/mozilla/ISRG_Root_X1.crt` certificate is still available. See the post at https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/ for more explaination.

mithro:
This seems to have fixed Kokoro.

mithro:
FYI - @acomodi / @kgugala

Issue #2248, opened 2021-08-26
MMCM primitive generate incorect clock frequencies in hardware


WhiteNinjaZ:
@acomodi @mkurc-ant @nelsobe I have been running several designs in hardware using the new MMCME2_ADV primitives and have noticed that the timing has been off. I ran a few LED testes with both the PLL and MMCM at what should be 25MHz for both and got the correct results for the PLL but it looks like the MMCM is generating a clock at about 3kHz instead of 25MHz. This is how I instanced both modules:
```
MMCME2_ADV #(
        .CLKFBOUT_MULT_F(9),
        .CLKIN1_PERIOD(10.0),
        .CLKOUT0_DIVIDE_F(36.0),
        .CLKOUT1_DIVIDE(1),  // included for Veriltator (TMDS clock)
        .DIVCLK_DIVIDE(1)
    ) MMCME2_ADV_inst (
        .CLKIN1(clk),
        .RST(rst),
        .CLKOUT0(clk_pix_unbuf),
        .LOCKED(locked),
        .CLKFBOUT(clk_fb),
        .CLKFBIN(clk_fb),
        /* verilator lint_off PINCONNECTEMPTY */
        .CLKOUT0B(),
        .CLKOUT1(),
        .CLKOUT1B(),
        .CLKOUT2(),
        .CLKOUT2B(),
        .CLKOUT3(),
        .CLKOUT3B(),
        .CLKOUT4(),
        .CLKOUT5(),
        .CLKOUT6(),
        .CLKFBOUTB(),
        .PWRDWN()
        /* verilator lint_on PINCONNECTEMPTY */
    );
```
```
PLLE2_ADV #(
        .CLKFBOUT_MULT(9),
        .CLKIN1_PERIOD(10.0),
        .CLKOUT0_DIVIDE(36),
        .CLKOUT1_DIVIDE(1),  // included for Veriltator (TMDS clock)
        .DIVCLK_DIVIDE(1)
    ) PLLE2_ADV_inst (
        .CLKIN1(clk),
        .RST(rst),
        .CLKOUT0(clk_pix_unbuf),
        .LOCKED(locked),
        .CLKFBOUT(clk_fb),
        .CLKFBIN(clk_fb),
        /* verilator lint_off PINCONNECTEMPTY */
        .CLKOUT1(),
        .CLKOUT2(),
        .CLKOUT3(),
        .CLKOUT4(),
        .CLKOUT5()
[...]

PR #2293, opened 2021-10-03
Added new mmcm test

WhiteNinjaZ:
This PR is meant to test the problem discussed in [issue #2248](https://github.com/SymbiFlow/symbiflow-arch-defs/issues/2248) and tests the timing for the MMCM. The design uses an MMCM, and PLL to drive a 25KHz clock and is meant for a basys3 board. If the design works properly leds 0-2 on the basys should blink at the same rate.

### DIFF
--- a/xc/xc7/tests/mmcm/CMakeLists.txt
+++ b/xc/xc7/tests/mmcm/CMakeLists.txt
@@ -60,6 +60,10 @@ add_file_target(FILE mmcm_int_basys3_bottom.v SCANNER_TYPE verilog)
 add_file_target(FILE mmcm_buf_basys3_bottom.v SCANNER_TYPE verilog)
 add_file_target(FILE mmcm_ext_basys3_bottom.v SCANNER_TYPE verilog)
 add_file_target(FILE mmcm_int_frac_basys3_bottom.v SCANNER_TYPE verilog)
+add_file_target(FILE mmcm_int_frac_basys3_bottom.v SCANNER_TYPE verilog)
+
+add_file_target(FILE clock_tester.v SCANNER_TYPE verilog)
+add_file_target(FILE frequency_test.v SCANNER_TYPE verilog)
 
 add_fpga_target(
   NAME mmcm_packing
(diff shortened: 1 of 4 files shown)


nelsobe:
@WhiteNinjaZ So, if i understand, you have added this test to the symbiflow-arch-defs flow to make it look like the other tests and so @mkurc-ant and others can better see the results and try to see what is wrong with the designs?  

And, once @mkurc-ant  or someone else approves the modified workflow how long will it take for the run to be done (are we talking minutes or is it a nightly sort of thing?)

mithro:
I assume he is following the diagram at https://docs.google.com/document/d/1ZQp8D_Jp1C2kmJR9WOYwis8U5N0xIUB0RFLnvyLlNH8/edit ?

xc/xc7/tests/mmcm/CMakeLists.txt
```diff
@@ -60,6 +60,10 @@ add_file_target(FILE mmcm_int_basys3_bottom.v SCANNER_TYPE verilog)
 add_file_target(FILE mmcm_buf_basys3_bottom.v SCANNER_TYPE verilog)
 add_file_target(FILE mmcm_ext_basys3_bottom.v SCANNER_TYPE verilog)
 add_file_target(FILE mmcm_int_frac_basys3_bottom.v SCANNER_TYPE verilog)
+add_file_target(FILE mmcm_int_frac_basys3_bottom.v SCANNER_TYPE verilog)
```
acomodi, line 63: This file is already added as a target. A double target addition will cause CMake to fail

WhiteNinjaZ:
@acomodi @mkurc-ant I removed the extra file target as requested but I am still getting an error that looks like this:
```
CMake Error at common/cmake/file_targets.cmake:101 (get_target_property):
  get_target_property() called with non-existent target
  "file_xc_xc7_tests_common_basys3_bottom_frequency.pcf".
  common/cmake/devices.cmake:1545 (append_file_dependency)
  xc/xc7/tests/mmcm/CMakeLists.txt:107 (add_fpga_target)
```
I have done my best to debug this on my own but I'll admit I am a little stumped.

WhiteNinjaZ:
@acomodi would this MMCM thing be something to move into prjxray then? What is the best way to implement the test you suggest?

acomodi:
@WhiteNinjaZ Yeah, we may add a test for MMCMs similar to the [PLL](https://github.com/SymbiFlow/prjxray/tree/master/minitests/plle2_adv) one in the prjxray minitests. The new test would probably have another makefile target to perform the `bit -> fasm` translation.

This test can be useful to check whether the `yosys -> vivado` flow is working properly. If `yosys (synthesis) + vivado (pnr)` does not work, but the full-vivado flow does, there might be a bug in synthesis.

PR #2325, opened 2021-11-22
gtp: add full LiteSATA example with workarounds

acomodi:

This generates a bitstream that can read/write on an M.2 SSD through SATA, using the following [sata adapter](https://github.com/antmicro/fmc-sata-adapter) and a `nexys_video` board


mithro:
Super cool!

Can you make sure that the work arounds have GitHub issues?

xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -991,7 +991,7 @@ def main():
 
     idelayctrl_instances = place_constraints.get_used_instances("IDELAYCTRL")
 
-    assert len(idelayctrl_cmts) == len(
+    assert len(idelayctrl_cmts) <= len(
```
mithro, line 994: This doesn't feel right? Why is less than now allowed?

mithro:
Seems like there are some generic improvements to the LiteX output that are mixed in with the GTP/libsata addition?

xc/common/utils/prjxray_create_place_constraints.py
acomodi: Yeah, I think this was meant for some situations in which the number of IDELAYCTRLs in the design is less than the number of IDELAY cmts. Basically Vivado can replicate the IDELAYCTRLs if they are needed in different banks, even though there is only one of them explicitly instantiated in the design.

acomodi:
> Can you make sure that the work arounds have GitHub issues?

Sure, opened issues and linked them in the comments (https://github.com/SymbiFlow/symbiflow-arch-defs/issues/2328 https://github.com/SymbiFlow/symbiflow-arch-defs/issues/2327).

PR #2329, opened 2021-11-23
Fix conda packaging issue

acomodi:
This fixes the error described here: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2323#issuecomment-974268385.

This also adds the `merge-if-green` label, to auto merge all successful conda bot update PRs

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -31,6 +31,7 @@ dependencies:
   - libuuid
   - pkg-config
   - cython
+  - packaging=21.3
   - pip
   - pip:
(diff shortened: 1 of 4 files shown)


environment.yml
```diff
@@ -31,6 +31,7 @@ dependencies:
   - libuuid
   - pkg-config
   - cython
+  - packaging=21.3
```
mithro, line 34: Please add a comment about how/when this version lock can be removed.

PR #2330, opened 2021-11-24
Clarify how to make the demo bitstreams

MidsummerNight:
Current instructions can mislead people to `make` from the root of the `symbiflow-arch-defs` directory, whereas it should be run in `symbiflow-arch-defs/build`.


MidsummerNight:
Sorry, this one repeats #2315, closing now.

PR #2335, opened 2021-11-30
Change MMCME2 advanced COMPENSATION parameter to have a default of INTERNAL

nelsobe:
The Xilinx 7 Series Clocking Resources User Guide (UG 472) instructs users to NOT set a value for COMPENSATION since the Xilinx tools will choose one based on the design.  However, the guide offers no additional detail on how the Xilinx tools do that.  

As a result, standard user designs without a COMPENSATION parameter value are defaulting to ZHOLD (in the Symbiflow techmapper) which results in non-functional bitstreams. 

Testing has shown that setting the default COMPENSATION to INTERNAL or EXTERNAL generates functional bitstreams (by generating a bitstream with FASM feature Z_ZHOLD for the MMCM).

I note that the MMCM tests currently in the repo all set COMPENSATION to a specific value rather than leaving it at a default.  Thus, before this PR is ready for review the intention is to add an additional test for the case where no COMPENSATION value is specified.

Signed-off-by: Brent Nelson <nelson@ee.byu.edu>

### DIFF
--- a/xc/xc7/techmap/cells_map.v
+++ b/xc/xc7/techmap/cells_map.v
@@ -4790,6 +4790,16 @@ output [15:0] DO
 
   parameter BANDWIDTH = "OPTIMIZED";
   parameter STARTUP_WAIT = "FALSE";
+
+// Previously, the default COMPENSATION value was ZHOLD, resulting in non-functional 
+//   bitstreams when the feedback loop is closed on-chip.
+  // Setting it to INTERNAL as the default creates working bitstreams for that case.
+// This bug was not previously uncovered since the MMCM tests all explicitly
+//   specified a COMPENSATION value so the ZHOLD default was never used.
+// Setting it here in the techmapper means that existing code using on-chip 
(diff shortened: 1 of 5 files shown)


xc/xc7/techmap/cells_map.v
```diff
@@ -4790,7 +4790,7 @@ output [15:0] DO
 
   parameter BANDWIDTH = "OPTIMIZED";
   parameter STARTUP_WAIT = "FALSE";
-  parameter COMPENSATION = "ZHOLD";
```
mithro, line 4793: Probably a good idea to write some documentation about why the value of `COMPENSATION` should be set here...

nelsobe: @mithro Agreed it is a good idea.  Have added comments stating what is known about COMPENSATION to the code.

nelsobe:
@mkurc-ant What do you think - is it ready for merging?

mkurc-ant (approved):
Alright, this one looks good.

PR #2347, opened 2021-12-15
Fixes due to yosys changes

acomodi:
The conda lock update revealed a couple of issues with the most recent yosys.

This PR tries to fix these.

1. FD retargetting seems not to be working anymore. Now, the cell to retarget is add as a library module and it gets techmapped after synthesis
2. Yosys introduced `-r <topmodule>` option. This is now added when calling yosys.


acomodi:
@mithro For some reason I cannot get kokoro to run on this PR

PR #2350, opened 2021-12-16
xc: common: slicem: enforce correct intra-pb-type path for ADI1MUX input

acomodi:

This fixes an issue for which the `BDI` input of the `ADI1MUX` was coming directly from the common slice's DI input, which is a connection that is not present in the Series7 slicem. Instead, the `BDI` input must come from the `BDI1MUX` output, as clearly seen in the device view:
Screenshot from 2021-12-16 13-44-49

The previous bug was causing conflicting FASM features on the BDI1MUX which was getting configured to use two inputs at the same time (namely, `BI` and `DI`).

### DIFF
--- a/xc/common/primitives/slicem/slicem.pb_type.xml
+++ b/xc/common/primitives/slicem/slicem.pb_type.xml
@@ -400,6 +400,25 @@
         </metadata>
       </pb_type>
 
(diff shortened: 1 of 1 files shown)


tmichalak (approved):
This LGTM and it has been verified on hardware.

PR #2353, opened 2021-12-22
[WIP] EOS-S3 update [1/5] : fix xc7 dram block usage limits

lpawelcz:
This PR is first out of five PRs that are integrating and updating QuickLogic EOS-S3 support.

It is fixing the usage limits for `SYN-OUTPAD`, `SYN-INPAD` and `BLK-TL-SLICEM` in dram `ASSERT_USAGE` test cases.

Changes in following PRs reveal that those test cases were not working correctly because the usage for dram designs was exactly the same as current limits but previous limits were lower and tests were not failing on that.


lpawelcz:
Changes from this PR will be merged with https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2354

Issue #2363, opened 2022-01-12
Check-status CI job is failing?


mithro:
https://github.com/SymbiFlow/symbiflow-arch-defs/actions/runs/1689091015

```
Run ./.github/check-status
Building docker image
  Dockerfile for action: '/home/runner/work/symbiflow-arch-defs/symbiflow-arch-defs/./.github/check-status/Dockerfile'.
  /usr/bin/docker build -t 9916a7:d2bdedf2b8c554ce2aed8802e3176a61 -f "/home/runner/work/symbiflow-arch-defs/symbiflow-arch-defs/./.github/check-status/Dockerfile" "/home/runner/work/symbiflow-arch-defs/symbiflow-arch-defs/.github/check-status"
  Sending build context to Docker daemon  6.144kB
  
  Step 1/5 : FROM python:alpine
  alpine: Pulling from library/python
  59bf1c3509f3: Pulling fs layer
  07a400e93df3: Pulling fs layer
  d8c681bc2183: Pulling fs layer
  c7f1ca1171d9: Pulling fs layer
  dc9256fcea2e: Pulling fs layer
  c7f1ca1171d9: Waiting
  dc9256fcea2e: Waiting
  07a400e93df3: Verifying Checksum
  07a400e93df3: Download complete
  59bf1c3509f3: Verifying Checksum
  59bf1c3509f3: Download complete
  d8c681bc2183: Verifying Checksum
  d8c681bc2183: Download complete
  59bf1c3509f3: Pull complete
  c7f1ca1171d9: Verifying Checksum
  c7f1ca1171d9: Download complete
  dc9256fcea2e: Verifying Checksum
  dc9256fcea2e: Download complete
  07a400e93df3: Pull complete
  d8c681bc2183: Pull complete
  c7f1ca1171d9: Pull complete
  dc9256fcea2e: Pull complete
  Digest: sha256:dce56d40d885d2c8847aa2a278a29d50450c8e3d10f9d7ffeb2f38dcc1eb0ea4
  Status: Downloaded newer image for python:alpine
   ---> eb5bc7d10d52
  Step 2/5 : RUN apk add git && pip install PyGithub git+https://github.com/SymbiFlow/symbiflow-tools-data-manager#egg=stdm --progress-bar off
   ---> Running in 8367a438dd8b
  fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/main/x86_64/APKINDEX.tar.gz
[...]

PR #2365, opened 2022-01-13
ci: fix check tests docker image build

acomodi:
This PR should fix https://github.com/SymbiFlow/symbiflow-arch-defs/issues/2363


acomodi:
The Docer image is correctly built: https://github.com/SymbiFlow/symbiflow-arch-defs/runs/4802846844?check_suite_focus=true (I have removed the change to the CI, as that is only supposed to run on `master`)

mithro (approved):
LGTM

PR #2375, opened 2022-02-03
CI: initial gh custom runners

acomodi:
This PR is to start transitioning towards custom runners for GH actions to replace kokoro


.github/ci/docs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
umarcor, line 1: These scripts might be located as functions inside `common.sh`.

acomodi: As an initial support for running CI on GH actions I think we may keep them separated. Mainly because there are some of them which require running different targets.

I think having the steps defined in each of the CI scripts makes it easier to track which CI is executing which target

PR #2480, opened 2022-03-08
ci: cleanup kokoro utils

umarcor:


acomodi (commented):
LGTM, but before landing this, there are a couple of things to do IMO:
- Add push mechanism to a GCP bucket, similar to https://github.com/chipsalliance/fpga-interchange-tests/issues/102
- Verify that we are adding all relevant artifacts that we want (and push them as well on GCP) so that debugging gets easier

acomodi (approved):
LGTM

PR #2493, opened 2022-03-11
Fix issues with constraining multiple PLL/MMCM for xc7

mkurc-ant:
This PR relaxes the constraint that enforces a `PLL`/`MMCM` and its driving `IBUF` to be in the same CMT region. This is required in case when a single `IBUF` drives more than one `PLL`/`MMCM`.

The constraint is relaxed in a way that among all `PLL`/`MMCM` driven by an `IBUF` at least one of them is required to be in the same CMT region. Therefore the change should not affect placement quality of existing designs.

### DIFF
--- a/xc/common/utils/prjxray_create_place_constraints.py
+++ b/xc/common/utils/prjxray_create_place_constraints.py
@@ -9,6 +9,7 @@ import vpr_place_constraints
 import lxml.etree as ET
 import constraint
 import prjxray.db
+from collections import defaultdict
 
 CLOCKS = {
     "PLLE2_ADV_VPR":
@@ -116,7 +117,7 @@ CLOCKS = {
     "IBUF_VPR":
         {
             "sources": frozenset(("O", )),
-            "sinks": frozenset(),
+            "sinks": frozenset(("I", )),
             "type": "IBUF",
         },
     "IBUFDS_GTE2_VPR":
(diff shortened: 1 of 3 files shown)


xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -559,6 +559,7 @@ def assign_cmts(self, vpr_grid, blocks, block_locs):
                     )
 
                 if net in self.input_pins:
+                    assert False, "Got a pure input net {}".format(net)
```
acomodi, line 562: Not sure if there was an error while committing the changes, because this might just cause the script to fail every time it enters this `if`

mkurc-ant: Yes, that was a mistake... Now the actual code with the fix should be pushed.

acomodi (approved):
LGTM, let's wait for the CI's response

xc/common/utils/prjxray_create_place_constraints.py
```diff
@@ -883,6 +926,7 @@ def main():
     parser.add_argument('--graph_limit', help='Graph limit parameters')
 
     args = parser.parse_args()
+    eprint(" ".join(sys.argv))
```
acomodi, line 929: This might be a leftover, it should be probably removed

acomodi (commented):
LGTM, one small comment

xc/common/utils/prjxray_create_place_constraints.py
mkurc-ant: Right, removed.

PR #2509, opened 2022-03-18
move toolchain wrappers to repo f4pga

umarcor:


kgugala (changes requested):
current version emits complete toolchain (including wrappers) when a user calls `make install`. The new version will not do that at all - it will only emit architecture definitions. We should clearly say about this change somewhere (as it will be confusing for "old" users)

umarcor:
FTR, there is https://github.com/antmicro/f4pga-arch-defs/commits/umarcor/pip-req (note https://github.com/antmicro/f4pga-arch-defs/commit/c612abd991d3ed571315993a34397c9527702f86) but unfortunately I couldn't get it to work...

kgugala:
> FTR, there is https://github.com/antmicro/f4pga-arch-defs/commits/umarcor/pip-req (note [antmicro@c612abd](https://github.com/antmicro/f4pga-arch-defs/commit/c612abd991d3ed571315993a34397c9527702f86)) but unfortunately I couldn't get it to work...
> 
> >  GitHub**[Commits · antmicro/f4pga-arch-defs](https://github.com/antmicro/f4pga-arch-defs)**FOSS architecture definitions of FPGA hardware useful for doing PnR device generation. - Commits · antmicro/f4pga-arch-defs

sure, but this is CI only. I'm talking about users who would want to manually install stuff on their machines

umarcor:
This PR was updated to install the f4pga package through the `conda-lock.yml` file. Therefore, users should see no difference compared to the previous behaviour.

tmichalak:
LGTM. Merging.

Xiretza:
So what happened to these wrappers? Are they now stored in another repo, or are they gone completely?

tmichalak:
@Xiretza the toolchains have been moved to the [f4pga](https://github.com/chipsalliance/f4pga/tree/main/f4pga) repository, but this is just a transition stage before switching to their newer Python version that are also in the f4pga repo. You can read about the CLI in the [documentation](https://f4pga.readthedocs.io/en/latest/f4pga/Usage.html). The current changes should be transparent to arch-defs users, but we will be heading towards adopting the new CLI in f4pga-examples so it effectlively becomes the default interface.

Xiretza:
Ah, I see, thanks for the information - I'll get acquainted with the new tooling then!

umarcor:
@Xiretza, see #2669.

- We merged this PR on Monday.
- We merged chipsalliance/f4pga#530 on Tuesday.
- Yesterday, I created a bunch of issues:
  - #2669
  - chipsalliance/f4pga-examples#297
  - chipsalliance/f4pga-examples#298
  
So, we are in a transition period. Well, actually, we've spent the last 2-3 month transitioning (from Kokoro to GitHub runners, from SymbiFlow to F4PGA and now from makefiles to Python).

If you want to bump the arch-defs packages in your environment, and keep using the same makefiles/scripts you have, I suggest to read #2669. You just need to define F4PGA_ENV_BIN and F4PGA_ENV_SHARE pointing to the Conda bin and the location where you extracted the packages, respectively. That will "temporarily stabilize" your setup, until you get familiar with the new tooling system (and we improve the docs about how to use `f4pga build` in this repo, in examples, etc.).

PR #2601, opened 2022-04-24
add dsp

liibin:
I have added DSP related information, but failed to generate architecture definitions file and RRG.

What did I do wrong？


liibin:
@[acomodi](https://github.com/acomodi) Can you check it for me？

acomodi:
Hi @liibin, the changes you added are only affecting the build system, so that the DSP can be added, what is missing is the actual addition of the information required to add the DSP primitive.

What is missing is the following:
- DSP primitive block which should be similar as [the other ones](https://github.com/SymbiFlow/f4pga-arch-defs/tree/main/xc/xc7/archs/artix7/tiles)
- The DSP definition in the `cells_map.v` and `cells_sim.v` files. You can check https://f4pga.readthedocs.io/en/latest/flows/f4pga.html#technology-mapping-for-vpr for more information on this.

liibin:
> Hi @liibin, the changes you added are only affecting the build system, so that the DSP can be added, what is missing is the actual addition of the information required to add the DSP primitive.
> 
> What is missing is the following:
> 
> * DSP primitive block which should be similar as [the other ones](https://github.com/SymbiFlow/f4pga-arch-defs/tree/main/xc/xc7/archs/artix7/tiles)
> * The DSP definition in the `cells_map.v` and `cells_sim.v` files. You can check https://f4pga.readthedocs.io/en/latest/flows/f4pga.html#technology-mapping-for-vpr for more information on this.
> 
> > **[In F4PGA — F4PGA  documentation](https://f4pga.readthedocs.io/en/latest/flows/f4pga.html#technology-mapping-for-vpr)**

OK,Thank you,I'll have a try.

xc/common/utils/prjxray_arch_import.py
```diff
@@ -521,6 +521,8 @@ def is_in_roi(conn, roi, tile_pkey):
     "IBUFDS_GTE2": ("Y", 2),
     "IPAD": (None, None),
     "OPAD": (None, None),
+    "DSP48E1": (None, None),
+    "TIEOFF": (None, None),
```
acomodi, line 525: The `TIEOFF` pin is actually not really required as it contains hooks to the power rails (GND and VCC), these should already be included in the RR graph.

acomodi (changes requested):
The only thing thing that is missing now is the physical block type, in the `xc/common/primitives` directory, which should be something similar to the [BRAM](https://github.com/SymbiFlow/f4pga-arch-defs/tree/main/xc/common/primitives/bram_l) one. I see there already exists a directory called `dsp48e1`, but this containing some simulation models as far as I see, probably intended to be used with `v2x`.

What should end up in there is a DSP `model.xml` and `pb_type.xml` files that describe the primitive.

xc/common/utils/prjxray_arch_import.py
liibin: But if I don't add it,  it will report an error

liibin:
> . I see there already exists a directory called `dsp48e1`, but this containing some simulation models as far as I see, probably intended to be used with `v2x`.

I use v2x to generate `model.xml` and `pb_type.xml`, many errors are encountered, and the correct file cannot be produced after modification

liibin:
I wrote `model.xml` and `pb_ type.xml`, and `spartan7` is added. The architecture definitions file and RRG file are successfully obtained, but the following errors occur when executing test. @acomodi can you help me check what's wrong？
image

acomodi:
Hi @liibin. I think that arriving at the routing step is quite a good milestone.

I have looked briefly at the `MULTISIGNIN` input of the DSP and it appears that it is directly connected with the below DSP, and it does not have a possible connection with the `GND` or `VCC` constants. Probably that input is left unused, and it is automatically assigned to `$false`, which causes the failure as no path is possible.

To temporarily work around it you may remove the following port from [here](https://github.com/SymbiFlow/f4pga-arch-defs/pull/2601/files#diff-9fd4f54c8e45c3795f2736c80da52cb129e119ce9b272e8a41a5892af374a292R10924) as well as in the model and pb_type.

liibin:
Generated `top.bit`,but it do not run on FPGA board.

acomodi:
> Generated top.bit,but it do not run on FPGA board.

@liibin this is quite good news! What is left to do is adding the FASM features to the DSP, but before diving into that, we might first clean up and check-in the changes added in this PR, so that we have a gradual increment of the added features.

This PR does two main things, which can be independently split in two smaller PRs in my opinion:
1. Addition of the spartan7 devices
2. Addition of the DSP primitive

One thing that is missing also is the signoff on the commits. You may refer to [this](https://github.com/f4pga/prjxray/blob/master/CONTRIBUTING.md#sign-your-work) link to fix the failing DCO check.

I believe that once those PRs are green and merged, we can than proceed on finishing up the DSP support with the FASM addition, so that we can produce a valid and working bitstream.

kgugala:
@liibin why did you close this PR? It seems it was going in a right direction

liibin:
@acomodi Do you have a reading list for the FASM? I can learn from it.

liibin:
> @acomodi Do you have a reading list for the FASM? I can learn from it.

https://docs.verilogtorouting.org/en/latest/utils/fasm/

liibin:
#2658 #2707

PR #2658, opened 2022-05-16
Support for spartan7

liibin:


acomodi (commented):
Hi @liibin, this looks good. I think that, to verify that everything will be correctly working in CI, you may enable the `counter` test for the spartan7 along side with the others: https://github.com/SymbiFlow/f4pga-arch-defs/tree/main/xc/xc7/tests/counter.

Once the test is added and CI goes green, I think we can mege.

acomodi:
@liibin CI went green but I see that the spartan7 test did not actually run, this because the newly added board was not added to the boards that are being tested in CI. To fix the problem you can add `all_ax7050_bin` here: https://github.com/SymbiFlow/f4pga-arch-defs/blob/main/CMakeLists.txt#L172-L185

acomodi (approved):
LGTM, thanks for all the changes! We can merge as soon as CI goes green

Issue #2673, opened 2022-05-18
Optimise CI workflow and reorganise the content accordingly


umarcor:
The CI workflow in this repo takes ~10h to run. On the one hand, that's because running some of the tasks on the largest architectures (xc7200t) takes 5h. On the other hand, job `install` is executed after all the previous tasks are finished, and it takes 5h as well. This is partially due to some tasks being executed twice (i.e. job `install` is bootstraped instead of reusing artifacts from previous jobs). We should store some of the packages (`arch.xml`, rr graph, etc.) as artifacts (apart from pushing them to GCP), so we can reuse them in job `install`.

Moreover, the structure of the repo is misleading for newcomers. There are multiple locations for tests (`tests`, `testarch`, `xc/xc7/tests`, `quicklogic/pp3/tests`) and subdirs for architectures are located in the root along with subdirs for tools. Having subdirs `archs` and `tools`, along with reorganising the tests, would be helpful to reduce the maintenance burden and make contributions easier. See https://github.com/SymbiFlow/f4pga-arch-defs/pull/2647.

mithro:
It is important from a visibility to have the support architecture directories at the top level.

PR #2674, opened 2022-05-19
ci: print packages' timestamp in job 'install'

umarcor:
Packages are uploaded to GCS through `gsutil cp ${package} gs://${GCP_PATH}/${TIMESTAMP}/`. Unfortunately, the target is hidden in the logs:

https://github.com/SymbiFlow/f4pga-arch-defs/runs/6472094719?check_suite_focus=true#step:4:24660
```
Copying file://symbiflow-arch-defs-xc7a50t-arty-swbut-overlay_test-11aa9fd.tar.xz [Content-Type=application/x-tar]...
```

Therefore, users cannot guess the URL of the arch-def packages generated by some specific commit. They know the format, they know the hash, but they don't know the specific timestamp. This PR prints it.

Ref chipsalliance/f4pga-examples#292

### DIFF
--- a/.github/ci/install.sh
+++ b/.github/ci/install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 INSTALL_DIR="$(pwd)/install"
 mkdir -p $INSTALL_DIR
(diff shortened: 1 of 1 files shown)


mithro:
Would be great to print the full URL too?

mithro:
Maybe a candidate for the new job summary stuff - https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/?

umarcor:
> Would be great to print the full URL too?

Since there are several URLs and all of them have a similar syntax, I don't think it's worth. The main use case is to update the date and hash used in a for loop: https://github.com/chipsalliance/f4pga-examples/pull/300/files. I.e., it's a matter of copying two values. Copying the full URL is not helpful in that case.

umarcor:
> Maybe a candidate for the new job summary stuff - https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/?

That's a very interesting feature! Thanks for the reference!

However, I feel it to be out of scope of this PR. I'd like to address #2675, #2676 and #2673 before generating reports/summaries.

umarcor:
@mithro I did some tests with GITHUB_STEP_SUMMARY in hdl/containers. There seem to be issues when using composite actions, because the summaries are overwritten. See, for instance, https://github.com/hdl/containers/actions/runs/2355254788. However, it looks quite nice if the content is written from a "regular" step. Therefore, I added printing the timestamp and the hash in the summary.

PR #2685, opened 2022-05-23
ci: use f4pga.common.tdm instead of stdm 

umarcor:


umarcor:
This is no longer relevant, since the logic for uploading the packages to GCP was reworked.

PR #2713, opened 2022-06-01
ci: fix updating *-latest pointers

umarcor:


.github/scripts/check-status.py
```diff
@@ -1,59 +1,22 @@
...
-
-    PACKAGE_RE = re.compile("symbiflow-arch-defs-([a-zA-Z0-9_-]+)-([a-z0-9])")
-
-    for artifact in artifacts:
-        name = artifact["name"].split(".")[0]
-        url = artifact["url"]
-
-        m = PACKAGE_RE.match(name)
-        assert m, "Package name not recognized! {}".format(name)
+with (Path(__file__).parent.parent.parent / 'packages.list').open('r') as rptr:
```
acomodi, line 9: Where is the `packages.list` file coming from? I cannot find it checked-in, so maybe it is present only in CI. Maybe it is worth adding a comment on what it is and what it contains.

.github/scripts/check-status.py
```diff
@@ -1,59 +1,22 @@
...
 
-try:
-    check_status()
-except Exception:
-    print('::set-output name=skip::true')
+        with (Path("install") / f"symbiflow-{package_name}-latest").open("w") as wptr:
+            wptr.write(
+                'https://storage.googleapis.com/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/'
+                f'symbiflow-arch-defs/continuous/install/{argv[1]}/{artifact}'
+            )
```
acomodi, line 22: It might be worth starting to transition to the `f4pga` naming of the packages as well, at least for the `latest` packages.

.github/scripts/check-status.py
umarcor: It's created in CI, indeed: https://github.com/SymbiFlow/f4pga-arch-defs/pull/2713/files#diff-f88e7ee60a80885f57e6c7794c2dd3d3604ffed08c7c2bde856b1669186e8444R92. Eventually, I think we can merge steps 'Upload to Google Cloud Storage (GCS) bucket', 'Check Status' and 'Update latest URL files' into a single script. See https://github.com/SymbiFlow/f4pga-arch-defs/pull/2723/files#diff-f88e7ee60a80885f57e6c7794c2dd3d3604ffed08c7c2bde856b1669186e8444.

.github/scripts/check-status.py
umarcor: I believe that preserving the old URL for backwards compatibility is a requirement. However, we can push to multiple URLs at the same time. What's the new f4pga URL/naming?

acomodi: Yeah, I think we might go with `f"f4pga-{package_name}-latest"` and push it alongside with the legacy `symbiflow` one than.

tmichalak:
LGTM

PR #2799, opened 2022-07-13
xc: xc7: fix parameters width for RAMB

acomodi:

This fixes a situation where yosys outputs parameters with more bits than required, causing a FASM writing failure, as the expected width of the parameter is not respected.

E.g. the `DOA_REG` parameter has only one bit and, if an integer is used to define the parameter in verilog, the `DOA_REG` parameter will have a total of 32 bits, generating a failure when writing FASM from VPR.


mkurc-ant (approved):
Ok, LGTM

PR #2821, opened 2022-07-27
Fix QuickLogic install job

lpawelcz:
This PR moves QuickLogic installed toolchain tests to a separate job in order to fix package generation for QL. It also fixes some inconsistencies in locations of utility scripts. This PR is Cross-dependent with: https://github.com/chipsalliance/f4pga/pull/587

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -60,6 +60,15 @@ jobs:
       - Linux
       - X64
 
+    strategy:
+      fail-fast: false
+      matrix:
+        install:
+          - install_xc7
+          - install_ql
+
+    name: ${{ matrix.install }}
+
     env:
       MAX_CORES: 80
       GHA_EXTERNAL_DISK: "tools"
@@ -72,7 +81,7 @@ jobs:
           submodules: recursive
 
       - name: '🚧 Execute test script'
-        run: stdbuf -i0 -o0 -e0 ./.github/scripts/install.sh
+        run: stdbuf -i0 -o0 -e0 ./.github/scripts/${{ matrix.install }}.sh
 
       - name: '📤 Upload artifact: arch-defs packages'
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/Tests.yml
```diff
@@ -60,6 +60,15 @@ jobs:
...
       - X64
 
+    strategy:
+      fail-fast: false
+      matrix:
+        install:
+          - xc7
+          - ql
+
+    name: ${{ matrix.install }}
```
umarcor, line 70: Please use `name: install-${{ matrix.install }}`. There is another job named xc7 in the Test matrix, and it might be confusing.

lpawelcz: I edited the whole job name inside `install` matrix. Is that ok?

lpawelcz:
Also cross-dependent with: https://github.com/chipsalliance/f4pga-examples/pull/338

umarcor (commented):
LGTM!

It feels that `install_ql.sh` and `install_xc7.sh` might be combined into a single script that accepts an argument to select the family. Nonetheless, that's out of scope of this PR. We can handle it afterwards.

On the other hand, I don't see where is `ql.sh` used at the moment. Again, not relevant to this PR, but we should have a look at it later.

I won't mark this as approved now, because we need to wait until CI is green and then remove the temporary commit to use the f4pga branch. Other than that, very nice work. Congrats!

lpawelcz:
@umarcor `ql.sh` is used in this test: https://github.com/SymbiFlow/f4pga-arch-defs/blob/a11cc174344c83bfb2cc8db77b5ba3abf9c17f82/.github/workflows/Tests.yml#L28
It runs a series of test designs from internal cmake infrastructure while `install_ql` installs the toolchain and uses it directly.

lpawelcz:
CI went green with temporary reference to custom `f4pga` branch.
CI run: https://github.com/SymbiFlow/f4pga-arch-defs/actions/runs/2753793137
temporary commit: https://github.com/SymbiFlow/f4pga-arch-defs/pull/2821/commits/a11cc174344c83bfb2cc8db77b5ba3abf9c17f82

@umarcor I deleted this commit so we can merge this PR.

tmichalak (approved):
LGTM

PR #2827, opened 2022-07-31
create subdir 'lattice'

umarcor:


mithro:
I guess we have kind of lost this already, but the idea was that from the top level it should be pretty easy to see which FPGA architectures were supported.

If we are going this direction, should `xc` really be `xilinx` then?

umarcor:
IMHO, the README and the main page of the documentation do make explicit the families supported. Overall, we are seeking consistency. QuickLogic stuff is grouped, and while we are working on reorganizing and cleaning up the content of this repo, it is desirable to have common scripts for each vendor grouped. In the future, with more homogeneous flows, we might prefer to have e.g. `lattice_ice40`, `lattice_ecp5`, `ql_eos-s3`, `xilinx_7`...

> If we are going this direction, should `xc` really be `xilinx` then?

I think so. We should be consistent with `quicklogic`/`ql`/`eos-s3` and `xilinx`/`xc`/`xc7`. I started with Lattice, because it is the less disruptive change.

mkurc-ant (approved):
LGTM

PR #2829, opened 2022-08-01
ci/scripts: combine install_xc7 and install_ql into install

umarcor:


lpawelcz (approved):
Tested locally QL eos-s3 side. Generated packages with `./.github/scripts/install.sh eos-s3` and used them in `f4pga` (rev: https://github.com/chipsalliance/f4pga/commit/c342fc6ff1684f3dc6072713730ac9fc574ab2f3) repository for running QuickLogic example design in both `f4pga build` flow and deprecated toolchain wrappers. Everything works, LGTM!

PR #3826, opened 2024-08-13
add CFGLUT5

liibin:


liibin:
I would like to add a description of CFGLUT5. Is this correct for me? Can someone help me check it

